PR #18, opened 2018-06-10
[Dropdown] Only check option value to determine if option is selected

ColinFrick:
Fixes Semantic-Org/Semantic-UI#5817


lubber-de:
Also fixes https://github.com/Semantic-Org/Semantic-UI/issues/5481

lubber-de:
Also fixes https://github.com/Semantic-Org/Semantic-UI/issues/7111

PR #19, opened 2018-06-11
Fix last tab inside container which does not have correct bottom margin

prudho:
Fixes Semantic-Org/Semantic-UI#5408.


hammy2899:
I think this needs some looking into because there must of been a reason for it being `0em`

PR #27, opened 2018-06-13
[Form Validation] Avoid trim of values before validation

prudho:
Fixes Semantic-Org/Semantic-UI#6237.


hammy2899:
This needs some discussion before we move forward.

It was brought up that users may want to trim at the start, end or never and this also needs support for backwards compatibility.

PR #31, opened 2018-06-19
Begin CI implementation

prudho:
Implements CI building trough TravisCI.

For now it just try to build FUI with the latest NodeJS version. Next steps will be tests and deploy 😃


hammy2899:
Great start for https://github.com/hammy2899/Fomantic-UI/issues/29

preinstall.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+mv semantic.json.example semantic.json
```
hammy2899: This could be `cp` instead of moving.

prudho: It could be yes, actually it doesn't really matter, since this preinstall will be launched only by Travis in a VM. But if you want, I can modify `mv` to `cp`.

hammy2899: I think the Travis installation should be identical to how it is installed for users therefore it doesn't lead to any problems.

prudho: No problem, will do then 😃

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+
+# Caching node_modules folder for speed up build
+cache: npm
+
+before_install:
+  - sh preinstall.sh
+
+env:
+  global:
+    - NODE_ENV=production
```
hammy2899: Could you add a newline to the EOF?

preinstall.sh
```diff
@@ -0,0 +1,3 @@
...
+sed -i '/"autoInstall": false/s/false/true/' semantic.json
```
hammy2899: Could you add a newline to the EOF?

.travis.yml
prudho: Done in bfa25a9.

preinstall.sh
prudho: Done in bfa25a9.

preinstall.sh
prudho: Done in 76027c6.

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+
```
hammy2899: Can you remove the unnecessary spaces here

prudho: Mmmm these spaces are pretty standard in the travis configuration file. (https://docs.travis-ci.com/user/environment-variables#global-variables)

hammy2899: At the EOF?

hammy2899: image

prudho: Oh, sorry I did not saw those 😕

prudho: Done in  9ac43f3.

PR #41, opened 2018-06-27
[Search] Add setting for error header in search results (Fix #6442)

prudho:
Fix #6442. Thanks **@ColinFrick** !


hammy2899:
As discussed in discord you need to change the function to have another argument instead of getting the text from the global settings.

prudho:
Yeah, we clearly need to find another approach to fix this... shall we close this PR or keep using this one ?

hammy2899:
Leave this open we can just modify what you currently have

PR #45, opened 2018-06-29
 [Tab] Allow centered tab menu 

prudho:
This PR allow tab menu to be centered. Also works with the secondary style. Closes Semantic-Org/Semantic-UI#5589.

Thanks @franckpopchef for the implementation idea !


romaninsh:
shouldn't it be "centered menu" or "center aligned menu", not "center menu" ?

no sample or doc.

ColinFrick:
@romaninsh it follows the same syntax as the alignment modifiers for the menu (left / right)

PR #57, opened 2018-07-06
Implements dropdown different sizes (standalone and in forms)

prudho:
This PR allow different sizes for dropdowns. It will need a lot of tests before merging it !

### Closed Issues
Semantic-Org/Semantic-UI#5154

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -1050,4 +1050,56 @@
   font-size: @massive;
 }
 
+/* Dropdown */
+.ui.mini.form .field .dropdown {
+  font-size: @mini;
+}
+.ui.tiny.form .field .dropdown {
+  font-size: @tiny;
+}
+.ui.small.form .field .dropdown {
+  font-size: @small;
+}
+.ui.form .field .dropdown {
+  font-size: @medium;
+}
+.ui.large.form .field .dropdown {
+  font-size: @large;
+}
+.ui.big.form .field .dropdown {
+  font-size: @big;
+}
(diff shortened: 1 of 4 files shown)


prudho:
I have added CSS for menus dropdowns and buttons groups dropdowns.

Here is a gist I use to tests many dropdown cases: https://gist.github.com/prudho/5a4af7e96a5abf6d08df6b3a0fb7aec4

There's some little tweaks to provide yet:
- [ ] images aren't exactly vertical aligned to text.
- [ ] labels aren't exactly vertical aligned to text.

But aside this little issues, this code should now me mergeable 😎

brendon:
With this change, if one has a button driven dropdown (where the button is also the dropdown container), and one sizes the button as `tiny` for example, the dropdown menu contents will also be font-sized `tiny` which in my case was unexpected. Previously, the button would be sized tiny, but the menu would be normal sized.

Ideally there should be a way to override the size of the dropdown if the driving container's size has been set to something else.

a.k.a. just because we want the button to be tiny, doesn't mean we want the dropdown menu to also be tiny.

prudho:
Could you provide a JSFiddle please ? Or open a new issue with an example ? Thanks

brendon:
Sorry, here are the fiddles. Let me know if you'd still rather me open a new issue:

Fomantic-UI: https://jsfiddle.net/4ytcfw8b/
Semantic-UI: https://jsfiddle.net/gpakxocL/4/

Further investigation shows me that more complex menu's break down under SUI when using `tiny` on the button but I didn't have a very complex menu :)

PR #64, opened 2018-07-09
Lint fixes

prudho:
First batch of lint fixing commits.

Just the more obvious errors here: unused variables, double declarations, typos.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -46,10 +46,7 @@ $.fn.form = function(parameters) {
         $field,
         $group,
         $message,
-        $prompt,
         $submit,
-        $clear,
-        $reset,
 
         settings,
         validation,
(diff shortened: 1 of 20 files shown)


hammy2899:
Going to put this on hold until we redo build tools/modules (3.0)

src/definitions/behaviors/form.js
```diff
@@ -782,7 +781,7 @@ $.fn.form = function(parameters) {
           // alias
           rules: function(field, rules) {
             if($.isArray(field)) {
-              $.each(fields, function(index, field) {
```
lubber-de: This is definately a bug, not a linting improvement

prudho: Yeah the linter trigerred an `unused variable` but it's a bug. I'll submit a PR soon.

PR #78, opened 2018-07-28
Range slider component

hammy2899:
Added the range slider component made by @tyleryasaka.

Range repo issue: https://github.com/tyleryasaka/semantic-ui-range/issues/30

### Closed Issues
https://github.com/Semantic-Org/Semantic-UI/issues/1336


src/definitions/modules/range.js
```diff
@@ -0,0 +1,273 @@
+/*!
+ * # Range slider for Semantic UI.
```
prudho: Could you set the standard header ? Here it is:
```js
/*!
 * # Semantic UI - Range slider
 * http://github.com/semantic-org/semantic-ui/
 *
 *
 * Released under the MIT license
 * http://opensource.org/licenses/MIT
 *
 */
```

src/definitions/modules/range.js
```diff
@@ -0,0 +1,273 @@
...
+            
+            var closestHiddenParent = $(elem).parentsUntil(':visible');
+            
+            if (closestHiddenParent.length != 0) {
+              observer.observe(closestHiddenParent[closestHiddenParent.length - 1], {attributes: true});
+            }
+          },
+          
+        };
+
```
prudho: Not a blocking stuff, but we should add later common modules methods (debug, performance, display, verbose...).

hammy2899: Good shout, didn't think of that

PR #81, opened 2018-08-06
[Label] fixed incorrect label background color

hammy2899:
Fixed some ribbon labels having an incorrect background color and added primary & secondary colors.

image

### Closed Issues
https://github.com/Semantic-Org/Semantic-UI/issues/6497


lubber-de:
This also fixes
https://github.com/Semantic-Org/Semantic-UI/pull/6498

PR #82, opened 2018-08-06
[Search] added ability to disable link action

hammy2899:
Added ability to disable the link and turn it into a div when `actionURL` is equal to `false`.

actionurl-div

### Closed Issues
https://github.com/Semantic-Org/Semantic-UI/issues/6490


lubber-de:
This also fixes https://github.com/Semantic-Org/Semantic-UI/issues/5977

PR #91, opened 2018-08-15
[Form] Added a transparent textarea option

LewHgh:
This pull request is to add a transparent textarea option on forms as shown below:
image

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -159,6 +159,19 @@
   vertical-align: @checkboxVerticalAlign;
 }
 
+/*--------------------
+      Transparent
+---------------------*/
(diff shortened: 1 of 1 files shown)


ColinFrick:
Great addition!
Quick question: Do we have to handle the field error state for this?

Currently it would look like this:
grafik

hammy2899:
Transparent inputs currently work like that but maybe we should change them to looks something like this:

textarea:
image
image

input:
image
image

ColinFrick:
@hammy2899 We could merge this and open a new issue for the error state for both input and textarea.

hammy2899:
@ColinFrick Yep I agree

PR #97, opened 2018-08-20
[Dependency] Fix compatibility issue with less >= 3.5 due to math changes

ColinFrick:
Replaces all tilde string escaping with explicit function calls.
This was tested with gulp-less@3.5.0 and gulp-less@4.0.1

Reported in Semantic-Org/Semantic-UI#6512

### DIFF
--- a/src/definitions/collections/grid.less
+++ b/src/definitions/collections/grid.less
@@ -139,7 +139,7 @@
   margin: (@rowSpacing / 2) (@gutterWidth / 2);
 }
 .ui.grid .column + .ui.vertical.divider {
-  height: ~"calc(50% - "(@rowSpacing/2)~")";
+  height: e(%("calc(50%% - %d)", @rowSpacing / 2));
 }
 
 /* Remove Border on Last Horizontal Segment */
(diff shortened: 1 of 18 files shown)


src/definitions/collections/grid.less
```diff
@@ -1112,12 +1112,12 @@
 .ui.relaxed[class*="vertically divided"].grid > .row:before {
   margin-left: (@relaxedGutterWidth / 2);
   margin-right: (@relaxedGutterWidth / 2);
-  width: ~"calc(100% - "@relaxedGutterWidth~")";
+  width: e(%("calc(100%% - %d)", @relaxedGutterWidth));
```
prudho: `100%%` ? This might not be correct.

ColinFrick: It is:
> If you need to print the percentage symbol, escape it by another percentage %%.

% format
http://lesscss.org/functions/#string-functions

prudho: Oh yeah I see, sorry !

lubber-de:
This als fixes 
https://github.com/Semantic-Org/Semantic-UI/issues/6650

lubber-de:
This is save for LESS 3.8 aswell and also fixes
https://github.com/Semantic-Org/Semantic-UI/issues/6678

PR #119, opened 2018-09-14
Lower modals are not accessible anymore in multiple modals

prudho:
#### Description
This PR will prevent the ability to interact with modals that are not the focused one. It works by adding an inside dimmer to each modal that got the `allowMultiple: true` setting. If a modal is open, other will be dimmed so you cannot use them, until you close the last modal by pressing ESC (thanks to #118), or you close all of them by clicking on the dimmer.

#### Preview
[](http://tof.cx/image/zwi8V)

#### Closed Issues

Semantic-Org/Semantic-UI#5770

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -88,6 +88,11 @@ $.fn.modal = function(parameters) {
 
           module.create.id();
           module.create.dimmer();
+
+          if ( settings.allowMultiple ) {
+            module.create.innerDimmer();
+          }
+
           module.refreshModals();
 
           module.bind.events();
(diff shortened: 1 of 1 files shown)


lubber-de:
If 2.5.1 will include the latest SUI fix (release 2.3.4) for [modals](https://github.com/Semantic-Org/Semantic-UI/pull/6562): Will this PR still work then?

prudho:
Yes, the preview gif actually includes the SUI fix, so it will work. But I'm not sure that the modal fix will be included in 2.5.1 yet (massive changes needs heavy testing).

PR #122, opened 2018-09-18
[Toast] Add a toast module

prudho:
This PR adds a basic toast module, allowing to notify user of an event. It misses some features yet, but i think it's a good start already !

#### Try it out
[Demo here](http://prudho.fr/semantic/examples/toast.html)

#### Related issues
Semantic-Org/Semantic-UI#1430
Semantic-Org/Semantic-UI#2793


src/definitions/modules/toast.js
```diff
@@ -0,0 +1,515 @@
...
+
+              $toast
+                .addClass(className.icon)
+                .append($icon)
+              ;
+            }
+
+            if (settings.title !== '') {
+              var 
+                $titre = $('<div/>')
```
ColinFrick: titre? 😄

Ikaer:
Do you think, you will add some design for actions buttons ? Kind of like this:
image
(we use an old version of jquery.noty https://ned.im/noty/v2/ )
Or maybe plan some attached system to do this (like segment)

src/definitions/modules/toast.js
prudho: Yeah I was tired 😃

prudho:
Well, the toast module is pretty young, and it clearly misses some cool features yet.

This one could be a nice addition, feel free to open an issue, or eventually a PR 😉

BTW: this is an undocumented feature, because not fully tested yet, but you can add html code by the `message` parameter !

romaninsh:
Great addition. Here are some suggestions based on my previous experiences with them:


**USABILITY**
 - [ ] make sure you can copy-paste the text by selecting it.
 - [ ] do not hide it if the mouse hovered over it. Some users keep mouse together with their eyes
 - [ ] use newest message should be in the corner, pushing out previous message (DEFAULT). This generally is much more easier to read.

**USE CASES**
 - [ ] ability to add a button or a link is important as well as incorporate longer content.
 - [ ] icon should be optional
 - [ ] ability to use custom colours
 - [ ] test with modal window overlap

**IDEAS**
 - [ ] ideally if selecting a center position, should be proportional to the screen dimensions
 - [ ] no padding on mobile device.

nasatome:
In my opinion, it is best to do something similar to https://sciactive.github.io/pnotify/

In my particular case, I love it, I do not know if you could integrate some of that type directly to Fomantic, and add a theme, to avoid reinventing the wheel.

But I repeat: I have no idea about integrations of this type.

PR #139, opened 2018-09-28
[Toast] Progressbar, closeEasing, Bugfixing, message support

lubber-de:
### Added
- Option` showProgress` (default false), which displays a SUI attached progressbar increasing until displayTime is reached (ignored, if displayTime = 0) Values: `'bottom'`,`'top'`,`false`
- Option `progressUp` (default true). If set to false the progress is decreasing instead of increasing
- Support for 'message' component in addition to 'toast' to work with progressbar if `className: {toast: 'ui message'}` is used (all message color options are supported to display a progressbar with the same color)
- Option transition.closeEasing (default `'easeOutBounce'`) to animate the stacking when a toast closes. Set empty for old behaviour
- Optional `onBeforeHide` (will be called given a callback parameter) functionality to _transition.js_ to make the above closeEasing animation possible
- Support `className: {toast:'ui floating toast'}` just like in message to display a box-shadow underneath/around the toast.  (`className: {toast:'ui hoverfloating toast'}` will show the box-shadow only on hovering)
- Option `closeIcon` (default false). This will make the toast closable by the top right corner icon instead of clicking anywhere on the toast
- Support string to option `showIcon` to be able to use individual icons

### Changed
- Removed fixed width, added an option `compact` (default true) which does the same but does this independ of the className
- Removed fixed Opacity, added an option `opacity` (default 1) instead as discussed below

### Fixed
- Timeout bug: If toast was closed by click, the timeout was still running and returned error when reached because $toast was not existing anymore
- Icon positioning on large wrapping text in toast to stay at the top left of the text
- Icon will only be rendered, if class exists in icon object
- Alignment of toast with different widths in right and center position

#### Progressbar `showProgress: 'bottom'`
toast_progressbar
```javascript
$('body')
          .toast({
            displayTime: 5000,
            message: 'I am a toast, nice to meet you! this is a lot of text and should be displayed properly',
            class: 'info',
            showIcon: false,
            showProgress: 'bottom'
        });

```
#### Progressbar decreasing `progressUp: false`
toast_progressup_false
```javascript
$('body')
          .toast({
            displayTime: 4000,
//           title: 'Welcome!',  //set title:'header' below if you use 'ui message' instead of 'ui toast'
            message: 'I am a toast, nice to meet you! this is a lot of text and should be displayed properly',
            class: 'brown',
            className: {
                toast: 'ui compact message',
//                title: 'header'   //remember to set this aswell when using a title in a message
            },
            showProgress: 'bottom',
            progressUp: false
        });

```

#### Animated stacking on close `transition: closeEasing = 'easeOutBounce'`
toast_closeeasing
```javascript
$('body')
          .toast({
            displayTime: 2000,
            message: 'I am a toast, nice to meet you! this is a lot of text and should be displayed properly',
            class: 'purple',
            compact: false, //do not restrict to a fixed width
            className: {
                toast: 'ui message',
            },
            transition: {
                  closeEasing:'easeOutBounce', //this is default, you can omit it
//               closeEasing:'', //set to empty string to gain the previous behaviour without any animation
            },
            showProgress: 'bottom'
        });

```

#### Top Progressbar ```showProgress: 'top'```
toast_top_progress

```javascript
$('body')
          .toast({
            title: 'Welcome to a fresh Toast',
            displayTime: 4000,
            message: 'I am a toast, nice to meet you! this is a lot of text and should be displayed properly',
            class: 'error',
            className: {
                toast: 'ui floating toast',
            },
            showProgress: 'top'
        });
```

#### Close Icon ```closeIcon: true```

Difference to usual toast: It's only closable by close-icon instead of clicking anywhere in the toast. cursor changed accordingly
toast_close_pointer

#### Individual Icon  ```showIcon: 'your icon name'```

Using a string instead of true for showIcon will use that string for the icon display. Using `true` remains the old behaviour (displaying the appropriate icon which matches the class)

```
showIcon: 'accessible'
```
```
showIcon: 'microphone'
```

individual_icon


lubber-de:
> Thanks for the pull request. Could you change the base branch to `beta`?

Done ;)

hammy2899:
@lubber-de Great PR 👍, I will take a look and do some testing when I get some time.

src/definitions/modules/toast.less
```diff
@@ -119,7 +158,34 @@
     font-size: @toastIconFontSize;
     opacity: @toastIconOpacity;
 }
+
+.icon.flat.toast > .icon:not(.close) {
```
prudho: Where does this `flat` class come from ?

lubber-de: I invented it. This will override the normal 0.9 opacity toast to full opacity toast component.
Personally i like that better, but i did not want to completely break the original authors idea which only sets opactity to 1 when you hover over the toast.
Anyway, the message class will always have opacity 1, so there was no need to invent a flat class aswell there.

lubber-de: `className: {toast:'ui flat toast'}`
image

`className: {toast:'ui toast'}` (default)
image

To be honest: Even with the increased 0.9 opacity, this looks unuseful to me. I suggest having the toast always have opacity 1 as default like message and we can remove the flat class there

lubber-de: This is opacity: 0.97. If we really want to keep lower opacity, it should be raised to >=0.95 again imho. The idea of a toast component is to be displayed over existing content. In most cases, i guess 0.9 is still to low to read the toast-content nicely.
image

hammy2899: I do agree with this.

prudho: In this case why don't we let the user to set an opacity as a toast parameter ?
```
opacity: 0.8 //set as 1 by default
```

src/definitions/modules/toast.less
```diff
@@ -119,7 +158,34 @@
...
     vertical-align: @toastContentVerticalAlign;
-}
\ No newline at end of file
+}
+
+/*---------------
+   Progress Bar
+ ----------------*/
+
+.toast-box .attached.progress .bar {
```
prudho: This toast box should host all toasts, not just the toasts that have progress bar.

lubber-de: toast-box is only used when a progress bar is enabled. otherwise the additional wrapper toast-box is  not needed as before and the additional div is not rendered

lubber-de: This (simplified) is still used when a progressbar should not be enabled
```html
<div class="toast-container">
	<div class="ui toast></div>
</div>
```


This (simplified) is used when a progressbar is enabled (to wrap the progressbar correctly underneath the toast component
```html
<div class="toast-container">
	<div class="ui toast-box">
		<div class="ui toast></div>
		<div class="bottom attached progress">
			<div class="bar"></div>
		</div>
	</div>
</div>
```

prudho: Agreed but adding a parent to only some kind of toast might break some behaviors in the long term. For now the toast component is young and pretty simple, but with time its complexity will increase and we must think about it now

hammy2899: I agree with @prudho. Maybe we should do something like this
```html
<div class="toast-container">
  <div class="ui toast">
    <div class="content">
      <!-- title, message etc. -->
    </div>
    <div class="bottom attached progress">
      <div class="bar"></div>
    </div>
  </div>
</div>
```

src/definitions/modules/toast.less
lubber-de: ok, i'll remove the 'flat' class and will add an option for opacity (default 1)
Please keep in mind, whenever somebody is using something lower than 1 the fadeIn options of SUI will always animate to Opacity 1 and then immediatly the  toast would change to the lower opacity (otherwise the whole css-fade animation would need a change aswell, which i think is overkill) Here is ax example how it looked with the original fix 0.8 opacity
toast_org_fadein

prudho: Yeah I saw that too. But letting the user the ability to change the opactity is nice imho.

dist/components/toast.css
```diff
@@ -60,19 +61,45 @@
   padding: 0.78571429em 1em;
   margin: 0 0 6px;
   color: rgba(255, 255, 255, 0.9);
-  opacity: 0.8;
+  opacity: 0.9;
+}
+
+.floating.toast {
+  box-shadow: 0 0 0 1px #a3c293 inset, 0 2px 4px 0 rgba(34,36,38,.12), 0 2px 10px 0 rgba(34,36,38,.15);
```
prudho: I really like this ! Since we got rid of opacity, what do you think about display it (or a variant) on mouse over ?

lubber-de: I suggest implementing this as an additional class, so the user can decide how it should behave

Will only show the box-shadow on hover
```css
className: {
	toast: 'ui hoverfloating toast'
}
```

Will still show the box shadow any time
```css
className: {
	toast: 'ui floating toast'
}
```

src/definitions/modules/toast.less
lubber-de: Basically good idea. But how about supporting the message component easily then without the need to somehow recreate everything from message into toast aswell?
I was very glad, the complete message component functionality was working with toast just by changing the className.
At first, i was thinking of getting rid of the "own" toast class in favour of simply use the message component completely. But, again, i did not want to break everything @prudho already did :smile: 
For now i would just change to code to render "toast-box" wrapping all the time instead of just render it when a progressbar should be shown. What do you think?

dist/components/toast.css
lubber-de: Changed / added accordingly by commit f49a63cf3100149835360a7ecaef4e6d4c07b9ee.

src/definitions/modules/toast.less
lubber-de: By commit ccf7ad077c0cd1fd5e178acec49ac7b04b940eaf the parent element "text-box" is now always rendered. 

Btw: Also supported "top attached progress" and fixed icon display with this commit :grin:

src/themes/default/modules/toast.variables
```diff
@@ -24,10 +25,16 @@
 @toastSuccessColor: @successColor;
 
 /* Icon */
-@toastIconRightMargin: 0.6em;
+@toastIconContentPadding: 3em;
```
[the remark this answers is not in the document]
lubber-de: Alright, I am going to fix the mentioned issues.. 🙂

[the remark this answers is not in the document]
ColinFrick: Cool, keep up the good work :)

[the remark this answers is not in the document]
lubber-de: Wow, that was quite a headache today and took some time to solve, but i hope i covered everything now and fixed a lot:

- Every combination of toast or message with top or bottom progress is now fixed. Every different width (according to content) does not disturb any other toast or it's progres anymore
- added `compact` option (default true) This will display the toast (regardless if toast or message) in the fixed 300px width (if "compact2 is part of className it will also be considered the same way. Settting this to false will still nicely display every toast responsively, so i thought to make this option false by default (you guys can decide)
- fixed alignment for center and right positions if different toasts are shown at the same time
- removed unused opacity variables, which were left behind from the last commit
- I left out the dist css/js this time :smile:

lubber-de:
Meanwhile, while you are testing my latest commit :wink: :
What do you think about having the option "showProgress" support either `false` or `'top'` or `'bottom'` (instead of only true/false) as an addition to the current only possibility to use a top progressbar by defining `className: {progress:'ui top attached active progress'}` ?
 I think `showProgress: 'top'` is easier to use (and even less needed code if we maybe set showProgress:'bottom' to default.) If needed one can still use an own className definition.

hammy2899:
@lubber-de I totally agree with this. I think it should be `showProgress` but default to false but allow `false`, `true`, `'top'` and `'bottom'`

ColinFrick:
@lubber-de would make sense. Definitely one thing I didn't really like of the original commit.

etshy:
Can I ask what the `true` value will do ? 
Will it be the same as `'bottom'` (as "default value" to avoid break changes maybe) ?
Or something else ?

If it does the same as `'bottom'`, I don't see the point to have both options (except the breaking changes thing).

lubber-de:
@etshy 
Yes, it's actually not a difference right now. What i could add is to explicitely set `bottom` and remove `top` when `showProgress: 'bottom'` has been defined. Then the difference  to `true` would be that `true` simply uses the given `settings.className.progress` value without any change. Of course, by default this is also 'ui bottom attached active progress', but could be changed to something different by the user at runtime and toast will leave that untouched then. :thinking:

etshy:
Options should `false` (to not show the progress, ofc), `'top'` and `'bottom'` I guess.

@lubber-de I didn't take a look to the code or event the toast features yet so I don't really know what is `settings.className.progress`. A string to set classes to the progress bar ?
If yes, in the case you use `showProgress: true`, and don't specify a top or bottom class in `settings.className.progress` what happen ?

IMO, top or bottom should always be here so the `'top'` or `'bottom'` value is mandatory, either in `settings.className.progress` or in `showProgress`.

Let me know if I misunderstood something here.

hammy2899:
@lubber-de @etshy I can see where you both are coming from but I think we should only have `false`, `'top'` and `'bottom'` it may mean a little more code but at the end of the day, it should be straightforward and easy to understand as a user.

default:
`showProgress: false`

available:
`showProgress: false|'top'|'bottom'`

lubber-de:
Alright, but if somebody sets this to true we should raise an exception on the console then?

lubber-de:
I think, I simply add the string value of showProgress to the div and if it's any boolean value, the progress won't be rendered completely. This way the user would also be able to add more classes to the progress if needed

etshy:
@hammy2899 that's how I thought it would be better, just `false`, `'top'` and `'bottom'`.

I was just asking, IF there is a `true` value, how it will be handled.

I guees you could raise an exception if true or any others value is set, and not show anything ?
I don't know if it's possible to do that with the current code though.

Thanks a lot to work so much on this 👍

ColinFrick:
@lubber-de is this ready for review now, or do you have more changes planned?

lubber-de:
@ColinFrick It's ready now. I have more additions in progress, but this can be a separate PR then. I won't touch this branch anymore unless you'll need more fixes/changes

hammy2899:
If someone has some time and could do the docs this could be in our next release.

PR #143, opened 2018-10-01
[Dropdown] Multiple search keyboard + icon fix

lubber-de:
## Description

A multiple dropdown selection searchable box had

1. The close icon displayed over the search input field when many entries have been selected
2. Whenever an entry was selected via enter, it looses it's focus making quick keyboard selection impossible (you needed to click again into the search field

Before:
multidd_before

After:
multidd_after

### DIFF
--- a/dist/components/dropdown.js
+++ b/dist/components/dropdown.js
@@ -1451,6 +1451,7 @@ $.fn.dropdown = function(parameters) {
                     module.event.item.click.call($selectedItem, event);
                     if(module.is.searchSelection()) {
                       module.remove.searchTerm();
+                      $search.focus();
                     }
                   }
                   event.preventDefault();
--- a/dist/components/dropdown.css
+++ b/dist/components/dropdown.css
@@ -358,7 +358,7 @@
   min-height: 2.71428571em;
   background: #FFFFFF;
   display: inline-block;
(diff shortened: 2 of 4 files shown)


dist/components/dropdown.js
```diff
@@ -1451,6 +1451,7 @@ $.fn.dropdown = function(parameters) {
                     module.event.item.click.call($selectedItem, event);
                     if(module.is.searchSelection()) {
                       module.remove.searchTerm();
+                      $search.focus();
```
hammy2899: Why is this line added but the src not provided?

lubber-de: I am confused.. Which source is missing?

hammy2899: This is the dist file

ColinFrick:
Thanks for the pull request. Just a quick note, you shouldn't include the dist files in the commits, that would make merging easier :)

dist/components/dropdown.js
lubber-de: :fearful: My bad...now it's there.  :smile:

lubber-de:
> Thanks for the pull request. Just a quick note, you shouldn't include the dist files in the commits, that would make merging easier :)

Ok, sorry, i'll keep that in mind from now on

PR #148, opened 2018-10-03
force display:none on .placeholder.hidden.transition

lubber-de:
## Description
Using a hidden transition on a placeholder does not set display:none, because placeholder style will still override it. 
The Fix should make sure it wont happen to any other element having .hidden.transition

## Testcase
https://jsfiddle.net/zqfrjso1/4/ (from original issue)

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6608

### DIFF
--- a/src/definitions/modules/transition.less
+++ b/src/definitions/modules/transition.less
@@ -48,9 +48,10 @@
 }
 
 /* Hidden */
+.hidden.hidden.transition,
 .hidden.transition {
-  display: none !important;
-  visibility: hidden !important;
+  display: none;
+  visibility: hidden;
 }
 
 /* Visible */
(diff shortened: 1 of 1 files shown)


ColinFrick:
Thanks for the pull request.
IMHO the usage of !important in css should be discouraged, I don't know how the other guys want to handle this. Increasing the specificity of the hidden transition selector would fix this as well:
```css
.hidden.hidden.transition {
  display: none;
  visibility: hidden;
}
```

hammy2899:
I agree with @ColinFrick, we need to stop using `!important`

lubber-de:
Very interesting  approach by doubling the className :+1:

On the other and: !imporant is used 2230 times (!) in the final compiled semantic.css... :thinking: 
So, should i change the PR to your solution or keep mine?

lubber-de:
ok, just read @hammy2899 comment...will change it then

Issue #158, opened 2018-10-07
[Form/Input/Checkbox/Dropdown] Inconsistent inverted logic


lubber-de:
## Problem
The naming logic of inverted form elements is not consistent.
This could be considered a feature instead of a Bug but currently it's an inconsistent behaviour, which somehow breaks the "inverted" logic for Form Elements:

### Input
has two inverted variants:

`ui inverted input`
![image](https://user-images.githubusercontent.com/18379884/46583796-4f3d0080-ca5c-11e8-865c-f6e45aaddc7e.png)

`ui inverted transparent input`
![image](https://user-images.githubusercontent.com/18379884/46583805-654ac100-ca5c-11e8-9690-520d7df701b2.png)

### Checkbox
has only one variant:
`ui inverted checkbox`
![image](https://user-images.githubusercontent.com/18379884/46583874-1cdfd300-ca5d-11e8-915c-8553ed32f225.png)

The matching "inverted transparent" variant according to input logic is completely missing for checkbox.

### Dropdown 
also has only one variant, But this time "inverted" is completely black without adding "transparent". This breaks the logic acording to input and checkbox
![image](https://user-images.githubusercontent.com/18379884/46583990-61b83980-ca5e-11e8-8026-0675bc7e314a.png)

### Textarea
 does not have any inverted variant at all :disappointed: 

This does not feel consistent and makes clearly inverted forms using the default framework nearly impossible.

## Solution

I suggest implementing the following class additions :

`ui inverted transparent checkout`
Suggested Example:
![image](https://user-images.githubusercontent.com/18379884/46583825-9a571380-ca5c-11e8-8104-816fb16d4f02.png)

`ui inverted transparent dropdown` (which behaves as it currently is without "transparent")
`ui inverted dropdown` (that will be a breaking change)
`ui inverted textarea`
`ui inverted transparent textarea`

## Poll
What do you guys think?

prudho:
Yes, I agree that the `inverted` variation is pretty inconsistent between different components. But for now Fomantic's purpose is to be fully compatible with Semantic, on already existing components, so we're tied with these inconsistencies for now. Maybe for an eventual 3.0 ! 😃

PR #161, opened 2018-10-08
[Form] Support inverted transparent textarea

lubber-de:
## Description

Support for transparent textarea in an inverted segment or form (only the color needed to be changed)

### Related to
#158

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -163,8 +163,6 @@
       Transparent
 ---------------------*/
 
-.ui.inverted.form textarea.transparent,
-.ui.form .inverted.segment textarea.transparent,
 .ui.form .field textarea.transparent {
   border-color: transparent !important;
   background-color: transparent !important;
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -163,13 +163,15 @@
       Transparent
 ---------------------*/
 
-
+.ui.inverted.form textarea.transparent,
+.ui.form .inverted.segment textarea.transparent,
 .ui.form .field textarea.transparent {
   border-color: transparent !important;
   background-color: transparent !important;
   padding: 0em !important;
   box-shadow: none !important;
   border-radius: 0px !important;
+  color: inherit;
 }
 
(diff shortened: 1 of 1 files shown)


src/definitions/collections/form.less
```diff
@@ -1117,4 +1117,13 @@
   font-size: @massive;
 }
 
+
+.ui.inverted.form textarea.transparent,
+.ui.form .inverted.segment textarea.transparent {
+  color: inherit;
+}
+
```
hammy2899: Why so much whitespace?

lubber-de: I simply adjusted the logic the same as it is for input.

... or maybe i don't get, what you mean? :thinking:

lubber-de: if you simply mean the additional empty lines...then i am simply sorry for that, there was no reason for sure

ColinFrick: What he probably meant is that you added two empty lines before the selector and four lines after.
Additionally couldn't this be added to this:
https://github.com/fomantic/Fomantic-UI/blob/31ed0007b7133967b7ef2ad2d52cc8328acf4299/src/definitions/collections/form.less#L162-L173
or to the inverted variation section?
https://github.com/fomantic/Fomantic-UI/blob/31ed0007b7133967b7ef2ad2d52cc8328acf4299/src/definitions/collections/form.less#L714-L749

hammy2899: You have 4 blank lines of whitespace added and there isn't any need for it

hammy2899:
I agree with @ColinFrick you can probably achieve this without making another textarea definition.

lubber-de:
@ColinFrick Thanks, i moved it to a more suitable area, taking care of no extra whitespace now, within the less file as suggested. I'll take care of such things in the future! :blush:

src/definitions/collections/form.less
lubber-de: ok, got it now working in the suggested transparent section also. see latest commit. Sorry for any inconvenience :wink:

PR #167, opened 2018-10-12
[Toast] fix position for top and bottom centered container

lubber-de:
## Description
On scrolling body content the `top center` or `bottom center` positioning was broken


lubber-de:
@hammy2899 I am confused... github says here, you reverted the commit, but in beta it's still commited (which is fine :smile:) and i don't see a reason for a revert. I also cannot find the revert-commit in the git log anywhere

PR #169, opened 2018-10-12
fix(cards): fix horizontal card actions being forced to the right.

### LINKED ISSUE #159: [Card] horizontal cards meta & actions are forced to the right
# Bug Report

## Steps to reproduce
1. create a horizontal card
2. add some meta data or a button

## Expected result
The meta data or button should appear under the card

## Actual result
The meta data or button is forced to the right

## Description

This is caused because of `flex-direction: row` when setting `flex-wrap: wrap` buttons then flow under the card but meta data will force the content to go under the image. If the content is too long it will also wrap to the next row.

## Testcase
https://jsfiddle.net/Lwxau2yn/

## Screenshot
image

## Version
2.6.2

tcmal:
fixes #159
## Testcase

Previous: https://jsfiddle.net/Lwxau2yn/
New: https://jsfiddle.net/tcmal/dys0khbx/

## Screenshot (when possible)

image

## Closes

#159

### DIFF
--- a/src/definitions/views/card.less
+++ b/src/definitions/views/card.less
@@ -427,6 +427,7 @@
 .ui.horizontal.cards > .card,
 .ui.card.horizontal {
   flex-direction: row;
+  flex-wrap: wrap;
   min-width: @horizontalMinWidth;
   width: @horizontalWidth;
   max-width: 100%;
@@ -453,6 +454,12 @@
   width: 100%;
   height: 100%;
   border-radius: @defaultBorderRadius 0 0 @defaultBorderRadius;
+}
+
+.ui.horizontal.cards > .card > .bottom.attached,
+.ui.card.horizontal > .bottom.attached,
+.ui.card.horizontal > .top.attached {
+  flex-basis: 100%;
 }
 
 /*-------------------


hammy2899:
Using `flex-basis` is a good idea but it doesn't work for this https://jsfiddle.net/tcmal/dys0khbx/

tcmal:
It displays fine for me, MDN says it should be well implemented.

hammy2899:
This doesn't seem to work for me on Chrome or Firefox

image

https://jsfiddle.net/r3q2wco6/

Does anyone else have the problem?

tcmal:
Works for me on firefox & chrome. What versions?

Does this work better? https://jsfiddle.net/tcmal/5sbjqkh6/

I had a typo so the flex-basis wasn't actually being applied. I also added it to top attached buttons for consistency.

hammy2899:
No, that someone makes it worse, the image and content should be on the same row but meta data and buttons etc should be on another under the image and content.

tcmal:
Does it work better [with a grid](https://jsfiddle.net/tcmal/rhfznovb/)? It seems like the width isn't being set high enough for you, presumably it's a display size thing.

PR #178, opened 2018-10-16
[Toast] Change default class for title to header

lubber-de:
## Description
`header` as classname is used everywhere else (message, card, popup, ...), so i made this "compatible" now.
I left the option being named `title`. Of course for backward compatibility but especially because `popup` uses exact the same logic (parameter is named `title`, but rendered className is `header`.

This also reduces the need to change the classname manually to `header` in case somebody wants `ui message` as the toast-style

I changed the docs accordingly [here](https://github.com/fomantic/Fomantic-UI-Docs/pull/37)


hammy2899:
This will have to wait since its a breaking change.

lubber-de:
No, it isn't IMHO , because it's just used internally by the component. That's why I also did not change the parameter title itself, but just the default classname used

ColinFrick:
@lubber-de Theoretically it is, anyone who styled the `.title` differently with their custom less/css needs to change that to `.header`

lubber-de:
In that case, and very much I doubt, anybody already did this just because of the reason toast is new since the last version, it's still possible to use `className:{title:'title'} ` again.
Because the new toast enhancements coming with 2.6.3 will increase the chance people are using the component, I highly encourage you to think about again, if this PR should be merged into 2.6.3 before releasing it. It really makes the naming logic more consistent and we should clean that before it really gets a problem for many people :smile:

ColinFrick:
I'm not arguing that we shouldn't merge this change now (because Toast is still BETA), but that it's still a BREAKING CHANGE, because the user has to change their code to match this change.

lubber-de:
The next PR I prepared for toast is to support `ui card` as well, in preparation of using images and actions of that component as well. I am trying to limit the needed options for the user to be submitted to toast and card also uses `header`.
Anyway, we can mark that change in the docs with a proper label and as you said, it's still beta

OK, but in the end you guys decide :thinking :, i can continue doing the new stuff for toast without the need to have this merged first. Don't get me wrong, I don't want to force anything, just trying to discuss my Pov of benefits.

hammy2899:
After reading the comments I can see both sides to this, merging now would mean the next release we do would include the change (anything in the `beta` branch gets added to the next release) but I can see your point about the component being in beta so users should expect the change.

I think we should stick to how people expect changes to be made even if a component is in beta its still in a released version of the library therefore we should treat it like any other component and shouldn't make breaking changes without proper versioning.

PR #186, opened 2018-10-18
[Dimmer] Added support for using dimmers only partially on top,center,bottom

lubber-de:
## Description

This adds the possibility to use any dimmer only partially and limited only to the overall height of its content and not over the whole to be covered area. This can be used for example as a caption on any element of choice by simply using
 `top dimmer`, `center dimmer` or `bottom dimmer`

Works also very well with transitions 'fade up' or 'fade down'

#### Static without Javascript (just add `active` to the dimmer as usual)

```html
<div class="ui medium image">
  <div class="ui active top dimmer">
    <div class="content">
      <h3>Look!</h3>
      ui top dimmer
    </div>
  </div>
  <img src="assets/images/avatar/nan.jpg">
</div>
```
#### Animation triggered via Javascript

```html
<div class="ui medium image">
  <div class="ui center dimmer">
    <div class="content">
      <h3>Look!</h3>
      ui center dimmer
    </div>
  </div>
  <img src="assets/images/avatar/nan.jpg">
</div>
```
```javascript
$('.center.dimmer').dimmer({
    transition: 'fade down',
    on: 'hover'
});
```
## Screenshot
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/3459
https://github.com/Semantic-Org/Semantic-UI/pull/6621

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -332,10 +332,10 @@ body.dimmable > .dimmer {
     transform: translateY(-50%);
   }
   40% {
-    transform: translateY(calc(50% - 30px));
+    transform: translateY(calc(-50% - 30px));
   }
   60% {
-    transform: translateY(calc(50% - 15px));
+    transform: translateY(calc(-50% - 15px));
   }
 }
 .loadUIOverrides();
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -270,6 +270,7 @@ body.dimmable > .dimmer {
 .ui[class*="center caption"].dimmer{
   top:50% !important;
   transform: translateY(-50%);
+  -webkit-transform: translateY(calc(-50% - .5px));
 }
 
 .ui[class*="center caption"].dimmer.transition[class*="fade up"].in {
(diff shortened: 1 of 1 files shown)


lubber-de:
@prudho Thanks..., that's the old nasty webkit bug. Fixed by latest commit.

src/definitions/modules/dimmer.less
```diff
@@ -256,4 +256,96 @@ body.dimmable > .dimmer {
...
+    transform: translateY(-50%);
+    -webkit-transform: translateY(calc(-50% - .5px));
+  }
+  40% {
+    transform: translateY(calc(-50% - 30px));
+  }
+  60% {
+    transform: translateY(calc(-50% - 15px));
+  }
+}
```
prudho: Please add a blank line here and it's all good for me 😃

lubber-de: > Please add a blank line here and it's all good for me

Where?  (and curiously: why?) (i remember somebody was telling me once not to insert too much whitespace :wink:

prudho: Where ?
```less
@keyframes bounceCenter {
  0%, 20%, 50%, 80%, 100% {
    transform: translateY(-50%);
    -webkit-transform: translateY(calc(-50% - .5px));
  }
  40% {
    transform: translateY(calc(-50% - 30px));
  }
  60% {
    transform: translateY(calc(-50% - 15px));
  }
}
/* !!! NEW LINE HERE !!! */
.loadUIOverrides();
```

Why ? Well, just to be consistent with other less files actually.

ColinFrick:
I'm conflicted with naming this `.caption` as the content and container of the dimmer can be arbitrary: https://jsfiddle.net/Korinu/L053zgmu/
Why not use something like `.partly`, `.partially` or something similar?

src/definitions/modules/dimmer.less
lubber-de: ok then

lubber-de:
> Why not use something like `.partly`, `.partially` or something similar?

I am totally open to this. :slightly_smiling_face: @prudho @hammy2899 Your opinion?

prudho:
`caption` doesn't bother me, French translations feels pretty close to the purpose (`légende` or `sous-titre` eg. `subtitle`).

hammy2899:
IMHO I think `partially` makes the most sense when using a dimmer because to me `caption` seems like it would be its own component.

prudho:
@hammy2899 is the most english of us 😄, so i'll follow his advice.

lubber-de:
Then i am going to change it to "partially center" instead of "center partially"  @hammy2899  Agreed?

lubber-de:
We could also completely omit partially but simply use `top dimmer`, `center dimmer` `button dimmer` by using explicit contains selectors this would not interfere with something like `ui bottom aligned dimmer`

lubber-de:
Removed the need for any additional className. Much cleaner. I don't know why i haven't done it this way right from the start :thinking:

PR #196, opened 2018-10-23
[Progress] Display dark text on light inverted backgrounds

lubber-de:
## Description
Some progress text is unreadable on inverted progress bars, because their background color is too light.
This PR changes the text colors from white to black on `inverted yellow|olive|teal|grey progress`

### Before
image

### After
image

### DIFF
--- a/src/definitions/modules/progress.less
+++ b/src/definitions/modules/progress.less
@@ -462,6 +462,12 @@
   background-color: @lightBlack;
 }
 
+.ui.yellow.inverted.progress .bar>.progress,
+.ui.olive.inverted.progress .bar>.progress,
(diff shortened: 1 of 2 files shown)


ColinFrick:
We could also just change the `@invertedProgressColor` to something like `@black`:
image

This maybe would look more consistent.

hammy2899:
I agree with @ColinFrick so that it is consistent.

lubber-de:
OK, convinced :wink:. I am gonna changing this Pr accordingly and for toast I am creating a separate PR

prudho:
Was also on @ColinFrick's opinion.

PR #204, opened 2018-10-25
Make the pointing tips in inverted menu reflecting menu background co…

exoego:
### Description

Currently, pointing tips of `.active.item` in `.inverted.pointing.menu` are all black.
Those looks incosistent to tips in non-inverted pointing menu. 
before

This PR makes the pointing tips can reflect a color of parent (`.active.item`), when those are in `.inverted.pointing.menu`.
after

### Testcase
https://jsfiddle.net/z0vzw26e/3/

### Closed Issues
https://github.com/Semantic-Org/Semantic-UI/issues/3486 (already closed accidentally)

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1863,7 +1863,43 @@ Floated Menu / Item
   background-color: @arrowVerticalSubMenuColor;
 }
 
-
+/* Inverted Colors */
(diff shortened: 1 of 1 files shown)


lubber-de:
Thanks for the Pull request. Can you please change the base to beta?

PR #208, opened 2018-10-26
[Modal] Basic Modal and inverted modal on inverted dimmer had wrong close icon color

lubber-de:
## Description
- Basic Fullscreen Modal had black close icon on black dimmer
- Basic Modal on inverted dimmer had grey borders on action and header
- Basic Modal on usual black dimmer had the borders aswell (although nearly invisible)
- inverted modal on inverted dimmer in small viewports had black close icon on black modal

#### Black close icon when using `basic fullscreen modal`
Before:
image
After:
image

#### Border lines when using `basic modal` in `inverted dimmer`
Before:
image

After:
image

#### Black (thus invisible) close icon on small screen when using `inverted modal` in `inverted dimmer` 
Before:
image

After:
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5831

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -171,6 +171,9 @@
 .ui.modal .actions > .button {
   margin-left: @buttonDistance;
 }
+.ui.basic.modal > .actions {
+  border-top:none;
+}
 
 /*-------------------
        Responsive
@@ -301,6 +304,7 @@
 .ui.basic.modal > .close {
   top: @basicModalCloseTop;
   right: @basicModalCloseRight;
+  color: @basicInnerCloseColor;
 }
 
 .ui.inverted.dimmer > .basic.modal {
@@ -324,12 +328,6 @@
   top: auto !important;
 }
 
-/* Tablet and Mobile */
-@media only screen and (max-width : @largestTabletScreen) {
-  .ui.basic.modal > .close {
-    color: @basicInnerCloseColor;
-  }
-}
 
 
 /*******************************
@@ -432,8 +430,9 @@
   right: @innerCloseRight;
   color: @innerCloseColor;
 }
-
-
+.ui.basic.fullscreen.modal > .close {
+  color: @basicInnerCloseColor;
+}
 /*--------------
       Size
 ---------------*/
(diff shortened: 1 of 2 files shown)


hammy2899:
@lubber-de Can you add a screenshot of before and after?

lubber-de:
@hammy2899 Added them to the description

src/definitions/modules/modal.less
```diff
@@ -324,12 +328,6 @@
   top: auto !important;
 }
 
-/* Tablet and Mobile */
```
prudho: Did you test this on mobile to see if everything is good ?

src/definitions/modules/modal.less
```diff
@@ -607,13 +606,13 @@
   border-top: @invertedActionBorder;
 }
 
-.ui.fullscreen.inverted.modal > .close {
-  color: @invertedInnerCloseColor;
```
ColinFrick: This removes all usages of `@invertedInnerCloseColor`. Is this correct?

src/definitions/modules/modal.less
lubber-de: I tested everything responsively. This setting is obsolete, that's why I have deleted it. I moved this setting for the close color in basic modals to line 107 instead. The color is the same regardless of the viewport dimensions, so this separate setting didn't make any sense anymore.

prudho: That's what I wanted to know 😉

src/definitions/modules/modal.less
ColinFrick: I didn't unterstand why the css rule was removed, but I just saw that there is a rule overriding this anyway:
https://github.com/fomantic/Fomantic-UI/blob/2c3bc578defd5fe92e6e47fbdb795636a935e105/src/definitions/modules/modal.less#L620-L622

Yes, please remove it from modal.variables
@hammy2899 @prudho Because the variable was essentially not used before, this isn't a breaking change, right?

prudho: For me, now, since removing it will not have any effect anywhere.

PR #209, opened 2018-10-26
[Modal] image content images can go out of content and description does not wrap

lubber-de:
## Description
Using images within `image content` as suggested in the docs, which are too large, they overlay the modal and the description does not wrap

Original Fix provided by @GammaGames

## Testcase
https://jsfiddle.net/j7qsb24n/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6636


lubber-de:
I had to partly remove this. See #599 for workarounds

PR #213, opened 2018-10-29
[Icon] Fix the "th large" icon with 

exoego:
## Description
The `th large` icon size should reflect the size passed as another class (e.g. 'mini').

## Testcase
https://jsfiddle.net/agrwpkvh/

### Before
th-large-before
### After
th-large-ater

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6499

## Note
It is hard to discriminate the `th icon` with `large` from `th large icon` (marked as red). Thi s issue is reported in https://github.com/Semantic-Org/Semantic-UI/issues/6336.
Perhaps it is better to rename `th large` to `th-large` as similar as `sign-in` icon, `th quadrant` or something like that ? 🤔 
If it is renamed, we don't need a workaround in this PR.

### DIFF
--- a/src/themes/default/elements/icon.overrides
+++ b/src/themes/default/elements/icon.overrides
@@ -777,6 +777,15 @@ i.icon.text.height:before { content: "\f034"; }
 i.icon.text.width:before { content: "\f035"; }
 i.icon.th:before { content: "\f00a"; }
 i.icon.th.large:before { content: "\f009"; }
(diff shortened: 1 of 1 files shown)


hammy2899:
These changes will need to be made in the icon script (https://github.com/fomantic/icon-script)

exoego:
Opened a PR in icon-script.
https://github.com/fomantic/icon-script/pull/2

Issue #29, opened 2018-06-18
Testing


y0hami:
Checklist for testing

Key:
⛔️ = Blocked
🌀 = Upcoming
⚡️ = Started
✔️ = Complete

- [ ] Create pages for all components containing all variations 🌀 
- [ ] Setup [Nightwatch](http://nightwatchjs.org/) with custom CSS style checks (as discussed in discord) 🌀 
- [ ] Setup mocha to test JS (will probably need to change a lot of JS to make this work therefore may happen at later stage)
- [x] Request Open Source access to [BrowserStack](https://www.browserstack.com/) (Accepted as OS project and now have full access to BrowserStack) ✔️ 
- [x] Set up linter ([eslint](https://eslint.org/)) ✔️
- [ ] Setup [Travis-CI](https://travis-ci.org/) to run the tests 🌀 
- [ ] Setup [Travis-CI](https://travis-ci.org/) to build and deploy ⚡️ 
- [ ] Setup [coveralls](https://coveralls.io/) to show coverage (only when JS is done)

romaninsh:
Perhaps https://github.com/BBC-News/wraith or something similar to run the renders?

hammy2899:
Thanks, I will look into these.

lubber-de:
I also came across https://codecov.io which seems to integrate seamlessly into gihub (for auto creates PR comments for code coverage)
Free for open source

mvorisek:
Eslint can be added and tested in CI easily, example from atk4/ui repo setup:

https://github.com/atk4/ui/blob/a1a8a2812a1332732d87667a2ff2a7cd780b81ef/js/.eslintrc.js
https://github.com/atk4/ui/blob/a1a8a2812a1332732d87667a2ff2a7cd780b81ef/js/package.json#L11
https://github.com/atk4/ui/blob/a1a8a2812a1332732d87667a2ff2a7cd780b81ef/.github/workflows/test-unit.yml#L308

PR #216, opened 2018-10-29
[Multi] Fixes some CSS lint issues

prudho:
## Description
This PR fixes several small CSS issues raised by my local linter ([stylelint](https://stylelint.io)). Most of them are duplicate properties or CSS selectors, and I just fixed the most obvious.

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -383,13 +383,11 @@
 }
 .scrolling.dimmable > .dimmer {
   justify-content: flex-start;
+  position: fixed;
 }
 .scrolling.dimmable.dimmed > .dimmer {
   overflow: auto;
   -webkit-overflow-scrolling: touch;
-}
-.scrolling.dimmable > .dimmer {
-  position: fixed;
 }
 .modals.dimmer .ui.scrolling.modal {
   margin: @scrollingMargin auto;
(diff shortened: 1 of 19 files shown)


lubber-de:
You may aswell consider https://github.com/Semantic-Org/Semantic-UI/pull/6401 to get rid of all 0em, 0px when refactoring the less files. I dont know, if you can configure this in styleint or if we should create an additional issue for this :thinking:

prudho:
For now, this is just a first shot, fired while I was making some linting implementation for #29. This is just the top of the iceberg on lint issues, and there's many fixes and tweaks to provide yet, before passing the stylelint check. We should not create lining issues while linting is not set up and running.

src/definitions/modules/modal.less
```diff
@@ -479,9 +477,9 @@
   }
 }
 
-/* mini */
-.ui.small.modal > .header:not(.ui) {
-  font-size: @miniHeaderSize;
+/* Tiny */
+.ui.tiny.modal > .header:not(.ui) {
+  font-size: @tinyHeaderSize;
```
hammy2899: What does changing this to `tiny` fix?

prudho: `mini` was defined two times, and `tiny` was missing from the file. This was a typo.

hammy2899: Ok cool.

PR #230, opened 2018-10-31
[Label] Inverted Variant, basic grouping, basic tag, bugfixes

### LINKED ISSUE #116: [Label] add inverted variation
Add an inverted variation of Label

Same as colored variation but use light colors instead https://github.com/fomantic/Fomantic-UI/blob/master/src/themes/default/globals/site.variables#L239 like inverted buttons https://fomantic-ui.com/elements/button.html#inverted

lubber-de:
## Description
- added inverted variant (normal|basic|tag|corner|ribbon) 
- added support for grouping basic labels
- added support for `basic tag label`
- bug fixes
- moved color generation into less macro for shorter code and easier maintainability
- all color related `!important` statements (> 450 in previous compiled css :scream: ) are not necessary anymore and were removed :slightly_smiling_face: 

## Screenshots
inverted_label_v2
inverted_basic_label_v2
basic_tag_label

## Closes
#116 
https://github.com/Semantic-Org/Semantic-UI/issues/6639
https://github.com/Semantic-Org/Semantic-UI/issues/3413


hammy2899:
I'm not 100% on this, I don't think they should have borders.

@fomantic/admins what's your opinions?

hammy2899:
I should have been more clear when I said

> like inverted buttons

I ment the colors used not the style and hover behavior.

lubber-de:
OK, this could of course reduce complexity, but before I am touching this again, let us clearly decide how every element should behave. It's a bit of inconsistency again compared to other components.

Suggestion :
- remove the hover from `inverted label`, instead always display them with full light background (what the hover currently does) 
- let `basic inverted` behave like the current `inverted` above instead (or remove the hovering there aswell, so it's always bordered)
- keep `basic inverted pointing ' as it is (bordered without any hovering) 
- add `inverted pointing ` which displays them with full background
- same for tags :only `inverted tag' has full light background and `basic inverted tag' is bordered but the hovering removed

hammy2899:
I agree with apart from the fact that inverted labels will look a bit weird if they don't have any hover behavior. Let's wait to see what the others think.

hammy2899:
@lubber-de how about if you have the `ui inverted label` a solid color and on hover it changes to the normal color instead of the inverted version.

lubber-de:
I will use '@lightRed / @lightRedHover' instead of '@lightRed / @red' for hovering then .. Good hint.
Want me to wait until @prudho and @ColinFrick agree to the changes?

hammy2899:
@lubber-de You can do if you want.

ColinFrick:
I'm on the fence here. We have to different approaches for inverted style: button (sui) and toast (fui)
As far as I understand the inverted style shouldn't change the style drastically, but it still does in the button module.
 
The question is, should we change the inverted button implementation accordingly to be more consistent with this implementation?

Additionally we also have inconsistency between the inverted input and inverted dropdown fields.
image
image

lubber-de:
> The question is, should we change the inverted button implementation accordingly to be more consistent with this implementation?
> 
I vote for this. The current `basic inverted button` should also become the current `inverted button`.
It would be fully consistent with the above mentioned label changes then.

If we do this, then this inverted label PR should not be part of 2.6.4 but rather a 2.7.x, don't you think?

> Additionally we also have inconsistency between the inverted input and inverted dropdown fields.
Yes, we still have the discussion at #158 :wink:

hammy2899:
IMHO I would argue that buttons don't have an inconsistency but more that `basic` doesn't exactly make sense. Maybe we want to look at renaming `basic` to something like `outlined button` or `bordered button`. I do agree with the forms and we will need to do something about this in the future.

My main problem right now with the labels is that the current solution in this PR makes them look too much like a button. A label should show information not mislead a user into clicking it.

ColinFrick:
@hammy2899 We already have the problem with label / button:
image
image

Which are the buttons? :thinking:

hammy2899:
@ColinFrick Yeah I know but this is what I am trying to avoid further

lubber-de:
So, how should i proceed with this PR? Changing as suggest above  and we'll see how to deal with the buttons afterwards? (This would be my suggestion)

lubber-de:
Btw: For the `inverted toast` i adopted the `progress bar` logic where the background color is still solid using the light background.
A transparent toast background would make the inside Text nearly unreadable (remember, we had the discussion about the opacity-setting for the same reason) and a always black background with just borders color can be easily accomplished by using `ui inverted message` there

That said i again suggest to adjust this PR according to the above mentioned changes, and let`s discuss/prepare the button afterwards. As @ColinFrick already pointed out: Button is the only component with a different inverted meaning

hammy2899:
@lubber-de I agree, let's change this to use the inverted colors and have a discussion about buttons in discord maybe.

lubber-de:
2nd attempt pushed. :slightly_smiling_face:  Updated screenshots in first comment accordingly.
I now used the light hover variables. Every label has a hovering now, but is only barely recognizable (especially for basic (tag|pointing) labels) because of the tight small change to the original SUI lightcolor variables.
Check it out and tell me if we have a winner now or again some changes are needed :wink:

hammy2899:
I think they look better now but I do agree they have very little hover effect. We may want to see what uses those light hover colors and see if making them more drastic would affect anything.

lubber-de:
The `@light{color}Hover`variables are currently only used in `inverted header`, so i think we can increase the darkening on those variables without any much hassle for a slightly better hover recognition

hammy2899:
@lubber-de I agree.

@fomantic/admins If you disagree feel free to say.

lubber-de:
@hammy2899 
> I think that looks good, only thing I can think of changing is to make the text color white on the solid labels.

:thinking: Well, that would break the logic we now implemented in `inverted toast` and `inverted progress`, where we decided to keep every text color black (because white was more unreadable on light background colors). Let us not get inconsistent again and keep track with those decisions for toast and progress, thus keep the text-color black here aswell. Only exception in comparison to toast and progress: There is no hovering, so we could discuss about having white as text color only on hovering the inverted labels, but this will also break with the non-inverted label-hover logic: hovering over normal colored labels does not change the text-color. (`@{color}TextColor` is the same as `@{color}HoverTextColor`)

hammy2899:
@lubber-de That's a good point, lets just leave it 😄

faheel:
@lubber-de @hammy2899 The related issues haven't been closed yet.

lubber-de:
FUI issues will be closed once we release a new version where the issue is fixed. The SUI issues stay open unless they get fixed there aswell, but this is up to jack who maintains SUI on his own.

PR #232, opened 2018-11-02
[Item] Inverted Variant

### LINKED ISSUE #109: [Item] add inverted variation
Add an inverted variation of item.

lubber-de:
## Description
Inverted variant of Item.

## Screenshot
image

## Closes
#109

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -148,6 +148,7 @@ ol.ui.list ol li,
 .ui.list .list > .item > .content,
 .ui.list > .item > .content {
   line-height: @contentLineHeight;
+  color: @contentColor;
 }
 .ui.list .list > .item > .image + .content,
 .ui.list .list > .item > .icon + .content,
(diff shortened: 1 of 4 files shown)


lubber-de:
So I am removing inverteditemsbackground (which is currently set to black) only then. Because for items, extra and content there already was a background setting in the non inverted variant aswell, but set to transparent or none(as already discussed in the related issue)

lubber-de:
@hammy2899 i removed the black background, so it's up to the user now to set it accordingly. 
I currently kept the inverted background settings for  item, content and extra (missed `inverted` anyway in the commits before) as they are set to the same value (transparent or none) as the normal item

If i should remove those aswell, then just tell me :slightly_smiling_face:

hammy2899:
Maybe we should look into why the background is set by the item before we move forward. If it turns out it isn't for any reason maybe we should just not set it.

lubber-de:
The latest commit supports texts directly in the `.content` element (normal or inverted) when used in `.items` or in `.list` aswell. Fixes https://github.com/Semantic-Org/Semantic-UI/issues/6652

Regarding the background: Also in `.list` it seems every background setting was just prepared for someone to be able to change the background via less variables. That said also for `.list` the background- (and inverted-background) defaults are again always either transparent or none,

While reading through the code of all other elements/modules i could not see any other reason, why item _needs_ to set the background (it was not even `!important`). So for backward compatibility we should not remove it.

So i think we should follow that for item alone too.

PR #267, opened 2018-11-28
fix(form validation): broken `remove rules` behavior

prudho:
### Description
As discovered by @lubber-de, the lint PR trigerred an `unused variable` warning in the `remove rules` behavior from the form validation module. After investigation, turns out that the code was pretty bad and useless, since it allowed removing the same set of rules for multiple fields, which is a nonsense. It's now simplified.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -781,14 +781,7 @@ $.fn.form = function(parameters) {
           },
           // alias
           rules: function(field, rules) {
-            if($.isArray(field)) {
-              $.each(fields, function(index, field) {
-                module.remove.rule(field, rules);
-              });
-            }
-            else {
-              module.remove.rule(field, rules);
-            }
+            module.remove.rule(field, rules);
           },
           fields: function(fields) {
             module.remove.field(fields);


lubber-de:
Why is it nonsense to be able to remove the same set of rules for multiple fields? Probably a rare case and also probably never used by anybody (such people would have encountered a JS error because `fields` is undefined there and the feature `remove rules` is not even documentd in the docs).

If would simply use the same code as for `remove.field`

```javascript
  rules: function(field,rules) {
	var
	  fields = $.isArray(field)
		? field
		: [field]
	;
	$.each(fields, function(index, field) {
	  module.remove.rule(field,rules);
	});
  },
```

or, to even shorten/simplify the code we could make remove.rules a real alias to remove.field by adding the rules parameter to the it.

```javascript
  field: function(field, rules) {
	var
	  fields = $.isArray(field)
		? field
		: [field]
	;
	$.each(fields, function(index, field) {
	  module.remove.rule(field, rules);
	});
  },
    module.remove.field(field, rules);
  },
```

prudho:
Forget this PR... The whole stuff needs to be rewritten as there's many bugs and strange behaviors.

PR #272, opened 2018-11-28
[Dropdown] Clearable dropdown initialized with a select always showed remove icon

### LINKED ISSUE #271: [Dropdown] Clearable dropdown initialized with a select always show the remove icon
# Bug Report

## Expected result
A none selected dropdown without any remove icon.

## Actual result
A none selected dropdown with a remove icon, which is not correct.

## Testcase
https://jsfiddle.net/zw349kst/

## Screenshot
[](http://tof.cx/image/kk14G)

## Version
2.6.4

## Primary analysis
The css selector for hiding the remove icon is not valid when the dropdown is initialized from a `select`. Start investigating here: [/src/definitions/modules/dropdown.less#L650](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/modules/dropdown.less#L650).

lubber-de:
## Description
A clearable Dropdown initiated via an existing `select` tag always showed the clearable icon.
Thats because it's unfortunately not possible to either check an empty value via css on a `select` tag and there also is no clear/easy indication for CSS from the generated DOM if any value was selected.
So i added a class `noselection`  which is set to the hidden input/select html element on runtime if nothing is/was selected to be able to easily fetch empty values via css and finally fix the issue :wink: 

In addition i also fixed the cursor for the clearable icon which, on multiple search dropdowns, was set to text, and the indication to click on it was missing.

## Testcase
Same as @prudho but injects the fixed dropdown.js from the github branch :slightly_smiling_face: 
https://jsfiddle.net/zw349kst/1/

## Closes
#271

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2147,9 +2147,9 @@ $.fn.dropdown = function(parameters) {
               module.set.selected();
             }
             if(module.get.item()) {
-              $input.removeClass('noselection');
+              $input.removeClass(className.noselection);
             } else {
-              $input.addClass('noselection');
+              $input.addClass(className.noselection);
             }
             module.remove.initialLoad();
           },
@@ -2520,9 +2520,9 @@ $.fn.dropdown = function(parameters) {
           },
           value: function(value, text, $selected) {
             if(value !== undefined && value !== '' && !(Array.isArray(value) && value.length === 0)) {
-              $input.removeClass('noselection');
+              $input.removeClass(className.noselection);
             } else {
-              $input.addClass('noselection');
+              $input.addClass(className.noselection);
             }
             var
               escapedValue = module.escape.value(value),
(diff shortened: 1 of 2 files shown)


prudho:
This is the kind of fix I imagined too, great ! Do you think we could apply this `noselection` class to all dropdowns and only rely to this for display or no the remove icon ? Code would be simpler.

lubber-de:
@prudho The current addition in dropdown.js actually applies that extra class `noselection` to every dropdown already. I was unsure if it's wise and future proof to only rely on that class for every kind of dropdown, so i kept the css solution for input
If we only rely on `noselection`then basically it might be simplified to

```css
.ui.dropdown .noselection ~ .remove.icon,
.ui.dropdown.loading .remove.icon {
```

@hammy2899  @ColinFrick What do you think?

lubber-de:
> Also, could you add the `noselection` class to the `className` setting please ?

Done ;)

hammy2899:
@lubber-de For now I would say just leave it as is since it works that way we aren't changing things which may cause problems.

Issue #274, opened 2018-11-30
List of errors in compiled CSS


Atulin:
## Why?
![results](https://user-images.githubusercontent.com/11233299/49276506-04010600-f47f-11e8-80f9-abb1983c0a05.png)

## Disclaimer

The list is not comprehensive. It mostly shows types of errors occuring in the compiled CSS, rather than listing all 250 of them. I believe I grouped them by probable source well enough, but feel free to take a look at the code through various linters yourselves.

Checking was done with WebStorm 2018.3

## Errors

Only unique types are being listed. Well, most of them. The second error, for example, is prevalent throughout the entire file, far too many occurrences to count.

- [ ] ``::-webkit-selection``

* **Line:** 673
* **Error:** Unknown pseudo-selector
* **Probable cause:** Linter isn't up to date

- [ ] ``will-change: '';``

* **Line:** 806
* **Error:** Mismatched property value
* **Probable cause:** Incorrect variable or compilation error

- [ ] ``-webkit-box-shadow: 0px 0px 0px 1px transparent inset, none;``

* **Line:** 855
* **Error:** Mismatched property value (at ``none``)
* **Probable cause:** Incorrect variable or compilation error

- [ ] ``display: -webkit-inline-box;``

* **Line:** 1198
* **Error:** Mismatched property value
* **Probable cause:** Linter isn't up to date or obsolete value

- [ ] ``-webkit-box-shadow: 0px 0px 0px 1px rgba(34, 36, 38, 0.35) inset, 0px 0px 0px 0px rgba(34, 36, 38, 0.15) inset inset;``

* **Line:** 1739
* **Error:** Mismatched property value (at second ``inset``)
* **Probable cause:** Incorrect variable or compilation error

- [ ] ``i.icon.\35 00px,``

* **Line:** 14395
* **Error:** Unexpected token
* **Probable cause:** Incorrect variable

- [ ] ``-webkit-box-shadow: 1px solid rgba(34, 36, 38, 0.15);``

* **Line:** 18755
* **Error:** Mismatched property value
* **Probable cause:** Incorrect variable?

- [ ] ``border-width: none;``

* **Line:** 18838
* **Error:** Mismatched property value
* **Probable cause:** Incorrect variable

- [ ] ``-webkit-transition: 0.25s color ease 0.1s, 0.25s padding-left ease 0.1s, 0.25s background-color ease 0.1s;``

* **Line:** 19599
* **Error:** Mismatched property value
[...]

PR #275, opened 2018-11-30
[CSS] Fix some errors

prudho:
Following #274.

Most of theses error are wrong css properties or values


src/definitions/elements/label.less
```diff
@@ -1181,7 +1180,7 @@ a.ui.basic.label:hover {
 .ui.pointing.label:before {
   background-color: inherit;
   background-image: inherit;
-  border-width: none;
+  border-width: 0px;
```
hammy2899: Could you change the likes of these to be `0` instead of having a floating `px` at the end

src/definitions/elements/step.less
```diff
@@ -153,9 +153,6 @@
   flex: 0 1 auto;
   align-self: @iconAlign;
 }
-.ui.steps .step > .icon ~ .content {
```
hammy2899: Why was this removed?

prudho: The rule above was a duplicate as `flex-grow` is a part of the `flex` property.

hammy2899: Ok cool

src/definitions/elements/label.less
prudho: Just this one or all of them ?

hammy2899: All of them, if you want you could do it for all files.

PR #282, opened 2018-12-03
[Global]: Remove deprecated jquery functions 

lubber-de:
## Description
jquery deprecated lots of methods. 3 of them are used widely in FUI. This PR prepares to work with a future jquery Version where those functions were removed.
You can also see this PR as a little step towards independency of jquery :slightly_smiling_face: 

- `$.isArray` is replaced by native `Array.isArray`
- `$.isFunction` and `$.isWindow` statements are adopted from jquery (the same wrapper as before, but put in FUI if not found in jquery)

### DIFF
--- a/src/definitions/globals/site.js
+++ b/src/definitions/globals/site.js
@@ -10,6 +10,13 @@
 
 ;(function ($, window, document, undefined) {
 
+$.isFunction = $.isFunction || function(obj) {
+    return typeof obj === "function" && typeof obj.nodeType !== "number";
+};
+$.isWindow = $.isWindow || function(obj) {
+    return obj != null && obj === obj.window;
+};
+
 $.site = $.fn.site = function(parameters) {
   var
     time           = new Date().getTime(),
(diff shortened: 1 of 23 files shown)


src/definitions/globals/site.js
```diff
@@ -10,6 +10,13 @@
 
 ;(function ($, window, document, undefined) {
 
+$.isFunction = $.isFunction || function(obj) {
```
prudho: Does this work if we use single component ?

lubber-de: Well, if you just include single component files, then currently you would need to have site.js aswell, yes.
I was trying not to invent another `wrapper_functions.js` or similiar, but use existing code.
As it is needed to ensure those functions are available before any other FUI code is loaded, site.js was my only option because it is compiled at the very beginning already.
Otherwise we would need to put those function into every single component, which is wrong and overkill in my pov. 
As long as we either rely on jquery or some wrapper of jquery functions into our own helper file (to avoid having big changes in the current code) then you always need to include at least site.js (or a new `helper/wrapper.js`) just as you do with jquery.

prudho: Unfortunately for now we'll have to put those functions in each module, to keep 100% compatibility with SUI, even if, and i agree with you, it's overkill (we already do it with debug functions).

lubber-de: I moved/copied the function declaration accordingly now. See latest commit

PR #288, opened 2018-12-04
[Modal] Dimmer-Click closed Modal when allowMultiple=true even if onHide()=false 

### LINKED ISSUE #284: [Modal] - Using allowMultiple = true
# Bug Report

Closing Modal, when allowMultiple is set to true, via the Dimmer area of the page will close Modal even if onHide callback return false.

## Steps to reproduce
1. Create Modals with allowMultiple = true;
2. Create onHide callback that return false;
3. Open first modal and click dimmer area to close modal

## Expected result
onHide return false should prevent modal from closing.

## Actual result
onHide is called but Modal still close.
Modal does not reopen after.

## Testcase
https://jsfiddle.net/ibelar/h8a3se4o/

## Version
2.6.4

lubber-de:
## Description
Trying to close a modal by clicking in the dimmer area but having `return false` at the `onHide`-Method did still close the modal if `allowMultiple` was set to true

### Unfixed
https://jsfiddle.net/ibelar/h8a3se4o/

### Fixed
https://jsfiddle.net/bagwd05n/

## Closes
#284

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -287,6 +287,10 @@ $.fn.modal = function(parameters) {
               module.debug('Dimmer clicked but mouse down was initially registered inside the modal');
               return;
             }
+            if(settings.onHide.call(element, $(this)) === false) {
+              module.verbose('Hide callback returned false cancelling hide');
+              return false;
+            }
             var
               $target   = $(event.target),
               isInModal = ($target.closest(selector.modal).length > 0),
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -287,24 +287,22 @@ $.fn.modal = function(parameters) {
               module.debug('Dimmer clicked but mouse down was initially registered inside the modal');
               return;
             }
-            if(settings.onHide.call(element, $(this)) === false) {
-              module.verbose('Hide callback returned false cancelling hide');
-              return;
-            }
             var
               $target   = $(event.target),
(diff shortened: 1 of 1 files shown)


prudho:
Modal event `onHide` seem to be triggered multiple times the second time you open the modal.

lubber-de:
You mean when clicking on the button within the first modal after opening the first (just because the fiddle does not allow to close any modal when it is openened once) ? Then it's wanted behavior because the onhide is called for each opened modal when clicking on the dimmer. The fiddle defines the onhide method globally in the modals prototype so it is assigned to every modal

prudho:
I have set `return true` in the onHide callback.

lubber-de:
:scream: You are right, something is wrong with the attached events... i'll take care of it. Thanks for finding it out

lubber-de:
@prudho I fixed it now, was quite nasty. It was indeed totally bugged when using multiple modals, because the return false of onHide was not respected within the loop in `hideAll` ... Anyway, please review again, the jsfiddle from above should have been automatically updated by the included link to modal.js the latest repo-branch.
Please note: onHide (an so the alert) will be triggered for each open modal. So:
- onHide returns true and 2 modals are open -> 2 alerts will be shown
- onHide returns false and 2 modals are open -> only 1 alert is shown 
- onHide returns true or false and 1 modal is open -> only 1 alert is shown

lubber-de:
:roll_eyes: Also fixed dimmer-clicknow when allowMultiple=false and onHide return false...

PR #295, opened 2018-12-11
[Checkbox] Support onChange for cursor keynav

lubber-de:
## Description
Changing Radio-Button Selections via Cursor Key-Navigation was not triggering the onChange Event

## Testcase
- Click an option of the Radio Button Group
- User cursor keys to switch between the options

### Unfixed
- Message is not updated, because onChange event is not fired
https://jsfiddle.net/n9sLf81w/

### Fixed
- Message gets updated just as you would have clicked on another option
https://jsfiddle.net/ujkdreqs/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6676

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -175,6 +175,11 @@ $.fn.checkbox = function(parameters) {
         },
 
         event: {
+          change: function(event) {
+            if( !module.should.ignoreCallbacks() ) {
+              settings.onChange.call(input);
+            }
+          },
           click: function(event) {
             var
               $target = $(event.target)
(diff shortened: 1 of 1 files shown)


ColinFrick:
You can change the value of a radio box with the keyboard even if `beforeChecked` / `beforeUnchecked` returns false.

Option 2 is "disabled":
https://jsfiddle.net/2qrxm9kh/

lubber-de:
@ColinFrick :open_mouth: That was quite challenging, but keynav is not stopped if next option does not allow to be checked or current option does not allow to be unchecked
https://jsfiddle.net/ujkdreqs/

Issue #268, opened 2018-11-28
[SECURITY] event-stream incident


dhaavi:
There was a security issue with the npm package `event-stream`.

Original issue: https://github.com/dominictarr/event-stream/issues/116#issuecomment-441759047
Semantic issue: https://github.com/Semantic-Org/Semantic-UI/issues/6687

Please update `event-stream` to version 3.3.4:
> By this time fixes are being deployed and npm has yanked the malicious version. Ensure that the developer(s) of the package you are using are aware of this post. If you are a developer update your event-stream dependency to event-stream@3.3.4. This protects people with cached versions of event-stream.

hammy2899:
@dhaavi See the PR I just created. https://github.com/fomantic/Fomantic-UI/pull/269

dhaavi:
Looks great!
(I am not experienced with node, so I can't tell you how effective this fix is.)

etshy:
what if I do `npm update` ? The lock file block the maximum version to 3.3.4 or it will download the 3.3.6 (which is infected if I remember correctly the version number) ?

If the package-lock works the same as composer.lock, it will force this version on install only, If I'm not mistaken. On update it will search dor the highest version that match with the package.json

I also saw this issue yesterday, and seems some packages are updating to 4.0.1 (when the flatmap dep was removed)

dhaavi:
> it will download the 3.3.6 (which is infected if I remember correctly the version number) ?

No, because

> npm has yanked the malicious version

The version change is there to defeat caches. For example, if you've downloaded the infected version and do `git pull` and `npm ...` it should remove the infected version cached locally.

etshy:
Oh ok, I missed the information about npm removing the malicious versions.

Nice to know it should be fixed. I don't work with crypto but still good to not have malicious code.

Atulin:
Wouldn't it be a good idea to think of removing ``prompt-sui`` in the future? From what I've seen, it's vanity stuff at best.

Overall, I'd consider removing as many dependencides as possible, but that's a matter for another discussion.

hammy2899:
@Atulin We currently have plans to rewrite the build process when we do 3.0 but that is a whole other project. When we do this we will be getting rid of a lot of the dependencies.

Atulin:
That + removing the dependency on jQuery would be a dream come true. Glad there are steps being taken in that direction 👌

PR #302, opened 2018-12-15
[Build] Travis npm fix

lubber-de:
## Description
The current Travis CI BUild process fails, because npm gets new sha512 hashes which do not match with the sha1 ones stored in package-lock.json.
Simply removing this file from the repo fixes it.


lubber-de:
@hammy2899  I am unsure, if this would reintroduce #268 , so you may just create and push  a fresh working package-lock.json (also npm suggests to keep the lock file in the repo) instead and decline this PR.

lubber-de:
As discussed and since @prudho  seemed to fix it, lets forget about my attempt here

PR #317, opened 2018-12-19
[Emojis] EmojiOne-awesome 4.0 and Tweemoji v2 support

lubber-de:
## Description
This PR adds support to use emojis from 
- https://www.emojione.com 4.0 under their free license model
- Tweemoji v2 from https://twemoji.twitter.com/ under CC-BY 4.0 license

By default all icons are linked to CDN and are not part of the FUI repo itself. If needed, download your personal icon set from the respective websites, host it locally and change the `@emojipath` in `emoji.variables` of the theme.

I simply ported https://github.com/emojione/emojione/tree/master/lib/emojione-awesome to FUI.
Emojis are included inline as they usually are textpart replacements.

## Usage
You use the `em` tag for emojis giving the official short name as a `data-emoji` attribute. Normal strings like `wink` aswell as colon surrounded strings like `:wink:` are supported

### Basic
```html
<em data-emoji="wink"></em>
<em data-emoji=":wink:"></em>
```
#### Sizes
Available Sizes are `small` (default), `medium`, `large` and `big`
```html
<em data-emoji="wink" class="medium"></em>
<em data-emoji="wink" class="large"></em>
<em data-emoji="wink" class="big"></em>
```
#### Disabled
```html
<em data-emoji=":wink:" class="disabled"></em>
```
#### Loading
```html
<em data-emoji="wink" class="loading"></em>
```

> **Hint:** To easily get emoji-shortnames you need, visit https://www.emojione.com/emoji/v4 . Click on your desired emoji and the following page shows every information you need.

For a simple conversion of existing strings, which include the short names already, you may use this code snippet to convert them into the FUI `<em>` format:

#### Javascript Frontend / NodeJS Backend
```javascript
var message="Hi honey :WiNK:, please look at this :cool: thing!";
message = message.replace(/(:[a-zA-Z0-9_]+:)/g,function(match) {
    return '<em data-emoji="'+match.toLowerCase()+'"></em>';
});
```

#### PHP Backend example
```php
$message="Hi honey :WiNK:, please look at this :cool: thing!";
$message = preg_replace_callback('/(:[a-zA-Z0-9_]+:)/',function($match){
    return '<em data-emoji="'.strtolower($match[0]).'"></em>';
},$message);
```

Both examples return
```html
Hi honey <em data-emoji=":wink:"></em>, please look at this <em data-emoji=":cool:"></em> thing!
```

### EmojiOne
emojione

emoji_disabled_loading

### Tweemoji
If you want Tweemoji from twitter instead of EmojiOne, just change the theme for `@emoji` in `theme-config` to `'twitter'` and you get the following different emojis

image


hammy2899:
This is cool but I'm not 100% on it, I have been thinking on how to add emoji's or a while now but I can't think of a perfect solution. I want to make it so users can pick which library they want to support like native, twitter, emojione etc.

lubber-de:
@hammy2899 Yes, this is possible using a different theme. The emojione specific code is in emoji.overrides
just as the font-awesome icons are in icon.overrides.
I also prepared the `@emoijpath` variable so one can change it to another provider.
I can take a look at how the other providers would make it possible to be integrated into FUI easily, so one could change `@emoji: 'default'`  to `@emoji: 'twitter'` and the emojis will be used from the twitter theme then.

hammy2899:
@lubber-de Yeah that might not be a bad idea.

I do have a few things I want to change 😛 

Also I don't think this will be a thing we add for 2.7 maybe a later change.

src/definitions/elements/emoji.less
```diff
@@ -0,0 +1,67 @@
+/*!
+ * # Fomantic UI - Emoji
+ * http://github.com/fomantic/fomantic-ui/
```
hammy2899: Capitalize the repo and SSL
`http://github.com/fomantic/fomantic-ui/` => `https://github.com/fomantic/Fomantic-UI/`

src/definitions/elements/emoji.less
```diff
@@ -0,0 +1,67 @@
...
+ *
+ *
```
hammy2899: Remove one line so there is only 2 blank lines

src/definitions/elements/emoji.less
```diff
@@ -0,0 +1,67 @@
...
+ * Released under the MIT license
+ * http://opensource.org/licenses/MIT
```
hammy2899: Add link to license in repo
`https://github.com/fomantic/Fomantic-UI/blob/master/LICENSE.md`

src/themes/default/elements/emoji.variables
```diff
@@ -0,0 +1,20 @@
+/*******************************
+             Emoji
+*******************************/
+
+/*--------------
+    Path
+---------------*/
+@emojiPath: "https://cdn.jsdelivr.net/emojione/assets/4.0/png/128/";
```
hammy2899: Load this as a local asset not via CDN

Take a look at icons for reference https://github.com/fomantic/Fomantic-UI/blob/master/LICENSE.md

lubber-de: As far as i understood, it would be against their license to include the files in our own repository, but i cannot find the terms about this anymore atm. Anyhow: Should we really include those tons of files in the repo instead of letting the user download the icons on his ownfrom their website when wanted, and adjust the path to local assets on his own then?

lubber-de:
Supporting Twitter Tweemojies was as easy as i expected :grinning: I just had to change the `@emojipath`
I commited appropriate files for the twitter theme and added some info/screenshot in the description

src/definitions/elements/emoji.less
lubber-de: :confused: If i remove 1 line, then its 1 line left instead of 2 (because there are only 2? Anyhow, this header is a copy from icon.less, that`s why ;)

src/definitions/elements/emoji.less
lubber-de: Will do

src/themes/default/elements/emoji.variables
hammy2899: Yeah your right (https://www.emojione.com/licenses/free) this could be a problem

lubber-de:
messed up base branch... new PR is #326

PR #326, opened 2018-12-23
feat(emoji): add emoji component using tweemoji & joypixels

lubber-de:
## Description
This PR adds support to use emojis from 
- https://www.joypixels.com 5 under their free license model
- Tweemoji v12 from https://twemoji.twitter.com/ under CC-BY 4.0 license

By default all icons are linked to CDN and are not part of the FUI repo itself. If needed, download your personal icon set from the respective websites, host it locally and change the `@emojipath` in `emoji.variables` of the theme.

I simply ported https://github.com/joypixels/emoji-toolkit/tree/master/lib/joypixels-awesome to FUI.
Emojis are included inline as they usually are textpart replacements.

## Usage
You use the `em` tag for emojis giving the official short name as a `data-emoji` attribute. Normal strings like `wink` aswell as colon surrounded strings like `:wink:` are supported

### Basic
```html
<em data-emoji="wink"></em>
<em data-emoji=":wink:"></em>
```
#### Sizes
Available Sizes are `small` (default), `medium`, `large` and `big`
```html
<em data-emoji="wink" class="medium"></em>
<em data-emoji="wink" class="large"></em>
<em data-emoji="wink" class="big"></em>
```
#### Disabled
```html
<em data-emoji=":wink:" class="disabled"></em>
```
#### Loading
```html
<em data-emoji="wink" class="loading"></em>
```

> **Hint:** To easily get emoji-shortnames: 
> - Check https://www.webfx.com/tools/emoji-cheat-sheet/ . Click on any desired emoji to copy the shortname into the clipboard
> - visit https://www.joypixels.com/emoji/v5 . Click on your desired emoji and the following page shows every information you need.

For a simple conversion of existing strings, which include the short names already, you may use this code snippet to convert them into the FUI `<em>` format:

#### Javascript Frontend / NodeJS Backend
```javascript
var message="Hi honey :WiNK:, please look at this :cool: thing!";
message = message.replace(/(:[a-zA-Z0-9_]+:)/g,function(match) {
    return '<em data-emoji="'+match.toLowerCase()+'"></em>';
});
```

#### PHP Backend example
```php
$message="Hi honey :WiNK:, please look at this :cool: thing!";
$message = preg_replace_callback('/(:[a-zA-Z0-9_]+:)/',function($match){
    return '<em data-emoji="'.strtolower($match[0]).'"></em>';
},$message);
```

Both examples return
```html
Hi honey <em data-emoji=":wink:"></em>, please look at this <em data-emoji=":cool:"></em> thing!
```

### JoyPixels
joypixels

joypixels_disabled_loading

### Tweemoji
If you want Tweemoji from twitter instead of JoyPixels, just change the theme for `@emoji` in `theme-config` to `'twitter'` and you get the following different emojis

image


hammy2899:
Going to wait until 2.8 or 3.0 to release this.

exoego:
@lubber-de 
What is a design intent to use `data-emoji` attribute instead of `class` like `icon` do so?
To minimize tech-debt of canonicalizing emoji name into css class?

lubber-de:
@exoego The reason for this is the same as for icons, what we discussed a while back and , from the current pov, will be used for icons in v3 as well: Using classes will most probably interfere with existing classnames which have a different meaning. The amount of icons and emojis is too huge to make sure the existing component-classnames won't get disturbed.
But i am totally open with another proposal. 🙂 
There were also ideas to use dashes instead of spaces to make this clear and then also usably as classnames. 
`class="small image icon"` vs (interferes with `image`
`data-icon="small image"` vs 
`class="japan flag emoji"` (interferes with `flag`) vs
`data-emoji="japan flag emoji"`

Possible different approach:
`class="image-icon"`
`class="japan-flag emoji"`

In the end i just would like to have the same logic for `icon` and `emoji`..
As emoji is a new component i wanted to go the same way that `icon` will be in v3.

exoego:
> The amount of icons and emojis is too huge to make sure the existing component-classnames won't get disturbed.

Understood, that is a sound reason 👍

prudho:
So, if my understanding is correct, external URLs will not be loaded if you don't use those Emojis right ?

hammy2899:
@prudho should only load images of emojis used since the image is loaded on the emoji name class here https://github.com/fomantic/Fomantic-UI/blob/163d3e2a070ecfe2672dc009c23b93e246b3c279/src/themes/default/elements/emoji.overrides#L3089

lubber-de:
> So, if my understanding is correct, external URLs will not be loaded if you don't use those Emojis right ?

Yes

PR #261, opened 2018-11-26
[Colors] Central dynamic color map for every component

lubber-de:
## Description
This PR provides the base to create a central dynamic color palette in a single mixin-map instead of hardcoding every possible color in each component.
To stay backward compatible, this map currently holds the already existing global color variables.

Now it's possible to define only the colors and naming you want and each component will generate the appropriate component with the colors and those namings which are defined in the central map.

This way it's now easy to define completely new color-names without touching any component manually.

##  Example
If somebody does not like the basic FUI color palette but want to use only brand color namings instead, the `colors.less` needs to be created in the appropriate `{site}` folder of the theme folder having this example content:

```less
// those variables should be defined in your site.variables  
@pinkie       : #ED458B;
@lightPinkie  : #f9b8d2;
@pinkieText   : #EB81B4;
@twilight     : #8689c8;
@lightTwilight: #d19fe3;
@twilightText : #662d8a;

@colors: {
  @pinkiepie: {
    color          : @pinkie;
    light          : @lightPinkie;
    border         : #BE1D77; 
    background     : @lightPinkie;
    header         : @pinkie;
    boxShadow      : none;
    boxFloatShadow : none;
    text           : @pinkieText;   
    lightText      : @pinkieText;   
    hoverText      : @pinkieText;                              
    focus          : @pinkieText; 
    lightFocus     : @pinkieText; 
    down           : @pinkieText; 
    lightDown      : @pinkieText; 
    active         : @pinkieText; 
    lightActive     : @pinkieText; 
    shadow         : @pinkieText; 
    lightShadow     : @pinkieText; 
    hover          : saturate(darken(@pinkie, 5), 10, relative);
    lightHover     : saturate(darken(@lightPinkie, 5), 10, relative);
    ribbon         : darken(@pinkie,10);
    invertedRibbon : darken(@lightPinkie,10);
    isDark         : false;
    isVeryDark     : false;
  };
  @twilight: {
    color          : @twilight;
    light          : @lightTwilight;
    border         : #263773; 
    background     : @lightTwilight;
    header         : @twilight;
    boxShadow      : none;
    boxFloatShadow : none;
    text           : @twilightText;
    lightText      : @twilightText;
    hoverText      : @twilightText;
    focus          : @twilightText; 
    lightFocus     : @twilightText; 
    down           : @twilightText; 
    lightDown      : @twilightText; 
    active         : @twilightText; 
    lightActive     : @twilightText; 
    shadow         : @twilightText; 
[...]

PR #331, opened 2018-12-30
[Label] Remove hardcoded color exception "not()" for menu item labels

lubber-de:
## Description
Menu Item Labels have a different default color (darker grey) than normal labels. To not interfere with the new dynamic colors, which do not use `!important` anymore, i originally (in #230) had to add  `:not({color})` statements for each color to the menu item label.
Now, having implemented the dynamic color system #261, this, still hardcoded, part of `:not()` had to be removed from menu item labels to fully support individual color maps. Thus an increase of specificity for the labels was needed.

## Testcase
http://jsfiddle.net/7hp8m5og/

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -324,7 +324,7 @@
 .ui.menu .item > .floating.label {
   padding: @labelVerticalPadding @labelHorizontalPadding;
 }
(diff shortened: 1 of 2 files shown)


ColinFrick:
LGTM

PR #332, opened 2018-12-30
[Table] Make colors work on striped, basic, selectable, stacked, definition

lubber-de:
## Description
Table fixes:
- on `striped` tables any additional color like `error` on rows were ignored
- `definition` tables still had their column color set when stacked
- rows on `selectable`, `basic`, `striped` tables did not work everytime
- some single color cells on those tables had their background lost

That is now fixed by proper specificity.

> Btw, all of above issues are still the case in SUI 2.4.1, they are not related to the dynamic color PR, nor the `!important` removements we implemented in FUI 2.7.x

## Testcase
http://jsfiddle.net/nv6q8d5e/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5318

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -186,10 +186,10 @@
   }
 
   .ui.table:not(.unstackable) tr > th,
-  .ui.table:not(.unstackable) tr > td {
+  .ui.ui.ui.ui.table:not(.unstackable) tr > td {
     background: none;
-    border: none !important;
-    padding: @responsiveCellVerticalPadding @responsiveCellHorizontalPadding !important;
+    border: none;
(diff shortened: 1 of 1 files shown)


ColinFrick:
Should a green striped table be always the same color? Or should the striping be green (green, dark green, green, etc.)?

hammy2899:
@ColinFrick Good point, that would make more sense since it is "green striped" 🤔

lubber-de:
@ColinFrick @hammy2899 This is a good idea, but would be a new feature (supporting any color in combination with `striped table` to create "colored stripes").
This PR only covers single colored tr/td elements which override a `striped table`  background.

PR #333, opened 2018-12-31
[Button] Attached buttons were shown above dropdown menu list

lubber-de:
## Description
The static z-index of `attached buttons` prevented dropdown Menu list to be shown above, when a dropdown was created in another attached buttons container.

## Testcase
- Open the top dropdown, works fine
- Remove the CSS part, re-run the fiddle and open the upper dropdown button again to see the issue 
https://jsfiddle.net/w2jayvsf/

## Screenshot
semantic

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5355

### DIFF
--- a/src/themes/default/elements/button.variables
+++ b/src/themes/default/elements/button.variables
@@ -378,7 +378,7 @@
 @attachedOffset: -1px;
 @attachedBoxShadow: 0 0 0 1px @borderColor;
 @attachedHorizontalPadding: 0.75em;
-@attachedZIndex: 2;
+@attachedZIndex: auto;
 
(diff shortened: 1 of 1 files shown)


lubber-de:
@ColinFrick Yes, I was looking for every z-index setting before.  That's why i put those extra buttons `animated button` and `or`buttons into the fiddle :wink: to test their z-index is still working. The mentioned issue requested to support attached button**s** which this is also covered in the jsfiddle. 
If the change would not work then the extra buttons would not appear in the button group...but they do :slightly_smiling_face: , so from my POV this is covered and working

PR #335, opened 2018-12-31
[Sidebar] Respect a change of closable setting

lubber-de:
## Description
The `closable` setting is ignored in sidebar when the setting was changed after initialization.

## Testcase
- Keep the red sidebar open
- Click on "Fix sidebar"
- Click on "Toggle sidebar"
-> Sidebar should not close

#### Unfixed (original SUI issue)
https://jsfiddle.net/r7fws0nn

#### Fixed
https://jsfiddle.net/r7fws0nn/23/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5329

### DIFF
--- a/src/definitions/modules/sidebar.js
+++ b/src/definitions/modules/sidebar.js
@@ -175,12 +175,10 @@ $.fn.sidebar = function(parameters) {
         bind: {
           clickaway: function() {
             module.verbose('Adding clickaway events to context', $context);
-            if(settings.closable) {
-              $context
-                .on('click'    + elementNamespace, module.event.clickaway)
-                .on('touchend' + elementNamespace, module.event.clickaway)
-              ;
-            }
+            $context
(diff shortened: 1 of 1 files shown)


ColinFrick:
The other way around doesn't work either. If you set closable to false, open it, then set closable to true from inside the sidebar doesn't bind the `clickaway` events:
https://github.com/fomantic/Fomantic-UI/blob/b3189443f2107edc9f6ffaf2739a89a3416ff6a5/src/definitions/modules/sidebar.js#L176-L184
https://jsfiddle.net/Korinu/5obdax1z/

lubber-de:
@ColinFrick As suggested, the clickaway is not bound in any case. This was the easiest (in in my pov cleanest) solution. otherwise i had to bin/unbind the clickway every time the closable settings changes and i   feared other bugs then.

PR #336, opened 2018-12-31
[List,Icon] List Item Icon Rotates Around Wrong Point when using Loading class

lubber-de:
## Description
When adding  the `loading` class to an icon in a list item, the entire icon rotates around a wrong point rather than rotating about it's own center.

## Testcase
http://jsfiddle.net/pn75L2tj/
- first list item shows the issue
- second item is fixed
- third one is normal item without loading class (to show alignment)

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6717

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -109,9 +109,11 @@ ol.ui.list ol li,
   display: table-cell;
   margin: 0;
   padding-top: @iconOffset;
-  padding-right: @iconDistance;
-  vertical-align: @iconContentVerticalAlign;
   transition: @iconTransition;
+  &:not(.loading) {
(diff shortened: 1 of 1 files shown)


src/definitions/elements/list.less
```diff
@@ -109,9 +109,11 @@ ol.ui.list ol li,
   display: table-cell;
   margin: 0;
   padding-top: @iconOffset;
-  padding-right: @iconDistance;
-  vertical-align: @iconContentVerticalAlign;
   transition: @iconTransition;
+  &:not(.loading) {
```
ColinFrick: Uh fancy nesting ;)

PR #339, opened 2018-12-31
[Dropdown] Fix simple dropdown positioning in vertical menus

lubber-de:
## Description
 A simple dropdown in a vertical menu had wrong positioning, thus nearly unusable.
Also a submenu had second submenu dropdown icon misplaced

## Testcase
http://jsfiddle.net/Lnba5udj/1/
The fix cannot be applied directly in the fiddle CSS here, because you need to remove the !important flags from the original library manually in the debugger console (which i did for testing and below screenshot)
I added the JS-Dropdown Menus in the fiddle to test and make sure the changes do not affect non-simple dropdowns.

### Unfixed
simple_dropdown_position

### Fixed
simple_dropdown_position_fixed

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5454
https://github.com/Semantic-Org/Semantic-UI/issues/5657

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -284,7 +284,7 @@
 }
 .ui.vertical.menu .dropdown.item .menu {
   left: 100%;
-  min-width: 0;
+  min-width: max-content;
   margin: 0 0 0 @dropdownMenuDistance;
   box-shadow: @dropdownVerticalMenuBoxShadow;
   border-radius: 0 @dropdownMenuBorderRadius @dropdownMenuBorderRadius @dropdownMenuBorderRadius;
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -284,6 +284,8 @@
 }
 .ui.vertical.menu .dropdown.item .menu {
   left: 100%;
+  /* IE needs 0, all others support max-content to show dropdown icon inline, so keep both settings! */
+  min-width: 0;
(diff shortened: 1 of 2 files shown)


lubber-de:
@ColinFrick I implemented a hack, so the dropdown icon (the only reason i changed the min-width setting at all) is now also correctly displayed inline in IE11. 
http://jsfiddle.net/Lnba5udj/2/ -> remember the initial reason of this PR (making the simple dropdown menu appear at the correct right position) is not covered in the fiddle, because the `!important` values have to be removed from the included library version. But here is a fixed screenshot from IE after having fun with console hacking (which is still a PITA in IE :roll_eyes: )
simple_dropdown_position_ie

However, as you can see, beside from this PR, a JS-dropdown having submenus in a vertical menu still displays the dropdown icon wrongly shifted. That's because the `display` style setting is taken by the transition.js fetching the value from the DOM and it does not seem to work with special IE `display`settings. Transition always receives `block` and also sets `!important` inline.  Just for this special case i really don't want to hack transition.js, which probably would break several other components.
I would move this additional issue to another PR (if we ever can fix this without massive IE changes to transition.js i am not sure) and leave it open as it is

lubber-de:
Found this PR https://github.com/Semantic-Org/Semantic-UI/pull/5685 which also suggested to use the max-content setting for vertical dropdowns, because the normal dropdown already uses that setting since ages.

PR #355, opened 2019-01-03
[List] Add suffixed ordered list variation

exoego:
## Description

Ordered list has conventionally dot or something after the end of counter like

> 1. 2019
> 2. 2018
> 3. 500px

But SUI/FUI's ordered list does not follow the convention and omit dot.
This may lower readability specially when list item starts with number (below example).

> 1&ensp;2019
> 2&ensp;2018
> 3&ensp;500px

So This PR 

* Add `.suffixed.ordered.list` variation that add the dots.
* Add LESS variables that allows developers to change separator and suffix like '1-2.' or '1.3)'.

### Before
https://jsfiddle.net/Lsawqmj6/

### After
https://jsfiddle.net/fso94dmx/1/
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4738

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -651,12 +651,6 @@ ol.ui.list li:before,
   opacity: @orderedCountOpacity;
 }
 
-ol.ui.basic.list li:before,
-.ui.basic.ordered.list .list > .item:before,
-.ui.basic.ordered.list > .item:before {
-  content: @orderedCountContentBasic;
-}
-
 ol.ui.inverted.list li:before,
 .ui.ordered.inverted.list .list > .item:before,
 .ui.ordered.inverted.list > .item:before {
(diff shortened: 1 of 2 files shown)


hammy2899:
@ColinFrick Yes I agree, this is a breaking change.

lubber-de:
The correct grammatical style for numbered outline list is **not** having a trailing dot, so FUI/SUI is right regarding this.
See https://en.wikipedia.org/wiki/Outline_(list)#Decimal_outline

So instead of making this a breaking change in other cases like the horizontal list example
image
why don't we just do it vice versa?
- Keep the previous logic for `ordered list` -> Not a breaking change anymore
- Instead of `basic` add a new feature `suffixed` or `trailing` or `alphanumeric` (because this is the correct named grammar type) type See https://en.wikipedia.org/wiki/Outline_(list)#Alphanumeric_outline with provides the additional dot suffix

exoego:
I think adding alphanumeric variation sounds better.
If others agree, I will update PR.

hammy2899:
Yeah I agree with this, I personally think `suffixed` makes the most sense.

PR #357, opened 2019-01-03
[Transition] Remember inline style display and dont show hidden objects like <script>

lubber-de:
## Description
Transition did always detect "block" as display value and did not respect or remembered an initial given inline block style which was probably wanted by the user. For example `inline` or even `none`, if an element within a transition should definately not be shown. 
Not only because of this, but it was also displaying `script` or other unwanted tags :roll_eyes: 
`script`, `link` or `style` tags are now skipped all the time.

Especially for accordion it turns out to be usefull that transition should be skipped if the item has initially been set to be hidden. This is true in most cases for the to be animated element itself, but in an accordion there could be some elements within the container where those elements should be kept hidden.

Therefore i added a new setting

 ```javascript
skipInlineHidden: false ; // (default false to stay backward compatible) 
```
to the transition module.
The accordion module uses this now to still hide any child element of a accordion container (not the container itself!) in case a child element was already initially hidden.
See the source of the fiddle below for better explanation :wink:

For additional testing i injected the new transition.js into all of my local `gh-pages`-branch to make sure no other components had negative side effects (way too many examples otherwise to put into a fiddle)

If you also want to test it, you can simply:

- Checkout gh-pages of the https://github.com/fomantic/Fomantic-UI-Docs/tree/gh-pages
- copy transition.js from this PR to /dist folder
- Change 
```html
<script src="/dist/semantic.min.js"></script>
<script src="/dist/transition.js"></script>
```
In the following docs html files: accordion, popup,dimmer,transition,form, search, modal, toast, visibility,modal,dropdown (those modules make use of transition.js , thats why. The docs itself use transition.js on every page)

### Broken
http://jsfiddle.net/qLp3md9n/

### Fixed
http://jsfiddle.net/qLp3md9n/2/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/2988
https://github.com/Semantic-Org/Semantic-UI/issues/3171
https://github.com/Semantic-Org/Semantic-UI/issues/6735

### DIFF
--- a/src/definitions/modules/accordion.js
+++ b/src/definitions/modules/accordion.js
@@ -189,12 +189,13 @@ $.fn.accordion = function(parameters) {
               $activeContent
                 .children()
                   .transition({
-                    animation   : 'fade in',
-                    queue       : false,
-                    useFailSafe : true,
-                    debug       : settings.debug,
-                    verbose     : settings.verbose,
-                    duration    : settings.duration
+                    animation        : 'fade in',
+                    queue            : false,
+                    useFailSafe      : true,
+                    debug            : settings.debug,
+                    verbose          : settings.verbose,
+                    duration         : settings.duration,
+                    skipInlineHidden : true
                   })
               ;
             }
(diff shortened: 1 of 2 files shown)


lubber-de:
Just found out i had to make an explicit check for `script`, `link` or `style` tags on mobile. So in case the above fiddle does still show the JS Code, either wait until jsdelivr has renewed its cache or put the whole transition.js in the fiddle before running it...

lubber-de:
Sorry, It seems I broke it for the docs.. I put this on hold until I fixed that also...

lubber-de:
Fixed it :sweat_drops: ...
Updated the fiddle link and description. 
Ready for review :smile:

PR #372, opened 2019-01-08
[Formvalidation] Radio buttons do not validate if the first is disabled

lubber-de:
## Description

- If the first option of a radio group was disabled the whole validation of that group was skipped
- Additional fix: `is empty` always returned `false` for empty radio button groups
- I also removed unnecessary sets of `fieldValid=true`..because it is initially `true` in all cases anyway

## Testcase
Follow the short info in the fiddles (TL;DR: press both buttons :smile: )
### Broken
https://jsfiddle.net/u2abzw3r/1/

### Fixed
https://jsfiddle.net/u2abzw3r/2/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6511

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -271,7 +271,7 @@ $.fn.form = function(parameters) {
             if(!$field || $field.length === 0) {
               return true;
             }
-            else if($field.is('input[type="checkbox"]')) {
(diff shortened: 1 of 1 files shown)


lubber-de:
The testcase fiddle urls were wrong...sorry for this, i changed it to the correct URLs in the description now.

PR #387, opened 2019-01-12
[Formvalidation] Input array validations should respect each single field on its own

lubber-de:
## Description
If a form uses array inputs by appending brackets to their names ( `name="myfield[]"`) the validation on those fields was always done for the whole field group. Means, if the first field was valid, it validated for the whole group which is wrong.
This behavior is now fixed, so even if multiple input fields have the same name to act as array inputs, the validation is done on each field separately and also the error class is set/removed to each field individually.
@prudho was already providing a [PR for that in SUI](https://github.com/Semantic-Org/Semantic-UI/pull/6370) some time ago, but the individual error display handling was missing there.

> Hint: If you ask yourself while reviewing, why i added an `internal` parameter to the `add.prompt()`and `validate.rule()` functions: This is to make sure those functions are still callable  as behavior from separate JS (for example: `$('.foo').form('validate rule', argumentOne, argumentTwo)` so they do not break existing code

### Before
http://jsfiddle.net/a1cnph4f/
Try to click on submit to invalidate each the form. If any of the multiple fields is  invalid it won't get the error class.

### After
http://jsfiddle.net/a1cnph4f/1/
Correct behavior now: error class for each invalid field
 I also put a normal (single) field into the fiddle to make sure the usual behavior is still working.

Special testcase: Radiobuttons are also working (which **need** the behavior)
http://jsfiddle.net/47Lysubr/

### Before
inputarrays_wrong

### After
inputarrays_right

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6368
https://github.com/Semantic-Org/Semantic-UI/issues/2683
https://github.com/Semantic-Org/Semantic-UI/issues/2688
https://github.com/Semantic-Org/Semantic-UI/issues/5469
https://github.com/Semantic-Org/Semantic-UI/issues/829

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -704,7 +704,7 @@ $.fn.form = function(parameters) {
             }
             validation = $.extend({}, validation, newValidation);
           },
-          prompt: function(identifier, errors) {
+          prompt: function(identifier, errors, internal) {
             var
               $field       = module.get.field(identifier),
               $fieldGroup  = $field.closest($group),
@@ -716,9 +716,11 @@ $.fn.form = function(parameters) {
               : errors
             ;
             module.verbose('Adding field error state', identifier);
-            $fieldGroup
-              .addClass(className.error)
-            ;
+            if(!internal) {
+              $fieldGroup
+                  .addClass(className.error)
+              ;
+            }
             if(settings.inline) {
               if(!promptExists) {
                 $prompt = settings.templates.prompt(errors);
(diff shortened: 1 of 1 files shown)


lubber-de:
:thinking: I think, I forgot one case, will check and possibly add another commit in 2 hours approx .. Don't merge yet

lubber-de:
Alright, i had to fix the behavior to work with radio buttons as before 
Testcase including radio buttons:
http://jsfiddle.net/47Lysubr/
Please review again

morkpl:
@lubber-de 
**Problem**
`inline: true`

ice_screenshot_20190408-150041

lubber-de:
@morkpl Thanks, I'll investigate

Clazherx:
maybe i can ask

> @lubber-de
> **Problem**
> `inline: true`
> 
> ice_screenshot_20190408-150041

this is fixed now? because im using actual release (2.8 ) and i have the same problem. anyway, thanks for such a great framework

lubber-de:
@morkpl 
@Clazherx 
Thanks for the reminder
👀 Seems i forgot about this...(because it's not a separate issue but "just" a comment inside a merged PR

Clazherx:
@lubber-de thanks to you for the reply.
maybe i will post this bug as a separated issue.

PR #388, opened 2019-01-13
[Label] Right icon should have correct margin when placed after text

### LINKED ISSUE #222: [Label] Icons should have correct margin when placed after the text
# Bug Report

### Original SUI Issue

> This works:
> 
> ```
> <a class="ui label">
>   <i class="mail icon"></i>
>   Mail
> </a>
> ```
> 
> ```
> <a class="ui label">
>   Mail
>   <i class="mail icon"></i>
> </a>
> ```
> 
> image
> 
> 
> ```
> <a class="ui label">
>   Tag
> </a>
> ```
> 
> image
> 

### Original SUI PR 

exoego:
## Description

This PR is based on https://github.com/Semantic-Org/Semantic-UI/pull/6210 by @w96k.

* Add `icon label` variation that allows to put an icon without text.
* Add `right icon label` variation that allows to put an icon in the right side of label with correct margin. 
* Add `left icon label` variation to let `delete icon` and `close icon` be placed in left with correct margin for backward-compatibility
* ~Add `hoverable` class that makes icon to be used for action, as same style as current `close`/`delete` icon have.~

### Before
https://jsfiddle.net/oLc3pqeu/

### After
https://jsfiddle.net/y4L7cadb/

### Normal Label
image

### Ribbon Label
image

## Closes
#222 
https://github.com/Semantic-Org/Semantic-UI/issues/5446
https://github.com/Semantic-Org/Semantic-UI/issues/5619

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -101,6 +101,7 @@ a.ui.label {
 }
 
 /* Backward compatible positioning */
+.ui.label:not(.icon) > .close.icon,
 .ui.label:not(.icon) > .delete.icon {
   margin: 0 0 0 @deleteMargin;
 }
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -105,6 +105,20 @@ a.ui.label {
   opacity: 1;
 }
 
+/* Label for only an icon */
+.ui.icon.label > .icon {
+  margin: 0 auto;
+}
+
+/* Right Side Icon */
+.ui.right.icon.label > .icon {
+  margin: 0 0 0 @iconDistance;
+}
+.ui.right.icon.label > .close.icon,
+.ui.right.icon.label > .delete.icon {
+  margin: 0 0 0 @deleteMargin;
+}
+
 /*-------------------
        Group
 --------------------*/
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -91,18 +91,16 @@ a.ui.label {
   margin: 0 @detailIconDistance 0 0;
 }
 
-/* Removable and Hoverable label */
+/* Removable label */
 .ui.label > .close.icon,
-.ui.label > .delete.icon,
-.ui.label > .hoverable.icon {
+.ui.label > .delete.icon {
   cursor: pointer;
   font-size: @deleteSize;
   opacity: @deleteOpacity;
   transition: @deleteTransition;
 }
 .ui.label > .close.icon:hover,
-.ui.label > .delete.icon:hover,
-.ui.label > .hoverable.icon:hover {
+.ui.label > .delete.icon:hover {
   opacity: 1;
 }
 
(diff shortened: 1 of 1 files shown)


exoego:
@lubber-de 
Thanks for backward-compatibility suggestion!!

However, close/delete icon will have incorrect margin when placed before text like
image
because of margin defined in 
```
/* Backward compatible positioning */
.ui.label:not(.icon) > .close.icon,
.ui.label:not(.icon) > .delete.icon {
  margin: 0 0 0 @deleteMargin;
}
```


Do you have idea to fix this and keep backward-compatibility all together?


BTW, there is `times` icon, as an alias of delete/close icon. So there is a workaround if users want to use `x` on left.
image

lubber-de:
@exoego Having a close/delete to the left of a label was never proposed part of the framework, so although this is a valid visual point, it's not a breaking change, because old used code (if anywhere used by somebody anyway) like
```html
<div class="">
    <i class="delete icon"></i>
  Delete me
</div>
```
will work/look (as ugly) as before

But it is a valid point. So to make this also work , what about inventing an explicit `.ui.left.icon.label` (for users who definately want to have those icons also displayed to the left)
adding it accordinly in `label.less` here:
```css
/* Icon */
.ui.left.icon.label,
.ui.label > .icon {
  width: auto;
  margin: 0 @iconDistance 0 0;
}
```
Also in Addition:
Respect the different `@deleteMargin`, if those icons are used to the left

```css
.ui.label.left.icon > .close.icon,
.ui.label.left.icon > .delete.icon {
  margin: 0 @deleteMargin 0 0;
}
```
See here: https://jsfiddle.net/6Lmh2k9p/1/

When doing FUI 3.0, we may take a look at it again to simplify the logic (at least getting rid of the separate `@deleteMargin` or have an additional class for it)

src/definitions/elements/label.less
```diff
@@ -90,21 +90,37 @@ a.ui.label {
...
-  margin-right: 0;
-  margin-left: @deleteMargin;
   font-size: @deleteSize;
   opacity: @deleteOpacity;
   transition: @deleteTransition;
 }
-.ui.label > .delete.icon:hover {
+
+/* Backward compatible positioning */
+.ui.label:not(.icon) > .delete.icon {
```
lubber-de: `.ui.label:not(.icon) > .close.icon,` support missing

exoego: Added.

exoego:
I agree to keep backward compatibility as much as possible during `v2.x`. 

Updated code, PR description, and example in docs(https://github.com/fomantic/Fomantic-UI-Docs/pull/72).

exoego:
Removed `hoverable` variation to keep this PR scope small.

Initially I thought `hoverable` is required to keep delete/close icon style for backward-compatiblity and broaden the style to other icon.
However, it is no longer required for backward-compatibility.

I think tha label module should focus to `displays content classification` as stated in the docs. 
`hoverable` or such dynamic behavior is good to be implemented in button module.

PR #397, opened 2019-01-15
[Step] make steps wrap by default as intended for single step

lubber-de:
## Description
A huge amount of steps were overflowing the viewport if the amount of steps was not explicit given as classname. It was already intended to wrap (a single step already got `flex-wrap:wrap`), but it was missing for grouping by using `steps`.

This is ignored when `unstackable steps` is used because a possible overflow is wanted behavior.

If a step amount is given to the group (`six steps`), the behavior stays as before: Each step will reduce itself to stay inline. (Also part of the testcase below)

## Testcase
http://jsfiddle.net/v8yrgs6t/
Remove the CSS to see the issue

### Before
image

### After
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4880
https://github.com/Semantic-Org/Semantic-UI-React/issues/1114

### DIFF
--- a/src/definitions/elements/step.less
+++ b/src/definitions/elements/step.less
@@ -29,6 +29,7 @@
 .ui.steps {
   display: inline-flex;
   flex-direction: row;
+  flex-wrap: wrap;
   align-items: stretch;
   margin: @stepMargin;
   background: @stepsBackground;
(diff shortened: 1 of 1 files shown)


lubber-de:
@hammy2899 Done. New fiddle http://jsfiddle.net/v8yrgs6t/

PR #398, opened 2019-01-15
[Segment] Introducing stackable horizontal segments

lubber-de:
## Description
Added `stackable` variant to `horizontal segments` to make those segments wrap if needed

## Testcase
http://jsfiddle.net/tmxLgno1/2/
Remove CSS to see usual behavior without stackable support

### _Normal behavior_: `.ui.horizontal.segments`
image

### _New behavior_: `.ui.stackable.horizontal.segments`
image

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -412,6 +412,9 @@
   border-radius: @borderRadius;
   border: @border;
 }
+.ui.stackable.horizontal.segments {
+  flex-wrap: wrap;
+}
 
 /* Nested Horizontal Group */
 .ui.segments > .horizontal.segments {
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -443,7 +443,7 @@
 .ui.segments > .horizontal.segments:first-child {
   border-top: none;
 }
-.ui.horizontal.segments > .segment:first-child {
+.ui.horizontal.segments:not(.stackable) > .segment:first-child {
   border-left: none;
 }
 


hammy2899:
@ko2in @lubber-de this will be due to the horizontal divider still being applied.

lubber-de:
Unfortunetaly, you  cannot determine, if an element has wrapped by `flex-wrap`, but for usual horizontal segments the left border of the first segment is removed by default, causing the visual issue
```css
ui.horizontal.segments > .segment:first-child {
    border-left: none;
}
```
I suggest not doing this for `stackable horizontal segments` , then at least it looks consistent again

image

lubber-de:
@ko2in @hammy2899 changed it accordingly

ko2in:
@lubber-de Using pseudo element `::after` could also solve this problem. It even gives the same look for both left and right border of the segment wrapper.

You can check in this [fiddle](http://jsfiddle.net/Le6tbawq/3/).

lubber-de:
@ko2in Very nice visual trick to cover that border  :ok_hand: , but only if the background has exactly the same color for every segment inside. Also the top and bottom border should get covered (overlay the verticacl divider by one pixel)
I most of the cases the background might stay white anyway, but i would not force a white covering border (which would need to force a white default background also) for the whole group.
If the background changes or a color like `inverted red` is used in one of the segments this get's a bit weird. In any of that case you won't recognice the slighly bolder border on the left anyway

http://jsfiddle.net/Lpavt5ek/1/
http://jsfiddle.net/Lpavt5ek/2/

ko2in:
@lubber-de Yes. Not perfect with background. Your fix might be more suitable for this reason.

PR #404, opened 2019-01-17
[Label] Content in floating labels should not wrap whitespace

lubber-de:
## Description
If a `floating label` had whitespace inside, especially when including icons, it did not stay inline. This happens for every whitespace.

To make it still look consistent even if the label content is a bit larger, i also made the label centered to the top right corner of the parent element, 

## Testcase
http://jsfiddle.net/k1qp7dgv/3/
Remove CSS to see the issue

#### Before
image

#### After
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4217
https://github.com/Semantic-Org/Semantic-UI/issues/4155

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -894,8 +894,10 @@ each(@colors,{
   position: absolute;
   z-index: @floatingZIndex;
   top: @floatingTopOffset;
-  left: 100%;
+  right: 0;
   margin: 0 0 0 @floatingLeftOffset !important;
+  white-space: nowrap;
+  transform: translateX(50%);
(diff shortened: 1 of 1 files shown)


lubber-de:
I also prepared more enhancements by introducing:
```css
.ui.left.floated.label {}
.ui.left.aligned.floated.label {}
.ui.right.aligned.floated.label {}
```

image

http://jsfiddle.net/04vgcfdt/

Should i commit this here or submit a new PR (which will depend on merging this one first)

hammy2899:
@lubber-de separate PR.

PR #443, opened 2019-01-29
[Reset] Updated normalize.css to 8.0.1

lubber-de:
## Description
The docs say Fomantic-UI includes the latest normalize.css, so we should update to the latest 8.0.1.
FUI currently includes 7.0 and we are not supporting IE <11 anymore.


hammy2899:
I will watch the normalize.css repo so we can stay on top of updates.

PR #464, opened 2019-02-07
[Shape] Fix 3D flip starting zoom, fix nested shapes, ignore child elements with class 'active'

lubber-de:
## Description
This PR fixes the following issues:

- Whenever a shape had child elements which also had the `active` class set (either from another element like `menu` or a nested shape but not the "own" `active side`), the whole animation was messed up (see screenshots/testcase)

- Therefore nested shape elements were not working

- The Flip animations always started with an hard increased size to gain the 3d-Cube effect. But it still looked misleading instead of a smooth transition. Reason for this was, how shape dealt with translateZ. The fix also reduced the amount of needed code :slightly_smiling_face: 

### Fixed
http://jsfiddle.net/qtL0dkr3/

### Broken
http://jsfiddle.net/mpkjave0/1/

### Nested shapes
|Before|After|
|-|-|
|shape_nested_bad|shape_good|

### Flip Animation
|Before|After|
|-|-|
|shape_3d_bad|shape_3d_good|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4704
https://github.com/Semantic-Org/Semantic-UI/issues/4699
https://github.com/Semantic-Org/Semantic-UI/issues/5127


lubber-de:
:thinking: Not completely satisfied with the transition between different sizes.
I am going to fix this also when the next shapebox differs in size with the current shapebox. Looks too ugly (even if probably rarely used)
Don't merge yet. I am on my way :slightly_smiling_face:

lubber-de:
Fixed the transition if shape sides differ in size.
**Now** it looks as smooth as it should have been before :smile: I also updated the jsfiddle to http://jsfiddle.net/qtL0dkr3/
@hammy2899 @prudho @ColinFrick Please review again.

|Before|After|
|-|-|
|shape_nested_good|shape_good|

PR #468, opened 2019-02-08
[Modal] Don't move content and modal on dimmer when scrollbar hides

### LINKED ISSUE #427: Close modal with a scrollbar page generate an unsightly effect
# Bug Report

## Steps to reproduce
When you close a modal and the page have a scrollbar, the scrollbar is re-display before the modal is completely hidden then the modal is move to the left and that's generate an unsightly effect (it's almost invisible)

## Expected result
The scrollbar of page need to be display only after completly hidden the modal

## Screencast
I make a slowmotion video to see the problem :
https://youtu.be/3MWSA2P-bYI

## Version
2.7.1

lubber-de:
## Description
When a modal is shown, a possible scrollbar gets hidden because of `overflow:hidden` to the body. This is correct to prevent scrolling of the background.
However, the overflow actually removes the scrollbar providing additional space (width of the scrollbar) to the body content. This again results in a recognizable width change of all background content and will be visible again when the modal closes.
Also, because of this behavior, the modal gets immediatly moved to the left by the width of the scrollbar when it gets hidden, because the dimmer resets the overflow setting **before** the dimmer is actually completely hidden.

This PR solves everything now:

- Still remove the scrollbar by overflow hidden (no change)
- **but** give a temporary right margin to the body having the same width as the scrollbar of the background
- This way all the content stays where it was before, while and after a modal is shown/hidden resulting in a smooth modal transition and non-disturbing background

### Broken / current behavior
https://jsfiddle.net/smudoy1k/4/

### Fixed / smooth behavior
https://jsfiddle.net/smudoy1k/5/

### Broken / current behavior
modal_scrollbar_move_bad

### Fixed / smooth behavior
modal_scrollbar_move_good

## Closes
#427 
https://github.com/Semantic-Org/Semantic-UI/issues/4025
https://github.com/Semantic-Org/Semantic-UI/issues/2061
https://github.com/Semantic-Org/Semantic-UI/issues/4207
https://github.com/Semantic-Org/Semantic-UI/issues/3760

### DIFF
--- a/src/definitions/modules/dimmer.js
+++ b/src/definitions/modules/dimmer.js
@@ -306,10 +306,8 @@ $.fn.dimmer = function(parameters) {
                   queue       : false,
                   duration    : module.get.duration(),
                   useFailSafe : true,
-                  onStart     : function() {
-                    module.remove.dimmed();
-                  },
                   onComplete  : function() {
+                    module.remove.dimmed();
                     module.remove.variation();
                     module.remove.active();
                     callback();
@@ -319,10 +317,10 @@ $.fn.dimmer = function(parameters) {
             }
             else {
               module.verbose('Hiding dimmer with javascript');
-              module.remove.dimmed();
               $dimmer
                 .stop()
                 .fadeOut(module.get.duration(), function() {
+                  module.remove.dimmed();
                   module.remove.active();
                   $dimmer.removeAttr('style');
                   callback();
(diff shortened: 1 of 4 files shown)


dutrieux:
I'm wondering if It's normal when the container sizes is little, the fade in is not properly display then the dialog is display ? :

dialog

and the dinner is not properly fade out when the dialog is closing !

I have the problem only on Firefox (v65), not on Chrome !

lubber-de:
@dutrieux Interesting, but this is definately some firefox issue not displaying the transition correctly in that case (it does it correctly on larger display sizes as you said), because it also happens with the old behavior/current code (https://jsfiddle.net/smudoy1k/4/), so it's not releated to the change in this PR.
Edge and even IE11 (!) are doing it correctly as expected.

PR #474, opened 2019-02-11
[Transition] Transitions made of hidden inline style elements were not animated correctly anymore

lubber-de:
## Description
This PR fixes a regression since #357. All transitions made out of elements being invisible by an inline style `display:none` were not working correctly anymore. 

## Testcase
The subheader of the title should animate by the `drop` transition
### Broken
http://jsfiddle.net/q2u0hjy1

### Fixed
http://jsfiddle.net/bc5hgv9j/1/


lubber-de:
This also fixes #581

PR #488, opened 2019-02-13
[Segment,Header] Color sub header in inverted segment just like normal header

### LINKED ISSUE #486: Sub headers in inverted segments are unreadable (black on black)
# Bug Report

## Steps to reproduce

See attached jsfiddle.

## Expected result

The sub header is readable (probably white, the same as the main header text).

## Actual result

The sub header is unreadable as it's the same colour as the background.

## Testcase

https://jsfiddle.net/thm9zvx2/

## Screenshot (when possible)

fomantic ui sandbox jsfiddle

This is slightly out of date, please see the JSFiddle.

lubber-de:
## Description
`header` is directly supported to change its color to white in `inverted segment` without the need to be specified as `inverted header`.
Unfortunately this is not the case for `sub header`, which this PR now fixes

## Testcase
https://jsfiddle.net/njvdhk2e/

## Closes
#486

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -61,6 +61,7 @@
 --------------------*/
 
 /* Header */
+.ui.inverted.segment .sub.header,
 .ui.inverted.segment > .ui.header {
   color: @white;
 }
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -61,7 +61,7 @@
 --------------------*/
 
 /* Header */
-.ui.inverted.segment .sub.header,
+.ui.inverted.segment > .ui.header > .sub.header,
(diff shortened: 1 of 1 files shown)


src/definitions/elements/segment.less
```diff
@@ -61,6 +61,7 @@
 --------------------*/
 
 /* Header */
+.ui.inverted.segment .sub.header,
```
hammy2899: I think this should be 
```css
.ui.inverted.segment > .ui.header > .sub.header
```

lubber-de: yes, it's better. fixed it accordingly

lubber-de: Also changed fiddle in the description ;)

PR #498, opened 2019-02-18
[API] Set context in beforesend to $module instead of possible stateContext to match docs

lubber-de:
## Description
According to [the docs](https://fomantic-ui.com/behaviors/api.html#/settings), `beforeSend` should have the "initialized element" as context, not the element matching `stateContext`!

### Current/Wrong
https://jsfiddle.net/7oz0bncp/
### Fixed
https://jsfiddle.net/7oz0bncp/1/

### Current/Wrong
beforesend_wrong

### Fixed
beforesend_right

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6207

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -721,7 +721,7 @@ $.api = $.fn.api = function(parameters) {
             var
               runSettings
             ;
-            runSettings = settings.beforeSend.call(context, settings);
+            runSettings = settings.beforeSend.call($module, settings);
             if(runSettings) {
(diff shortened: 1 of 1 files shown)


ColinFrick:
Should this be considered a breaking change? I mean if anyone used beforeSend and `this` in the callback, it wouldn't work anymore (even if `this` is wrong)

hammy2899:
@ColinFrick yeah that's true, lets put this on hold until we come to a better decision on what to do.

Issue #386, opened 2019-01-11
Fomantic and yarn - install issues


etshy:
# Bug Report

## Steps to reproduce
1. having an existing package.json with fomantic ui as dep
2. lauch `yarn install`
3. Error while building fomantic ui

## Expected result
The installation have no problems

## Actual result
error while building fomantic ui

## Screenshot (when possible)
![console screenshot](https://i.imgur.com/Z5eb0Ve.png)

## Version
fomantic : 2.7.1
yarn : 1.13.0

is fomantic not "compatible" with yarn ?

ps : I'm trying yarn because it's recommended to build Symfony assets.

lubber-de:
Less 1.3.3 is a bit old... Fomantic 2.7 needs less >=3.7

etshy:
Any idea why yarn try to install this 1.3.3 version ?

My package.json doesn't specify any less version

```
{
  "name": "public",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "blazy": "^1.8.2",
    "dropzone": "^5.5.1",
    "fomantic-ui": "^2.7.1",
    "historyjs": "^1.8.0-b2",
    "jquery-ui": "^1.12.1"
  },
  "devDependencies": {
    "@symfony/webpack-encore": "^0.17.1",
    "node-sass": "^4.10.0",
    "sass-loader": "^6.0.7",
    "webpack": "^3.12.0",
    "webpack-notifier": "^1.7.0"
  }
}
```

lubber-de:
Seems like yarn is not installing dependencies of dependencies. Maybe this thread  https://github.com/yarnpkg/yarn/issues/2763 and the proposed solutions (like editing the lock file) helps.
I am a bit confused because thee is a merged PR in yarn related to the mentioned issue, so maybe you just have to update yarn (tbh I haven't tested this myself) 
It's up to yarn to provide a working solution for that.

mathieutu:
Hey, I've a problem too with yarn:
``` 
...
[12:55:23] Starting 'run setup'...
? Set-up Semantic UI (Use arrow keys)
❯ Automatic (Use default locations and all components) 
  Express (Set components and output folder) 
  Custom (Customize all src/dist values) [12:55:24] The following tasks did not complete: install, run setup
[12:55:24] Did you forget to signal async completion?
[...]

PR #501, opened 2019-02-20
[Build] Refactor dependencies

hammy2899:
## Description

Refactor dependencies so they are in the correct place and so they all have specific versions instead of `*` and remove the unless peerDeps. Also update a few dependencies.

## Related
#386

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,3 @@ cache: npm
 
 before_install:
   - sh preinstall.sh
-
-env:
-  global:
-    - NODE_ENV=production
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,40 +5,39 @@
   "requires": true,
   "dependencies": {
     "@octokit/endpoint": {
-      "version": "3.1.1",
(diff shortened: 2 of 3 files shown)


lubber-de:
I think `package-lock.json` has to be regenerated also. Travis [does not build](https://travis-ci.org/fomantic/Fomantic-UI/builds/495876844?utm_source=github_status&utm_medium=notification) (i already tried to restart the build, but still not building) and i remember, we had a similar situation before while dealing with the Event stream issue #269

lubber-de:
@hammy2899 
🤔 I see now, `package-lock.json` is already part of your PR... any idea, why travis does not work then?

lubber-de:
Perhaps, because you added `dev:true`  to the dependencies, but travis.yml has set "production"?
https://github.com/fomantic/Fomantic-UI/blob/783786b5c27bf0252dc47c0277746b4a695210ca/.travis.yml#L18

hammy2899:
@lubber-de Yeah, I will change the env and see if that fixes it

hammy2899:
@lubber-de Fixed 👍

PR #529, opened 2019-02-28
[Shape] Focussed Checkbox inside shape breaks animation

lubber-de:
## Description
When a shape had a focussed input inside, any animation was broken.
Previously, you had to click outside of the shape to blur the input to make the animation work again.

Although in the example below, the checkbox indeed gets blurred right before the button to animate the shape is clicked, it seems the browser engine still gets confused and does not recognize the change early enough.
This PR now delays the animation for a short time in case of input being part of the the shape, so the blurred checkout is also recognized by the browser engine at animation time

## Testcase
Check/uncheck the checkbox. Then **quickly** (means don't hold the mousebutton for too long) click the Button to trigger the animation

### Broken
https://jsfiddle.net/o5tqeuha/

### Fixed
https://jsfiddle.net/xr2jvnof/

## Screenshot 
|Broken|Fixed|
|-|-|
|checkbox_shape_broken|checkbox_shape_fixed|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6583

### DIFF
--- a/src/definitions/modules/shape.js
+++ b/src/definitions/modules/shape.js
@@ -57,6 +57,7 @@ $.fn.shape = function(parameters) {
         selector      = settings.selector,
         error         = settings.error,
         className     = settings.className,
+        regExp        = settings.regExp,
 
         // define namespaces for modules
         eventNamespace  = '.' + namespace,
@@ -197,6 +198,20 @@ $.fn.shape = function(parameters) {
           },
           hidden: function() {
             return $module.closest(':hidden').length > 0;
+          },
+          mobile: function() {
+            var
+                userAgent    = navigator.userAgent,
+                isMobile     = userAgent.match(regExp.mobile)
+            ;
+            if(isMobile) {
+              module.verbose('Browser was found to be mobile', userAgent);
+              return true;
+            }
+            else {
+              module.verbose('Browser is not mobile, using regular transition', userAgent);
+              return false;
+            }
           }
         },
 
@@ -764,7 +779,7 @@ $.fn.shape = function(parameters) {
           $inputs.blur();
           setTimeout(function(){
             module.invoke(query);
-          }, 100);
+          }, module.is.mobile() ? 150:100);
         } else {
           module.invoke(query);
         }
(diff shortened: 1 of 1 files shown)


src/definitions/modules/shape.js
```diff
@@ -759,7 +759,15 @@ $.fn.shape = function(parameters) {
         if(instance === undefined) {
           module.initialize();
         }
-        module.invoke(query);
+        var $inputs = $module.find('input');
+        if( $inputs.length > 0) {
+          $inputs.blur();
+          setTimeout(function(){
```
hammy2899: Why is this timeout needed?

lubber-de: Without the timeout the browser does still have the information that any input is still focused and still breaks the animation, even a value  below 100ms did not work. Try it out without the timeout and you will recognize it won't work

hammy2899: Ok that's fine, was going to suggest changing to `0` but if you have tested that its fine 👍

lubber-de:
I had to adjust the delay for mobile, see latest commit and here
https://jsfiddle.net/cvw7p4zL/2/

lubber-de:
What do you guys think? Should I remove the mobile recognition again and just stick with 150ms all the time, also on desktop then?

src/definitions/modules/shape.js
```diff
@@ -197,6 +198,20 @@ $.fn.shape = function(parameters) {
           },
           hidden: function() {
             return $module.closest(':hidden').length > 0;
+          },
+          mobile: function() {
+            var
+                userAgent    = navigator.userAgent,
```
hammy2899: You may want to add a check to see if `navigator` exists, some browsers may not have support https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator#Browser_compatibility

lubber-de: I copied the code from sidebar.js. Would need to change this aswell the too. But MA, be I skip the whole check and just use 150ms all the time?

hammy2899: Yeah could do, that would be the easiest option.

lubber-de: Did it . See latest commit

lubber-de:
Removed mobile check and just increased to 150ms
New fiddle: https://jsfiddle.net/xr2jvnof/

PR #540, opened 2019-03-06
[Formvalidation] Support checkbox groups for "checked" rule like radio groups

lubber-de:
## Description
Checkbox groups where the checkboxes share the same name property were only "checked" validated if every single checkbox within the group was checked. This is wrong, it should behave the same as the radio group "checked" validation (SUI does it)

> Hint: The example uses the same name attribute for each checkbox also with a bracket suffix "[]"
The form validator already supports using this selector by putting it into quotes .
> Hint 2: selector.checkbox covers checkbox and radiobutton aswell..in case you wonder if the old behavior with radio still works 

```javascript
fields: {
    "sizes[]": 'checked'
  },
```

### Broken
Every toggle checkbox has to be checked to pass the validation
https://jsfiddle.net/ogz73ks5/

### Fixed
Only one checkbox has to be checked to pass validation
https://jsfiddle.net/ogz73ks5/2/

## Screenshot
|Broken|Fixed|
|-|-|
|checkboxgroup_bad|checkboxgroup_good|

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1040,9 +1040,9 @@ $.fn.form = function(parameters) {
               ruleName     = module.get.ruleName(rule),
               ruleFunction = settings.rules[ruleName],
               invalidFields = [],
-              isRadio = $field.is(selector.radio),
+              isCheckbox = $field.is(selector.checkbox),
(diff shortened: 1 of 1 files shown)


crewow:
That's great , I Love FUI

Issue #456, opened 2019-02-03
Use Typescript


28Smiles:
# Feature Request

## Description
Since with v3 the Formantic-UI is going under heavy development and a lot of the JS is rewritten, isn't it a good point of time to write the modules directly in Typescript. This would ensure a slim Typescript integration (#396 ) and probably would also eliminate some Bugs.

hammy2899:
@28Smiles TS is something I want to look into but it could hold back the development since all the members of the core team have no experience with TS (since the last time we discussed it) which would mean we would all need to learn it. That being said I do understand the benefits so this will definitely be a topic of discussing again when it comes to the time to make the decision.

lostfields:
You can use JavaScript in Typescript, but not the other way around - so it shouldn't be any issue at all if you are not experienced with Typescript. You could make it allow JavaScript as well, and implicit accept type "any" to make the transformation easy.

hammy2899:
@lostfields I have been doing a lot of typescript recently and feel fairly confident with it now and I know @ColinFrick uses it so I don't think that will be the problem going forward anymore. The decision on whether we use TS or not will still depend on whether we are writing the JS our self or using stardust (as discussed [here](https://github.com/fomantic/Fomantic-UI/issues/319))

YamiOdymel:
I've tried to write the modules with TypeScript few weeks ago and I had lots of the issue with DOM.

TypeScript suits for the data-only things (Node.js backend for example), but with DOM elements? Not really.

Since there are bunch of the different HTML element types (for example: `HTMLTableElement`, 
`HTMLTableSectionElement`, `HTMLTableCaptionElement`, `HTMLTableCellElement`), it's impossible to done the module with TypeScript because it's hard to specify the types of the elements.

If you are smart enough, you might know that you could use `HTMLElement` interface as a generic type instead of specifying their _real types_, but and then you lose the methods of each different type of the elements because they are now a `HTMLElement`.

Also, TypeScript [lacks some types](https://stackoverflow.com/questions/35789498/new-typescript-1-8-4-build-error-build-property-result-does-not-exist-on-t) in its definitions if you are interacting with the file input.

![50553587_1985212471775137_8991193277428400128_o](https://user-images.githubusercontent.com/7308718/53188628-c3a9c200-3640-11e9-87fa-7a32c3a30b49.jpg)

bjmiller:
[...]

PR #555, opened 2019-03-10
[Platform] Setup JavaScript linter and apply coding standards

exoego:
## Description

- [x] Setup eslint with the [relatively popular](https://standardjs.com/#who-uses-javascript-standard-style) rule named [estlint-config-standard](https://github.com/standard/eslint-config-standard)
- [x] Override some rules to suit the current coding styles, so diff are minimized. 
- [x] Disable some rules (e.g. `no-self-assign`) that may change behavior. (should be carefully enabled in subsequent, separate PRs)
- [x] Apply coding styles
- [x] Enable linting on CI (Travis) to ensure that every PR follows coding styles.
- [x] Add the coding standard section to `CONTRIBUTING.md` 

## Closes

Partially adresses "Set up linter (eslint) ⚡️" in #29


exoego:
I thought linting like https://github.com/fomantic/Fomantic-UI/pull/538 should be automated: instant check and fix by `eslint --fix`, and enforced by CI (Travis).
So contributors do not confuse coding standard and focus on logic.

Though this PR is WIP, I opend up it to gather feedbacks from core maintainers.

src/definitions/behaviors/form.js
```diff
@@ -789,8 +787,8 @@ $.fn.form = function(parameters) {
           },
           // alias
           rules: function(field, rules) {
-            if(Array.isArray(field)) {
-              $.each(fields, function(index, field) {
+            if (Array.isArray(field)) {
+              $.each(field, function(index, field) {
```
exoego: This issue was found by [`no-undef` rule](https://eslint.org/docs/rules/no-undef)

There is a separate PR #556 to fix this.

.eslintrc.js
```diff
@@ -0,0 +1,53 @@
...
+  'extends': 'standard',
+  'parserOptions': {
+    'ecmaVersion': 5
+  },
+  'rules': {
+    // -----------------------------------------------------------------------------------------------------------------
+    // Overrides 'standard' rules to suit the present format
+    // -----------------------------------------------------------------------------------------------------------------
+    'brace-style': ['error', 'stroustrup', { 'allowSingleLine': true }],
+    'indent': ['error', 2, { 'outerIIFEBody': 0 }],
```
lubber-de: I think the indent should be set to 4 characters instead of 2, just because var/let groups look ugly otherwise:
```javascript
var a,
    b,
    c,
    d;
```

would become 

```javascript
var a,
  b,
  c,
  d;
```

as seen in https://github.com/fomantic/Fomantic-UI/pull/555/commits/aea64d9b546961366b71ab8f60c8599d566f5da4#diff-77caea316b6e2c075a2c5d36f572b036  (weekOfYear function)

src/definitions/behaviors/api.js
```diff
@@ -633,14 +633,14 @@ $.api = $.fn.api = function(parameters) {
                 response = responder;
               }
               // simulating response
-              mockedXHR.resolveWith(context, [ response, textStatus, { responseText: response }]);
+              mockedXHR.resolveWith(context, [response, textStatus, { responseText: response }]);
```
lubber-de: Which rule in eslintrc.js is causing this?
Although i like removing bracket-spacing it "conflicts" with the "object-curly-spacing" rule, where a space is inserted instead of being removed. This seem inconsistent to me.
I suggest using `"space-in-brackets": ["error", "never"]` instead to remove trailing spaces in object and array lists  equally

.eslintrc.js
exoego: Instead of debating `2`-VS-`4` just for `var` alignment, I suggest we should 

1. Keep 2 chars indentation as-is 
2. Rewrite single var declarations into separate `var` declrations by enabling `'one-var': 'never'`, in subsequent PR

I propose these because I believe `one-var: never` is better choice for maintainability from other JavaScript/TypeScript OSS experiences..


# Reason 1: Clearer diff when adding/removing variables

## one-var 👎 

```diff
 var a,
-    b = true,
-    c = b.bar();
+    c = b.bar(),
+    d;
```

## no one-var 👍 

```diff
  var a;
- var b = true;
  var c = b.bar();
+ var d;
```

# Reason 2: Future-proof (`let` and `const`)

We will eventually use `let` and `const` when we adopt ES6 #319 or TypeScript #456 or similar.

Though [ESLint is very configurable](https://eslint.org/docs/rules/indent#enforce-consistent-indentation-indent), but I don't think it is good idea to chnge indentations like this (`4`for `var`/`let`, `6` chars for `const`)

```js
let a,
    b = true;
const c = $('.foo')
      d = c.bar();
d.buz();
```

or this (`6` for all)

```js
let   a,
      b = true;
const c = $('.foo')
      d = c.bar();
d.buz();
```

The below `one-var: nver` style makes each variable declarations more obvious.

```js
let a;
let b = true; 
const c = $('.foo');
const d = c.bar();
```

src/definitions/behaviors/api.js
exoego: The below rules are related.
* standard/array-bracket-even-spacing
* standard/computed-property-even-spacing
* standard/object-curly-even-spacing

According these rule, spacing around array/computed-property/object should be even.

👍 No spacing

```js
mockedXHR.resolveWith(context, [response, textStatus, { responseText: response }]); 
```

👍 Spacing 
```js
mockedXHR.resolveWith(context, [ response, textStatus, { responseText: response } ]); 
```

👎 Inconsist spacing

```js
mockedXHR.resolveWith(context, [ response, textStatus, { responseText: response }]); 
```

.eslintrc.js
lubber-de: We don't need to debate on 2/4, it was just my personal feeling, so i would still be fine with it, if everybody else also agrees.

But I disagree with using `no one-var`. Having lots of vars is bad style and IMHO not very much maintainable or readable and would also mean every class var definition would need to be changed aswell, so
```javascript
var a,
    b= function(){
    },
    c= function()= {
    },
    d= function() = {
    };
```
would become
```javascript
var a;
var b= function(){
  };
var c= function()= {
  };
var d= function() = {
 };
```
Even removing / commenting vars does not make any difference

So i dont think we should use `one-var: never`

src/definitions/behaviors/api.js
lubber-de: ok then :)

.eslintrc.js
lubber-de: Anyhow, as you said, a possible change to the handling of "vars/lets/const" would be part of an later PR, i approve this now 🙂

exoego:
@prudho 
Added to .gitignore in https://github.com/fomantic/Fomantic-UI/pull/555/commits/4ea0009334487c1ecbca025c07222269fe89afd8

ColinFrick:
I personally don't agree with the `space-before-function-paren` setting for anonymous functions. In my mind, why should `if` and anonymous `function` be any different:
`if () {}`
`function() {}`
looks inconsistent to me.

exoego:
The proposed configuration in PR is determined based on existing code base.
We can gradually change in subsequent PRs.

ColinFrick:
Ok, I see that now.
IMHO if we introduce eslint now, we should discuss the configuration and not be beholden to the current code base.

@fomantic/core what do you think?

exoego:
@ColinFrick 
Updated `.eslintrc.js` and now `space-before-function-paren` is `always`.

hammy2899:
I am on the side of @ColinFrick, the current code base is a mess when you introduce linters and we have already tried adding eslint [here](https://github.com/fomantic/Fomantic-UI/pull/64). As @ColinFrick said we need to discuss the config we want to move the code base to and when we do that it will be a huge task. IMHO moving the code base to a specific lint config is just a waste of time right now especially with v3 somewhat around the corner.

exoego:
> just a waste of time

It is a very sorry to hear like that.

From my experience, I believe linting make reviewing efficient and ensure that there are no human errors of some kinds (unused var, typo, duplicated properties...).
 

Why not introduce lint for v2 to review cost for upcoming PRs and rewriting to v3?
Discussing configuration may require time but then it should be postponed.

prudho:
I think that setting up eslint now is harmful for the code base and can be done now. On the future we still can provide some changes of the configuration, but enabling linting should not be tied to a v3 release.

exoego:
Updated the PR to resolve conflicts with latest PRs.

exoego:
Though I believe linting reduce cost of reviewing PRs and find possible bugs,
closing this since it does not get reviewed for a long time.

PR #560, opened 2019-03-11
Enhancements: [Dropdown] expose week cell class setting.

patilkiranm:
## Description
Possibility to set different week cell class

This PR exposes weekCell class setting within className setting.
- The week cell class setting defaults to disabled
- The week header cell must be identical to other header cell

## Screenshot (when possible)

image

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -298,6 +298,7 @@ $.fn.calendar = function(parameters) {
                   if(settings.showWeekNumbers) {
                       cell = $('<th/>').appendTo(row);
                       cell.text(settings.text.weekNo);
+                      cell.addClass(className.weekCell);
                       textColumns--;
                   }
                   for (i = 0; i < textColumns; i++) {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/calendar.js
```diff
@@ -298,7 +298,6 @@ $.fn.calendar = function(parameters) {
                   if(settings.showWeekNumbers) {
                       cell = $('<th/>').appendTo(row);
                       cell.text(settings.text.weekNo);
-                      cell.addClass(className.disabledCell);
```
lubber-de: Entirely removing the disabled class for the header will result in always displaying the Week-Number Text in the same prominent style as the weekdays which might be considered a breaking change (as the user cannot change it anymoire now)
I suggest to either have a separate option for the headerCell Class or use the same className.weekCell option for this to avoid being a breaking change and more ensure consistency

|Before|After|
|-|-|
|image|image|

patilkiranm: Updated PR: The week cell class is applied to header and normal row cell.

image

PR #562, opened 2019-03-11
Enhancements: [Dropdown] Possibility to select adjacent month dates.

patilkiranm:
## Description
The calendar can show dates from adjacent months. These adjacent month dates can also be made selectable.

Option: selectOtherMonths boolean defaults to false

## Screenshot

selectothermonths

### DIFF
--- a/src/definitions/modules/calendar.less
+++ b/src/definitions/modules/calendar.less
@@ -127,6 +127,10 @@
   color: @disabledTextColor;
 }
 
+.ui.calendar .ui.table tr .adjacent:not(.disabled) {
+  color: @disabledTextColor;
+}
+
 /*--------------
      States
 ---------------*/
--- a/src/definitions/modules/calendar.less
+++ b/src/definitions/modules/calendar.less
@@ -128,7 +128,8 @@
 }
 
 .ui.calendar .ui.table tr .adjacent:not(.disabled) {
-  color: @disabledTextColor;
+  color: @adjacentTextColor;
+  background: @adjacentBackground;
 }
 
 /*--------------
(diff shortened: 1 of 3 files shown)


src/definitions/modules/calendar.less
```diff
@@ -127,6 +127,10 @@
   color: @disabledTextColor;
 }
 
+.ui.calendar .ui.table tr .adjacent:not(.disabled) {
+  color: @disabledTextColor;
```
lubber-de: When an adjacent day should be selectable, there must be a clear visual indication about that. So displaying selectable adjacent days in the same color as disabled ones isn't clear to the user. I also think such days should be visualised to indicate they belong to one of the adjacent months.
Therefore i suggest the following change:

calendar.variables:
```less
@adjacentTextColor: @mutedTextColor;
@adjacentBackground: @subtleTransparentBlack;
```

calendar.less:
```less
.ui.calendar .ui.table tr .adjacent:not(.disabled) {
    color: @adjacentTextColor;
    background: @adjacentBackground;
}
```

That will look as follows and even if there are still disabled days in between  (Feb 25th) this is much more understandable to the user, what is selectable and what's not:

image

patilkiranm: Updated PR: Now selectable adjacent days are visually distinguished.

patilkiranm:
> I think renaming the option to `selectAdjacentDays` fits better

Updated PR: option name changed from `selectOtherMonths` to `selectAdjacentDays`

lubber-de:
As it is so far approved by @hammy2899 and me i added this to 2.7.3
@prudho @ColinFrick Fee free to change to 2.7.4 if you disagree

hammy2899:
@all-contributors please add @patilkiranm for code

PR #574, opened 2019-03-20
added check for date parse failure

### LINKED ISSUE #573: [Calendar] throw specific error for unparseable date format
# Bug Report

## Steps to reproduce
1.  $('#myCal').calendar()
2.  $('#myCal').calendar('set date', 'asadsadsa', true, false)

## Expected result
Error message saying that date format is invalid. 
## Actual result
Cannot read property 'getTime' of null/undefined

## Testcase
https://jsfiddle.net/m2cod85x/6/

## Version
2.7.2

tjaart:
## Description
Added check for null and undefined in sanitiseDate() method. I also added logging for invalid date formats being passed into 'set date'. The call to getTime() was crashing if the date parse failed.

## Testcase
https://jsfiddle.net/m2cod85x/6/

## Closes
#573

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -657,7 +657,7 @@ $.fn.calendar = function(parameters) {
             date = module.helper.sanitiseDate(date);
             
             if (!date) {
-                console.error('Invalid date value for calendar')
+                module.error(error.dateValue);
             }
             
             date = module.helper.dateInRange(date);
@@ -1483,7 +1483,8 @@ $.fn.calendar.settings = {
 
   error: {
     popup: 'UI Popup, a required component is not included in this page',
-    method: 'The method you called is not defined.'
+    method: 'The method you called is not defined.',
+    dateValue: 'Invalid date value for calendar'
   },
 
   className: {
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -655,6 +655,11 @@ $.fn.calendar = function(parameters) {
             updateInput = updateInput !== false;
             fireChange = fireChange !== false;
             date = module.helper.sanitiseDate(date);
+            
+            if (!date) {
+                console.error('Invalid date value for calendar')
+            }
+            
             date = module.helper.dateInRange(date);
 
             var mode = module.get.mode();
@@ -861,6 +866,9 @@ $.fn.calendar = function(parameters) {
             }
             if (!(date instanceof Date)) {
               date = parser.date('' + date, settings);
+              if (!date || date === null) {
+                  return undefined;
+              }
             }
             if (isNaN(date.getTime())) {
               return undefined;


src/definitions/modules/calendar.js
```diff
@@ -655,6 +655,11 @@ $.fn.calendar = function(parameters) {
             updateInput = updateInput !== false;
             fireChange = fireChange !== false;
             date = module.helper.sanitiseDate(date);
+            
+            if (!date) {
+                console.error('Invalid date value for calendar')
```
lubber-de: Please use the internal method `module.error` instead and terminate with semicolon
```javascript
module.error('Invalid date value for calendar');
```

lubber-de: You may also move the error text into the error text object (where already `error.popup` and `error.method` exist)

```javascript
module.error(error.dateValue);

...

error: {
    popup: 'UI Popup, a required component is not included in this page',
    method: 'The method you called is not defined.',
    dateValue: 'Invalid date value for calendar'
}
```

tjaart: I've pushed changes, I'm not sure what to do next. The issues should be resolved in the latest commit.

lubber-de:
@tjaart Maybe the easiest solution to fix this might be to simply use `module.debug` instead of `module.error`, but then you probably have to remove the text again from the errror-object for consistence.
Of course the message will only appear in the console if the module has been initialized with `debug:true` then

src/definitions/modules/calendar.js
```diff
@@ -1483,7 +1483,8 @@ $.fn.calendar.settings = {
 
   error: {
     popup: 'UI Popup, a required component is not included in this page',
-    method: 'The method you called is not defined.'
+    method: 'The method you called is not defined.',
+    dateValue: 'Invalid date value for calendar'
```
hammy2899: I think this should be `invalid: 'Invalid date value'` so it can potentially be used for more then one error.

tjaart:
> @tjaart Maybe the easiest solution to fix this might be to simply use `module.debug` instead of `module.error`, but then you probably have to remove the text again from the errror-object for consistence.

I think there are two problems here. 

The first one was that the date parse didn't return `undefined` and instead carried on and crashed when an invalid date was passed in. This problem can be fixed cleanly and easily, by the added `return undefined` from the `sanitiseDate()` function. 

The second problem is that when explicitly calling `'set date'` and passing in something invalid, you can't see that there was a problem with the manual date. This seems a little more complicated, because you would have to know when a set date call happened internally, versus being called from `'set date'`. Maybe it will be better to leave this for the time being and get the `sanitiseDate()` fix in?

I'd like to hear opinions on this @lubber-de @hammy2899 ?

lubber-de:
@tjaart 
> Maybe it will be better to leave this for the time being and get the `sanitiseDate()` fix in?

Agreed

lubber-de:
@tjaart Do you think you are going to finish the PR?

lubber-de:
Closing because the author did not respond for 3 weeks and the main issue is now fixed and merged by #677

PR #580, opened 2019-03-23
Introduce stylelint for LESS and apply coding standards

exoego:
## Description

- [x] Setup stylelint with the [popular `estlint-config-standard` rule](https://github.com/stylelint/stylelint-config-standard)
- [x] Remove `csscomb` and `csslint` that are superceded by `stylelint`
- [x] Override some rules to suit the current coding styles, so diff are minimized. 
- [x] Disable some rules (e.g. `declaration-block-no-duplicate-properties`) that may change behavior. (should be carefully enabled to fix potential bugs in subsequent, separate PRs)
- [x] Apply coding styles
- [x] Enable linting on CI (Travis) to ensure that every PR follows coding styles.
- [x] Add the coding standard section to `CONTRIBUTING.md` 

## Closes

Partially adresses "Set up linter (eslint) ⚡️" in #29 

Partially adresses #274.


src/definitions/elements/list.less
```diff
@@ -607,18 +594,18 @@ ul.ui.horizontal.bulleted.list li:first-child,
 .ui.horizontal.bulleted.list > .item:first-child {
   margin-left: 0;
 }
-ul.ui.horizontal.bulleted.list li::before,
-.ui.horizontal.bulleted.list > .item::before {
+ul.ui.horizontal.bulleted.list li:before,
```
lubber-de: I am not complaining for a change, because the CSS2 versions of pseudo-elements will still work in every browser (who knows for how long), but instead of changing it here, we should discuss about a general ruleset.

I understand you have chosen this change/rule because you compared the whole codebase for less changes and i have to admit the CSS2/CSS3 :/:: variants are not consistent in the overall FUI codebase, but especially for this rule, my personal opinion is to use/keep/focce the CSS3 version using double colons instead, just because we do not support IE6/IE7/8 anymore which are the only browsers which don't understand the CSS3 format.

exoego:
> 
> 
> The Travis build is broken by your config. The 'stylelint'- run breaks because of a missing fixed rule in calendar. See https://travis-ci.org/fomantic/Fomantic-UI/jobs/510194664
> 
> ```
> src/definitions/modules/calendar.less
>  115:1  ✖  Unexpected empty line before rule   rule-empty-line-before
> ```

This is fixed in https://github.com/fomantic/Fomantic-UI/pull/580/commits/4ed5efef5ff129af4deaf689302a3b963d95e369


> 
> Also, i am confused why you set 2 env settings for `JOBTYPE` in the matrix sections 🤔

This allows Travis to run multiple CI tasks concurrently (`test`, `stylelint`, `eslint` and alike), which means faster than sequential execution of such tasks.

src/definitions/elements/list.less
exoego: Addressed in https://github.com/fomantic/Fomantic-UI/pull/580/commits/7f117500b0cc7fd5d0be768fe16b37bf8f097d4d
I agree to choose `::` over `:` since CSS2-only browsers are no longer supported.

exoego:
Thanks for reviewing.
Updated the PR to resolve conflict with latest develop branch.

exoego:
Updated the PR to resolve conflicts with latest PRs.

exoego:
Though I believe linting reduce cost of reviewing PRs and find possible bugs,
closing this since it does not get reviewed for a long time.

PR #587, opened 2019-03-25
[Dropdown] Height variation

### LINKED ISSUE #545: [Dropdown] Taller dropdown so that less need to scroll
# Feature Request

## Description

I want to expand dropdown height as much as possible, so that end-user can glance many items at one time and a need of scrolling is reduced.

* css-class variation like `ui taller dropdown`
* JS optionthat extends its height

Currently, dropdown height varies between 3 items and 6 items depending on window width.

## Testcase
https://jsfiddle.net/Lhupngwa/1/

### Actual
image

### Expected 

image

exoego:
## Description

This PR adds height variation to dropdown module so that more items can be glanced at once.
This is helpful if dropdown have tons of items (e.g. Countries, States/Providences, Divisions, something like that).

This is usable if every items can be shown at once within the height of window without scroll bar.
If there are so many items that requires scroll bar, user can consider columnar variation proposed in #586.

## Screenshot (when possible)

 
image

## Closes
Closes #545

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -439,20 +439,44 @@ select.ui.dropdown {
   .ui.selection.dropdown .menu {
     max-height: @selectionMobileMaxMenuHeight;
   }
+  .ui.selection.dropdown.higher .menu {
+    max-height: @selectionMobileMaxMenuHeight * 2;
+  }
+  .ui.selection.dropdown.highest .menu {
+    max-height: @selectionMobileMaxMenuHeight * 3;
+  }
 }
 @media only screen and (min-width: @tabletBreakpoint) {
   .ui.selection.dropdown .menu {
     max-height: @selectionTabletMaxMenuHeight;
+  }
+  .ui.selection.dropdown.higher .menu {
+    max-height: @selectionTabletMaxMenuHeight * 2;
+  }
+  .ui.selection.dropdown.highest .menu {
+    max-height: @selectionTabletMaxMenuHeight * 3;
   }
 }
 @media only screen and (min-width: @computerBreakpoint) {
   .ui.selection.dropdown .menu {
     max-height: @selectionComputerMaxMenuHeight;
   }
+  .ui.selection.dropdown.higher .menu {
+    max-height: @selectionComputerMaxMenuHeight * 2;
+  }
+  .ui.selection.dropdown.highest .menu {
+    max-height: @selectionComputerMaxMenuHeight * 3;
+  }
 }
 @media only screen and (min-width: @widescreenMonitorBreakpoint) {
   .ui.selection.dropdown .menu {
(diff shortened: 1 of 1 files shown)


lubber-de:
Higher/ highest or taller/tallest... I am fine with both 😀 what does our native speaker @hammy2899 say?

hammy2899:
IMHO `taller` doesn't make much sense without more context. The word `tall` is an adjective so is used to describe something. In this case `tall dropdown` means that the dropdown is taller then the default height because its describing the height of the dropdown so the user assumes the height will be taller then what it is originally. When using `taller dropdown` it doesn't have any context of what it is taller than. A good example to this is you don't say "That person is taller" you would say "That person is tall" but with some context you could say "That person is taller than that building" which would make sense.

I think adding words for height will have just as much of a problem as the size modifiers. In FUI we have `mini`, `tiny`, `small`, `medium`, `large`, `big`, `huge`, `massive` (in order from smallest to largest) and even English speaking users have problems with deciding which is largest or smaller than the other. Take `big` and `large` for example, personally I would argue `large` should be larger then `big` but it isn't. This is even more of a problem when none English speaking users try to distinguish them.

I think we need to think more about the terminology we use before we add anything to the framework so we can prevent breaking changes in the future.

prudho:
And what about `long` and `very long`, which are already user by the Placeholder component ?

hammy2899:
@prudho That's a good point and that would also be reusing current terminology.

exoego:
Changed to `long` and `very long` in https://github.com/fomantic/Fomantic-UI/pull/587/commits/5921029d45803049133ba5c90365c4556b6f5057

hammy2899:
We should also add `short` and `very short` so it stays consistent.

exoego:
I don't think shorter dropdown make sense.

But if you want consitency is more important, what factors should be for `very short` and `short` ?
Is it ok for `0.5` for `very short` and `0.75` for `short`?

  | very short | short | standard | long | very log
-- | -- | -- | -- | -- | --
factor | 0.5 | 0.75 | 1 | 2 | 3
"@selectionMobileMaxItems" | 1.5 | 2.25 | 3 | 6 | 9
"@selectionTabletMaxItems" | 2 | 3 | 4 | 8 | 12
"@selectionComputerMaxItems" | 3 | 4.5 | 6 | 12 | 18
"@selectionWidescreenMaxItems" | 4 | 6 | 8 | 16 | 24
(unit: items)

lubber-de:
> We should also add `short` and `very short` so it stays consistent.

🤔 How should it behave then? So which max-height setting should be set when using `short` or `very short`.... i would leave the PR as it is now

hammy2899:
I think the table @exoego shows is perfect, I would agree that `short` and `very short` will not be used much but its best to keep the consistency especially when its a simple change.

PR #603, opened 2019-03-31
[Chore] Switched comment blocks to Fomantic UI references

lubber-de:
## Description
This PR just changes all Semantic UI References in code comment block headers (and in some parts of the readme) to Fomantic-UI references instead. Also in terms of separate npm/cdnjs/jsdlivr usage and the current situation of fomantic surpassing semantic by > 300 bugfixes and enhancements, i think this should be adjusted as long as fomantic will stay separate.

Feel free to decline this PR if you disagree.

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -1,6 +1,6 @@
 /*!
- * # Semantic UI - API
- * http://github.com/semantic-org/semantic-ui/
+ * # Fomantic UI - API
+ * http://github.com/fomantic/fomantic-ui/
  *
  *
  * Released under the MIT license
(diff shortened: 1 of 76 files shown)


src/definitions/behaviors/api.js
```diff
@@ -1,6 +1,6 @@
 /*!
- * # Semantic UI - API
- * http://github.com/semantic-org/semantic-ui/
+ * # Fomantic UI - API
+ * http://github.com/fomantic/fomantic-ui/
```
hammy2899: `Fomantic UI` should be `Fomantic-UI` and `http://github.com/fomantic/fomantic-ui/` should be `http://github.com/fomantic/Fomantic-UI` to keep consistency with other places we link to the repo.

hammy2899: These changes should be for all files.

lubber-de: Changed accordingly by a2b65e71322e3cab8d2e6d8084ea3a051bdb9524

PR #612, opened 2019-04-02
[Dropdown] inverted pointing menu had white arrow (also moved code from overrides to less file)

lubber-de:
### Fixes
- An `inverted pointing dropdown` still had the white arrow color from the normal dropdown
- I also moved the whole code for `inverted dropdown` from the `.overrides` file to the `.less` file, because it's a core feature.
- `@pointingArrowBoxShadow` was doubled using the same values in the `.variables` file (line 370 & Line 375), that's why i removed one of them

## Testcase
https://jsfiddle.net/du7noc0q/
Remove CSS to see the issue

## Screenshot
|Before|After|
|-|-|
|image|image|


lubber-de:
Let's merge this (everybody approved 😉) before #626 because I guess I have to merge the result into #626 (I forgot about the fact I already moved the inverted dropdown code here)

PR #625, opened 2019-04-05
[Progress] Display 0-width bar when value is 0

### LINKED ISSUE #624: [Progress] Progress bar is not shown when value is 0
# Bug Report

## Steps to reproduce
* Instantiate `progress` with `value: 0`, or
* Update `value` to `0`

## Expected result

`value: 0` should be accepted and empty progress bar should be shown. 

## Actual result
Progress bar is hidden.

## Testcase
case 1. Value is given via `data-value` 
https://jsfiddle.net/Ld6kjqe2/

case 2.  Value is given via JavaScript API.
https://jsfiddle.net/9r17bphw/

## Screenshots
progress-broken-when-value-is-0

## Version
2.7.4

exoego:
### Before
image

### After
fix-progress-broken-when-value-is-0

## Closes
#624

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -423,11 +423,19 @@ $.fn.progress = function(parameters) {
             else {
               var firstNonZeroIndex = -1;
               var lastNonZeroIndex = -1;
+              var percentSum = 0;
+              var barCounts = $bars.length;
+              var isMultiple = barCounts > 1;
               var percents = values.map(function(value, index) {
+                percentSum += value;
+                var allZero = (index === barCounts - 1 && percentSum === 0);
                 var $bar = $($bars[index]);
-                if (value === 0) {
+                if (value === 0 && isMultiple && !allZero) {
                   $bar.css('display', 'none');
                 } else {
+                  if (isMultiple && allZero) {
+                    $bar.css('background', 'transparent');
+                  }
                   if (firstNonZeroIndex == -1) {
                     firstNonZeroIndex = index;
                   }
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -423,12 +423,11 @@ $.fn.progress = function(parameters) {
             else {
               var firstNonZeroIndex = -1;
               var lastNonZeroIndex = -1;
-              var percentSum = 0;
+              var valuesSum = module.helper.sum(values);
               var barCounts = $bars.length;
               var isMultiple = barCounts > 1;
               var percents = values.map(function(value, index) {
-                percentSum += value;
-                var allZero = (index === barCounts - 1 && percentSum === 0);
+                var allZero = (index === barCounts - 1 && valuesSum === 0);
                 var $bar = $($bars[index]);
                 if (value === 0 && isMultiple && !allZero) {
                   $bar.css('display', 'none');


src/definitions/modules/progress.js
```diff
@@ -425,18 +425,13 @@ $.fn.progress = function(parameters) {
               var lastNonZeroIndex = -1;
               var percents = values.map(function(value, index) {
                 var $bar = $($bars[index]);
-                if (value === 0) {
```
lubber-de: Your change won't work for your new multiple bars. Every 0% bar within a multiple bar would have a default width then.
A more easier fix would be to check for the `$bars` array length instead and leave the old code as it was so far
```javascript
if (value === 0 && $bars.length>0) {
  $bar.css('display', 'none');
} else {
if (firstNonZeroIndex == -1) {
...
```
image

exoego: Oops.

That fix can not address when every bars are 0% for multiple progress bars: empty progress bar is not displayed.

lubber-de: This works for me:
```javascript
...
var percentSum = 0, isMultiple = $bars.length>1;
var percents = values.map(function(value, index) {
  percentSum+=value;
  var allZero = (index===$bars.length-1 && percentSum===0);
  var $bar = $($bars[index]);
  if (value === 0 && isMultiple && !allZero) {
    $bar.css('display', 'none');
  } else {
    if(isMultiple && allZero) {
	  $bar.css('background', 'transparent');
    }
    if (firstNonZeroIndex == -1) {
	  firstNonZeroIndex = index;
    }
    lastNonZeroIndex = index;
    $bar.css({
	  display: 'block',
	  width: value + '%'
    });
  }
  ...
```

exoego: Thanks!! 🙏 
Fixed in https://github.com/fomantic/Fomantic-UI/commit/a6a07ad93a315e9b1cd1d92239b9d358ba3496ac

src/definitions/modules/progress.js
```diff
@@ -423,11 +423,19 @@ $.fn.progress = function(parameters) {
             else {
               var firstNonZeroIndex = -1;
               var lastNonZeroIndex = -1;
+              var percentSum = 0;
```
lubber-de: while we are at it, this can be cleaned up a bit more, because you already have implemented a helper for the sum functionality
```javascript
var percentSum = module.helper.sum(values);
```

src/definitions/modules/progress.js
```diff
@@ -423,11 +423,19 @@ $.fn.progress = function(parameters) {
...
+              var barCounts = $bars.length;
+              var isMultiple = barCounts > 1;
               var percents = values.map(function(value, index) {
+                percentSum += value;
```
lubber-de: ..then this line can be removed (because in the end it only counts if **all** values are zero

exoego: Fixed in https://github.com/fomantic/Fomantic-UI/pull/625/commits/730f0c12011d18b14ae760bcd443ec46571b1f09

src/definitions/modules/progress.js
exoego: Fixed in https://github.com/fomantic/Fomantic-UI/pull/625/commits/730f0c12011d18b14ae760bcd443ec46571b1f09

PR #629, opened 2019-04-05
[Form] Adds segment corner rounding to form segments loader

### LINKED ISSUE #628: [Form Segment] Form group of segments lost rounded corner when loading
# Bug Report
If the group of segments is form and it has CSS class name `loading`, it lost the rounded corner.

## Steps to reproduce
1. Create the form with class name `ui form segments`.
2. Add the form fields within each `ui segment` container.
3. Manipulate the form to put loading state. Or add class name `loading` to `ui form segments`.

## Expected result
The segments should keep rounded corners.

## Actual result
The rounded corners become square until the class name `loading` is removed.

## Testcase
Open https://jsfiddle.net/mgfLs0e1/, and click the submit button to reproduce.

## Screenshot
FomanticUI-Segments

## Version
2.7.4

douglasg14b:
## Description

The ::before `.loader` pseudo-element did not have segment rounded corners on a `.segments.form`. Added those.

## Testcase
https://jsfiddle.net/douglasg14b/e7cs93zd/4/

## Closes
#628

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -651,7 +651,7 @@
 }
 
 .ui.loading.form.segments:before {
-  border-radius: @borderRadius;
+  border-radius: @defaultBorderRadius;
 }
 
 .ui.loading.form:after {
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -649,6 +649,11 @@
   height: 100%;
   z-index: @loaderDimmerZIndex;
 }
+
+.ui.loading.form.segments:before {
+  border-radius: @borderRadius;
+}
+
 .ui.loading.form:after {
   position: absolute;
   content: '';


src/definitions/collections/form.less
```diff
@@ -649,6 +649,11 @@
   height: 100%;
   z-index: @loaderDimmerZIndex;
 }
+
+.ui.loading.form.segments:before {
+  border-radius: @borderRadius;
```
lubber-de: `@borderRadius` is not avaiable at that time. I think you are referring to that variable from `segments.variables`
So either add
```less
@borderRadius: @defaultBorderRadius;
```
to `form.variables`

or use

```less
border-radius: @defaultBorderRadius;
```

instead (because `@defaultBorderRadius` is available globally from `sites.variables`

douglasg14b: Oh, gotcha. I'll change that.

douglasg14b:
@lubber-de  What does commitLint mean by message may not be empty and type may not be empty?

lubber-de:
@Douglasg14b your commits should follow the angular commit style as you did for the last commit. In general each commit should have a type and the module in parentheses set following a colon as a prefix. 
`fix(form): Corrects border radius variable reference`
was totally fine... But it seems the bot also checked all of your commits for the whole PR each time, so it was always moaning about the first commit message which was missing the desired prefix

ko2in:
Thanks both of you. Have a nice weekend.

PR #632, opened 2019-04-06
[Menu]: Fixes right dropdown sub-menu not being vertical

### LINKED ISSUE #382: Drop down with sub-menu doesn't display in menu
# Bug Report

A drop down with a sub menu displays inline if it's placed inside another menu. See the attached jfiddle.

## Expected result
Each option within sub menu should shown on its own line.

## Actual result
All options are displayed on the same line.

## Testcase
https://jsfiddle.net/u1f02thp/

douglasg14b:
This style was already added to the left menu

## Description

Fixes the issue where a right sub-menu dropdown of a menu would open up horizontally instead of vertically. This style was already added for the left menu, but was not for the right menu.

## Testcase
https://jsfiddle.net/986c0o5L/1/

## Closes

#382

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -483,7 +483,7 @@ Floated Menu / Item
 }
 /* Right Floated */
 .ui.menu:not(.vertical) .right.item,
-.ui.menu:not(.vertical) .right.menu {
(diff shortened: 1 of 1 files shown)


hammy2899:
@all-contributors please add @douglasg14b for code and talk

PR #638, opened 2019-04-08
[Menu] Cursor on item should be pointer

### LINKED ISSUE #637: [Menu] Mouse cursor should be pointer on div.item
# Feature request

## Steps to reproduce
1. Go to https://fomantic-ui.com/modules/tab.html#centered-item
2. Click tabs

## Expected result

Mouse cursor should be `pointer` on `tab` so that it looks clickable.
This seems possible if tab item is marked up as `a.item`.

## Actual result

Cursor do not change (keep `default`) if `div.item`.

## Screenshot (when possible)
cursor-tab

## Version
2.7.4

exoego:
### Before
image

### After
image

## Closes
#637

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -84,7 +84,7 @@
   -webkit-tap-highlight-color: transparent;
   flex: 0 0 auto;
   user-select: none;
-
+  cursor: pointer;
   background: @itemBackground;
   padding: @itemVerticalPadding @itemHorizontalPadding;
   text-transform: @itemTextTransform;
@@ -412,7 +412,6 @@
 .ui.menu .dropdown.item:hover,
 .ui.menu .link.item:hover,
 .ui.menu a.item:hover {
-  cursor: pointer;
   background: @hoverItemBackground;
   color: @hoverItemTextColor;
 }


src/definitions/modules/tab.less
```diff
@@ -25,6 +25,9 @@
 .ui.tab {
   display: none;
 }
+.ui.tab > .item {
```
lubber-de: By design the tab items are only set within menus not within the tab itself (which just holds the content).
There also is not other `.item` setting in `tab.less`  and there is no docs example 
That said, i think this is not neccessary and could be removed.

exoego: Fixed in
https://github.com/fomantic/Fomantic-UI/pull/638/commits/3ababe893b478eb4f2ea0182213221197eb32c4b

Removed `.ui.tab. > .item`.

And applied changes to not only `.ui.tabular.menu .item`  but every `.ui.menu .item` since `div.item` may be used in vertical menu or other menuvariation.

donaggio: For generic menus there's already a .link.item variation which makes them hoverable. If every .ui.menu .item gets  a pointer cursor, doesn't it make .link.item useless?

lubber-de: @donaggio Good catch! I'll add a comment to the other codepart of the PR

src/definitions/collections/menu.less
```diff
@@ -84,7 +84,7 @@
   -webkit-tap-highlight-color: transparent;
   flex: 0 0 auto;
   user-select: none;
-
+  cursor: pointer;
```
lubber-de: As @donaggio pointed out, i also don't think it is correct to __always__ set the cursor to pointer to __any__ kind of menu item (not only tabs would be affected)

So what about adding the `cursor:pointer` only to `.ui.tabular.menu .item` instead?

exoego: I think that menu items should always set the cursor to `pointer` except `.disabled` item, so that end-users can recognize items are clickable, in either case `a.item` or `div.item`.

As a SUI/FUI user, it is not clear for me that `div.item` requires `.link` class to be link-ish while `a.item` requires no `.link` even when no `href` given.
I even don't known there is `.link.item` until today.
I think this is a sign of confusing/complex point of SUI/FUI and should be simplified.

lubber-de: ok, convinced. 😄 
Btw the `link` is still necessary to provide hovering functionality to change color/background on a `did` tag, so completely remove `.link` wouldn't be an option (as @donaggio propably guessed)

hammy2899: I'm not convinced yet, don't get me wrong I complete agree with your points and I feel that way too about how confusing it is when you create a menu then the items don't seem clickable I just want to try and see if there was any comments made by Jack about this before we move forward.

donaggio: Menu items can contain just about anything. This can be really useful sometimes, especially with vertical menus: I know I used it to put an entire mini-form in it one time, and I certainly didn't want the mouse cursor to appear as if the whole form was a link.

hammy2899: @donaggio Agreed, I can't find any comments on the issue from Jack but your point is enough for me to deny this change this is a good example https://fomantic-ui.com/collections/menu.html#secondary-menu of what your saying.

exoego: @donaggio
That makes sense to keep div.item as simple container without pointer cursor.
I  think it is better to close this PR/issue as wontfix.
There are some use cases and some solutions (link item or a.item without href).

Suggestions are welcome.

lubber-de: So we don't even apply the pointer to the tabular menu item only then? , the original intention of the related issue)

exoego: No, we don't apply this PR to tabular menu.
There is tabular menu example of docs having search input in item.

lubber-de:
After conversation, it's up to @exoego to prematurely close this PR by himself

src/definitions/collections/menu.less
hammy2899: Yeah we can do that would make sense.

PR #641, opened 2019-04-08
[build] Task `watch` do not update files that depends on site.variable

### LINKED ISSUE #631: [build] Task `watch` do not update files that depends on site.variable
# Bug Report

## Steps to reproduce
1. `mkdir fui-test; cd fui-test`
2. `npm install --save fomantic-ui@2.7.4`
3. Choose `Automatic install`
4. `cd semantic`
5. `gulp build`
6. `gulp watch`
7. In different terminal, update `src/site/globals/site.variables` like below

```less
// those variables are widely used in many components
@red: #ff0000;
@h1 : unit(123, rem);
@textColor: rgba(210, 200, 200, 1.0);
```
8. Wait until watch task updates the files.

## Expected result

Every components that depends on modified variables should be re-generated.
This is possible in Semantic-UI@2.4.2.

## Actual result

It seems like that only `site(.min).css` are updated when `site.variables` are modified.
Other components are not updated and use old variables.
So, manully running `gulp build` is required to updated other components .

In short, the gulp task `watch` is useless.

## Version
2.7.4

ColinFrick:
## Description
If changes are detected in the site.variables file, all css files should be rebuilt.

## Closes
#631

### DIFF
--- a/tasks/build/css.js
+++ b/tasks/build/css.js
@@ -192,7 +192,10 @@ module.exports.watch = function (type, config) {
   const method = type === 'docs' ? docs : rtlAndNormal;
 
   // Watch theme.config file
-  gulp.watch([normalize(config.paths.source.config)])
+  gulp.watch([
+    normalize(config.paths.source.config),
+    normalize(config.paths.source.site + '/**/site.variables')
+  ])
     .on('all', function () {
       // Clear timeout and reset files
       timeout && clearTimeout(timeout);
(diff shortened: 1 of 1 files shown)


exoego:
Is `site.variable` the only file that are depended by `*.(less|variable|overrides)`?

ColinFrick:
@exoego Yes, this works the same as the Semantic watch task now.
https://github.com/Semantic-Org/Semantic-UI/blob/deb275d2d5fe9a522a0b7bd8b6b6a1c939552718/tasks/watch.js#L109-L119

PR #642, opened 2019-04-08
[API] Set context in beforesend to $module instead of possible stateContext to match docs

lubber-de:
According to [the docs](https://fomantic-ui.com/behaviors/api.html#/settings), `beforeSend` should have the "initialized element" as context, not the element matching `stateContext`!

### Current/Wrong
https://jsfiddle.net/7oz0bncp/
### Fixed
https://jsfiddle.net/7oz0bncp/1/

### Current/Wrong
beforesend_wrong

### Fixed
beforesend_right

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6207


hammy2899:
Again, I am sorry about the previous PR

PR #648, opened 2019-04-08
fix(form): Remove unnecessary property '-webkit-appearance: none' with :focus from error field

### LINKED ISSUE #576: [Form] Native drop-down selection in error field has inconsistent size
# Bug Report
The native drop-down selection within `field error` has inconsistent size and changing it's width and height when the mouse focus somewhere on the page. It also causes the other component around being inconsistent in the page.

## Steps to reproduce
1. Add `field` container with `<select>` in it. The select element shouldn't have special drop-down class `ui dropdown`.
2. Manipulate the `field` container to have `error` class, or add `error` class manually.
3. Now, click the `<select>` element and then click somewhere on the page

## Expected result
The drop-down should react as naturally. 

## Actual result
The drop-down changes it's width and height and moves itself obviously and other components around.

## Testcase
https://jsfiddle.net/nr7sgu5f/

## Screenshot
Select

## Version
2.7.2

ko2in:
## Description
The property `-webkit-appearance: none` causes the select element unstable on focus and unfocus
events. Removing native appearance from select on focus reduces the space between the border
and it's content which causes unexpected bouncing.

This property doesn't give any extra effect to all fields for focus event, but making select
element inconsistency alone. So, better to remove this property to make the select element stable
along with the other fields when it's focused or unfocused.

## Screenshot
Before | After
--------- | ---------
before | after

## Closes
#576

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -552,7 +552,6 @@
   border-color: @inputErrorFocusBorder;
   color: @inputErrorFocusColor;
 
-  -webkit-appearance: none;
   box-shadow: @inputErrorFocusBoxShadow;
 }
 


exoego:
It seems like this PR cause a new cosmetic issue: this grey empty box
image

ko2in:
@exoego That was from my byzanz screen recorder. It produced poor quality gif.

PR #658, opened 2019-04-10
fix(dropdown): margin missing when the position of floating menu of dropdown is upward

dutrieux:
## Description

When the position of floating menu of dropdown is upward, the margin is missing

## Screenshot (when possible)
**Before**
dropdown

**after**
dropdown

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1302,10 +1302,10 @@ select.ui.dropdown {
   border-radius: @floatingMenuBorderRadius !important;
 }
 .ui:not(.upward).floating.dropdown > .menu {
-  margin-top: @floatingMenuDistance !important;
+  margin-top: @floatingMenuDistance;
 }
 .ui.upward.floating.dropdown > .menu {
-  margin-bottom: @floatingMenuDistance !important;
+  margin-bottom: @floatingMenuDistance;
 }
 
 /*--------------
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.less
```diff
@@ -1299,9 +1299,14 @@ select.ui.dropdown {
   border-radius: @floatingMenuBorderRadius !important;
 }
 .ui.floating.dropdown > .menu {
-  margin-top: @floatingMenuDistance !important;
   border-radius: @floatingMenuBorderRadius !important;
 }
+.ui:not(.upward).floating.dropdown > .menu {
+  margin-top: @floatingMenuDistance !important;
```
lubber-de: Remove the !important, it's not needed

src/definitions/modules/dropdown.less
```diff
@@ -1299,9 +1299,14 @@ select.ui.dropdown {
...
+}
+.ui.upward.floating.dropdown > .menu {
+  margin-bottom: @floatingMenuDistance !important;
```
lubber-de: Also here: !important is not needed

src/definitions/modules/dropdown.less
dutrieux: That's how it is on version 2.7.4 : sure you want I change that ?

lubber-de: Yes, i tested it before posting and the !important is still inherit from SUI, not a specific 2.7.4 feature.
But of course, i advise you to test it and get convinced yourself.

PR #659, opened 2019-04-10
[Card] Force horizontal cards meta, extra content and actions to align properly

ko2in:
## Description
As reported in #159, the meta data, actions and extra content are forced to display in one row in horizontal cards. It's because of `flex-direction: row` setting `flex-wrap: wrap`, and the flex item for meta data has no initial size. So, it forces the other flex items to wrap in available spaces next to it which leads them displaying in the same row.

Since, the extra content and buttons are always supposed to appear on the next line and should take 100% width of the card. The meta data should have initial size with minimum value which will make sure the meta data to wrap in the available space on the current row and the extra content and buttons to appear in the next row respectively.

## Testcase
**Before:** https://jsfiddle.net/kv2sj3hq/
**After:** https://jsfiddle.net/axc9hmqy/

## Screenshot
Before | After
--------- | --------
Before-1 | After-1 |
----------------- | -----------------
Before-2 | After-2
----------------- | ----------------- 
Before-3 | After-3

Works on Firefox, Google Chrome, Opera, Microsoft Edge on Windows 10. And Firefox and Chromium on Ubuntu.

## Closes
I hope this PR fixes the issue #159.

### DIFF
--- a/src/definitions/views/card.less
+++ b/src/definitions/views/card.less
@@ -430,6 +430,7 @@
 .ui.horizontal.cards > .card,
 .ui.card.horizontal {
   flex-direction: row;
+  flex-wrap: wrap;
   min-width: @horizontalMinWidth;
   width: @horizontalWidth;
   max-width: 100%;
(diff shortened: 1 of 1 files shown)


hammy2899:
@all-contributors please add @ko2in for code

PR #661, opened 2019-04-11
[Form] Add dirty and clean states, and ability to cancel submit

prudho:
First, a great kudo to @simontaite for inspiring me this. I'm using his [plugin](https://github.com/simontaite/jquery.dirty) on a lot of projects and it works well with FUI.

This PR add the ability to define a state for the form: `clean` or `dirty`. Default state is `clean` and is changed to `dirty` when the user change at least one field. Reseting a form also change state to `clean`. Theses events are now triggered:
- `onDirty()`
- `onClean()`

Dirty fields can be showed by using the `get dirty fields` behavior.

State can also be changed programmaticaly by using behaviors `set as clean` and `set as dirty`. Its now also possible to prevent a user form leaving the page if the form is in `dirty` state, using `preventLeaving: true`, which can be useful when you don't want to loose data.

Bonus point: Dropdown are now correctly restored when calling the behavior `reset`.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -68,6 +68,10 @@ $.fn.form = function(parameters) {
         moduleNamespace,
         eventNamespace,
 
+        submitting = false,
+        dirty = false,
+        history = ['clean', 'clean'],
+
         instance,
         module
       ;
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -382,6 +450,19 @@ $.fn.form = function(parameters) {
...
               }
             }
+          },
+          beforeUnload: function(event) {
+            if (module.is.dirty() && !submitting) {
+              var event = event || window.event;
+
+              // For modern browsers
+              if (event) {
+                event.returnValue = settings.text.leavingMessage;
```
lubber-de: According to https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event the text itself is only supported on Internet explorer and is otherwise just a trigger (content does not matter) to display the native browser text, so keep this in mind, when writing the docs. 
Try https://jsfiddle.net/cbojyq1x/2/ . Chrome for example will always tell  you its own native leaving message instead of the value of `leavingMessage`

prudho: Yes, I saw that too, and users will be noted about this. If we decide to not support IE in the future, i'll drop it asap 😈 !

PR #678, opened 2019-04-19
feat(form): add calendar input field

prudho:
This PR add calendar as a form field, plus it implement its validation and allow to grab underlying value while validating in three different formats.

To do so, a new parameter has been added to the settings: `dateHandling`. It can contain one of these three string:
- `date` (default): field value will be the selected date as a string.
- `input`: field value will be the underlying input value of the calendar.
- `formatter` will let you customize the format of the selected date (or datetime, time, month and year).

Default formatter settings is:
```js
formatter: {
    date: function(date) {
      var date = new Date(date);

      return Intl.DateTimeFormat('en-GB', {
        year: "numeric",
        month: "2-digit",
        day: "2-digit"
      }).format(date);
    },
    datetime: function(date) {
      return Intl.DateTimeFormat('en-GB', {
        year: "numeric",
        month: "2-digit",
        day: "2-digit",
        hour: '2-digit',
        minute: '2-digit',
        second: '2-digit'
      }).format(date);
    },
    time: function(date) {
      return Intl.DateTimeFormat('en-GB', {
        hour: '2-digit',
        minute: '2-digit',
        second: '2-digit'
      }).format(date);
    },
    month: function(date) {
      return Intl.DateTimeFormat('en-GB', {
        month: '2-digit',
        year: 'numeric'
      }).format(date);
    },
    year: function(date) {
      return Intl.DateTimeFormat('en-GB', {
        year: 'numeric'
      }).format(date);
    }
  }
```

As you can see the default format is the english format, since its [the most used in the world](https://en.wikipedia.org/wiki/Date_format_by_country), but exposing this formatter allow editing it.

Also, the calendar module got a new behavior that return the input text, `get input date`.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -660,13 +660,16 @@ $.fn.form = function(parameters) {
             ;
             $fields.each(function(index, field) {
               var
-                $field     = $(field),
-                name       = $field.prop('name'),
-                value      = $field.val(),
-                isCheckbox = $field.is(selector.checkbox),
-                isRadio    = $field.is(selector.radio),
-                isMultiple = (name.indexOf('[]') !== -1),
-                isChecked  = (isCheckbox)
+                $field       = $(field),
+                $parent      = $field.parent(),
+                $grandParent = $parent.parent(),
+                name         = $field.prop('name'),
+                value        = $field.val(),
+                isCheckbox   = $field.is(selector.checkbox),
+                isRadio      = $field.is(selector.radio),
+                isMultiple   = (name.indexOf('[]') !== -1),
+                isCalendar   = ($parent.is(selector.uiCalendar) || $grandParent.is(selector.uiCalendar)),
+                isChecked    = (isCheckbox)
                   ? $field.is(':checked')
                   : false
               ;
@@ -705,7 +708,60 @@ $.fn.form = function(parameters) {
                       values[name] = false;
                     }
                   }
-                  else {
+                  else if(isCalendar) {
+                    // We must determine which item is the calendar module
+                    var $calendar = $parent.is(selector.uiCalendar)
+                      ? $parent
+                      : $grandParent.is(selector.uiCalendar)
+                        ? $grandParent
+                        : false
+                    ;
+
+                    if ($calendar) {
+                      var date = $calendar.calendar('get date');
+
+                      if (date !== null) {
+                        if (settings.dateHandling == 'date') {
+                          values[name] = date;
+                        } else if(settings.dateHandling == 'input') {
+                          values[name] = $calendar.calendar('get input date')
+                        } else if (settings.dateHandling == 'formatter') {
+                          var type = $calendar.calendar('get mode');
+
+                          switch(type) {
+                            case 'date':
+                            values[name] = settings.formatter.date(date);
+                            break;
+                            
+                            case 'datetime':
+                            values[name] = settings.formatter.datetime(date);
+                            break;
+                            
+                            case 'time':
+                            values[name] = settings.formatter.time(date);
+                            break;
+                            
+                            case 'month':
+                            values[name] = settings.formatter.month(date);
+                            break;
+                            
+                            case 'year':
+                            values[name] = settings.formatter.year(date);
+                            break;
+    
+                            default:
+                            module.debug('Wrong calendar mode', $calendar, type);
+                            values[name] = '';
+                          }
+                        }
+                      } else {
+                        values[name] = '';
+                      }
+                    } else {
+                      module.debug('Unable to find calendar module', $field);
+                      values[name] = '';
+                    }
+                  } else {
                     values[name] = value;
                   }
                 }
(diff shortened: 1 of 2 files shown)


src/definitions/behaviors/form.js
```diff
@@ -660,13 +660,16 @@ $.fn.form = function(parameters) {
...
-                isChecked  = (isCheckbox)
+                $field       = $(field),
+                $parent      = $field.parent(),
+                $grandParent = $parent.parent(),
+                name         = $field.prop('name'),
+                value        = $field.val(),
+                isCheckbox   = $field.is(selector.checkbox),
+                isRadio      = $field.is(selector.radio),
+                isMultiple   = (name.indexOf('[]') !== -1),
+                isCalendar   = ($parent.is(selector.uiCalendar) || $grandParent.is(selector.uiCalendar)),
```
lubber-de: What about using the `.closest()` functionality instead? You could get rid of the additional $parent/$grandparent variables and, in case the DOM structure changes, the code is already prepared prepared:

```javascript
$calendar = $field.closest(selector.uiCalendar),
isCalendar   = ($calendar.length>0),
```

src/definitions/behaviors/form.js
```diff
@@ -705,7 +708,60 @@ $.fn.form = function(parameters) {
                       values[name] = false;
                     }
                   }
-                  else {
+                  else if(isCalendar) {
+                    // We must determine which item is the calendar module
+                    var $calendar = $parent.is(selector.uiCalendar)
```
lubber-de: If using the above mentioned suggestion, this code can be reduced to `$calendar = $calendar[0] || false;`

src/definitions/behaviors/form.js
```diff
@@ -1486,6 +1544,46 @@ $.fn.form.settings = {
     }
   },
 
+  formatter: {
+    date: function(date) {
+      var date = new Date(date);
+
+      return Intl.DateTimeFormat('en-GB', {
+        year: "numeric",
```
lubber-de: Not needed. en-GB returns d/m/y by default

```javscript
return Intl.DateTimeFormat('en-GB').format(date);
```
is sufficient

lubber-de:
I think the `reset` and `clear` functions have to be adjusted as well, because they will empty the form field, but the underlying date within the calendar module stays untouched

lubber-de:
What about adding rules for the date field like `time`, `date`, `datetime` ? You added the format to the `get values` behavior / function, but the only possible validation would be `empty`

Or did i miss something? :thinking:

lubber-de:
`Intl.DateTimeFormat.protoype.format()` is not available for Android Webview (but `formatToParts` is).
Here is a simple polyfill, which we might add.
```javascript
    if(!('format' in Intl.DateTimeFormat.prototype)){
        Intl.DateTimeFormat.prototype.format = function(date){
            return this.formatToParts(date).map(function(o){return o.value}).join('');
        }
    }
```

src/definitions/behaviors/form.js
prudho: Wow good catch here ! I think that I have over complicated the code 👍

src/definitions/behaviors/form.js
prudho: With the simplification, I think that this piece of code is now useless...

src/definitions/behaviors/form.js
prudho: True. I really need some sleep !

prudho:
> I think the `reset` and `clear` functions have to be adjusted as well, because they will empty the form field, but the underlying date within the calendar module stays untouched

Yup, this is now done.

> What about adding rules for the date field like `time`, `date`, `datetime` ? You added the format to the `get values` behavior / function, but the only possible validation would be `empty`
> 
> Or did i miss something? 🤔

Yes, actually providing a validation other than `empty` is pretty hard to put up. The returned value depends on the type of calendar and the selected `dateHandling` parameter and defining rule for each case would be a real PITA. Plus don't forget that the calendar module already return a formatted value, so why re-validating it ?

lubber-de:
> Plus don't forget that the calendar module already return a formatted value, so why re-validating it ?
OK, then let's leave it for now 🙂

src/definitions/behaviors/form.js
```diff
@@ -1486,6 +1540,42 @@ $.fn.form.settings = {
     }
   },
 
+  formatter: {
+    date: function(date) {
+      var date = new Date(date);
```
lubber-de: Why is this necessary here? It's not used in the other formatters? :thinking:

prudho: Mmmmh. I defined it during my tests, and did not implement it in other formatters. The `Intl.DateTimeFormat.prototype.format()` seem to need a date as input, but looks like he perfectly handle a date string. So, should I remove `var date = new Date(date);` from this one or implement a Date object in other ?

lubber-de: In Line 721 you added 
```javascript
var date = $calendar.calendar('get date');
```
which will already return a valid Date Object (or null/undefined) and is passed to all the formatters, so you can remove the `var date=new Date(date)`

lubber-de: As said, you can safely remove the ´var date=new Date(date)` line.
I would love to approve this PR right after that 😉

prudho: Done.

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/131

PR #680, opened 2019-04-20
improve label positioning for vertical pointing menu #668

### LINKED ISSUE #668: Positioning issue of floating label with pointing menu
# Bug Report
Positioning of floating label with pointing menu is not optimal

## Steps to reproduce
See fiddle

## Expected result
The floating label should be better positioned with pointing menus

## Actual result
Positioning is too high

## Testcase

https://jsfiddle.net/kswyvoab/

## Screenshot (when possible)
Normal menu
Screenshot_20190415_084129

Pointing menu
Screenshot_20190415_084111

## Version
2.7.2

aardbol:
## Description
This PR will improve the positioning of labels for secondary pointing menus

## Testcase

https://jsfiddle.net/fkudetaw/

## Screenshot (when possible)
Evil:
Screenshot_20190420_111301

Angelic:
Screenshot_20190420_111234

## Closes
#668

### DIFF
--- a/src/themes/default/collections/menu.variables
+++ b/src/themes/default/collections/menu.variables
@@ -119,7 +119,6 @@
 
 @labelTextMargin: 1em;
 @labelVerticalPadding: 0.3em;
-@labelVerticalPaddingSecondaryMenu: 1.2em;
 @circularLabelVerticalPadding: 0.5em; /* has to be equal to @circularPadding from label.less */
 @labelHorizontalPadding: @relativeMini;
 
--- a/src/themes/default/collections/menu.variables
+++ b/src/themes/default/collections/menu.variables
@@ -119,6 +119,7 @@
 
 @labelTextMargin: 1em;
 @labelVerticalPadding: 0.3em;
+@labelVerticalPaddingSecondaryMenu: 1.2em;
 @circularLabelVerticalPadding: 0.5em; /* has to be equal to @circularPadding from label.less */
 @labelHorizontalPadding: @relativeMini;
 
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -942,12 +942,17 @@ Floated Menu / Item
 }
 
 .ui.secondary.pointing.menu .item > .label {
-  margin-top: -@labelVerticalPadding;
-  margin-bottom: -@labelVerticalPadding;
+  margin-top: @labelVerticalPaddingSecondaryMenu;
+  margin-bottom: unset;
 }
 .ui.secondary.pointing.menu .item > .circular.label {
   margin-top: -@circularLabelVerticalPadding;
   margin-bottom: -@circularLabelVerticalPadding;
+}
+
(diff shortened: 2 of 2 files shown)


src/themes/default/collections/menu.variables
```diff
@@ -119,6 +119,7 @@
 
 @labelTextMargin: 1em;
 @labelVerticalPadding: 0.3em;
+@labelVerticalPaddingSecondaryMenu: 1.2em;
```
exoego: Thanks for taking a time to open this PR. 👍 

I think this looks very nice if labels are very narrow.
However, I am afraid that it overlay the parent item if the label has 3 or more characters,
image

Another PR #669 looks like below, where I personally feels a gap is a bit longer.
image

What if we [shorten a gap between label and item like this](https://jsfiddle.net/o2ps4ve3/)?
image

lubber-de: I second @exoego . I also think 1.2em is too narrow for the mentioned reasons. However the suggestion from @exoego is still too narrow for me, because the top floating label will touch the icon and text of the adjacent item if the text gets too wide.
I suggest using `margin-top: 0`. See https://jsfiddle.net/mrupeckx/

lubber-de: More detailed reason to use `margin-top:0` : The gap is identical (even more when using 0.05em) for top/bottom to the item, especially when the item has a usual label inside. See https://jsfiddle.net/mrupeckx/2/
image

IsaakGroepT: Good idea. I didn't test it well enough to see these issues unfortunately

lubber-de: Will you continue to work in this PR?

IsaakGroepT: As in: consider the suggestions and modify the PR? Or as in: are you still doing something because I'd like to wrap this up?

lubber-de: Consider the suggestion as long as nobody else steps into the discussion 🙂

IsaakGroepT: Sorry, I don't know what to do now. Do I need to do something at the Changes requested  section?

lubber-de: If you follow my suggestion, then you only need to change the code in your branch accordingly and push the changes. Then this PR gets updated automatically and I can approve it. 🙂

IsaakGroepT: Done :+1:

lubber-de:
With the current changes, your PR is now identical to my previous PR #688 
Using `unset` is/was just necessary to make the changes work in a jsfiddle using the current 2.7.4
But for a new 2.7.5 setting `unset` equal to removing the block or limit it to `:not(.floating)`

I appreciate your effort and thank you for your time 👍  but this PR is now a duplicate of #688, which is already approved

ColinFrick:
As @lubber-de mentioned duplicate of #669 Closing this now

PR #686, opened 2019-04-23
fix(*): breakpoints ignoring next breakpoint in some cases

### LINKED ISSUE #681: [Table] starting at 2.7.4 on iPad tables look like iPhone
# Bug Report
I believe this could be a break point issue with media types.  Tables on iPad look like phone meaning table cells stack up as rows.  This is behaviour desired for phones not iPad.  You can use Chrome developer view set to iPad to reproduce this.

[Update] This is actually not only for tables.  All stackable grids start stacking in iPad which they shouldn't.

## Steps to reproduce
1. In Chrome go to docs, go to Tables: https://fomantic-ui.com/collections/table.html
2. Use inspector and set device to iPad (not iPad Pro)

## Expected result
Table cells should remain as cells

## Actual result
Table cells stack up into rows as they would on iPhone

## Version
2.7.4

lubber-de:
## Description
Especially recognized in `table` or `grid` elements a breakpoint wasn't correctly fetched by the browsers, because the floating max value covers a 0.99 of the "before pixel" (was implemented by #567)
While mathematically correct, it was not working on the browser to correctly recognize the next full pixel again. (768 pixel were still recognized as something below)
I found out that it was working again when we'll use 0.98 of the breakpoint pixel (instead of 0.99.
This way the dpi fix of #567 still works but also fixes #681 now.
While i was testing with different dpi settings and windows resizings, i never came across a value of xxxx.99, so xxxx.98 as a breakpoint should fix both issues now.

## Testcase
https://fomantic-ui.com/collections/table.html
See gif-video:
- Open the mentioned website in chrome
- In developer tools toggle device toolbar and change to "iPad"
- change the breakpoint from 767.99 to 767.98 and the 768px width is now correctly recognized.

## Screenshot
breakpoint

## Closes
#681

### DIFF
--- a/src/themes/default/globals/site.variables
+++ b/src/themes/default/globals/site.variables
@@ -780,10 +780,10 @@
 
 
 /* Responsive */
-@largestMobileScreen : (@tabletBreakpoint - 0.01px);
-@largestTabletScreen : (@computerBreakpoint - 0.01px);
-@largestSmallMonitor : (@largeMonitorBreakpoint - 0.01px);
-@largestLargeMonitor : (@widescreenMonitorBreakpoint - 0.01px);
+@largestMobileScreen : (@tabletBreakpoint - 0.02px);
+@largestTabletScreen : (@computerBreakpoint - 0.02px);
+@largestSmallMonitor : (@largeMonitorBreakpoint - 0.02px);
+@largestLargeMonitor : (@widescreenMonitorBreakpoint - 0.02px);
 
 
 /*-------------------


prudho:
Have you made some tests with other browsers too ?

lubber-de:
> Have you made some tests with other browsers too ?

I just did for IE/Edge/Firefox on Windows (dpi 125%,150%,175%). IE and Edge were not even calculating any float values at all (at least i was not able to recognize or set it it devtools).
I don't have a Mac and Browserstack does not help, because i would need to change the DPI settings in the environment. That said, i cannot test Safari in this case.
I have Ubuntu 18 available in the evening today if you want me to test Chrome+Firefox there as well.

ko2in:
This fix works for Chrome on Ubuntu yet.

Ubuntu-Chrome-Table-Responsive(iPad)

Firefox doesn't even need this fix on my Ubuntu. It's already working with the current break point.

Firefox-Table-Responsive(iPad)

I hope this info helps for you.

PR #706, opened 2019-04-29
[Progess] Fix timing issue of updating value/percent

### LINKED ISSUE #705: [Progress] TypeError happens on module.value if progress is re-initiated
# Bug Report

## Steps to reproduce
1. Initiate `.progess` module
2. Then, re-initiate `.progress` module 
    * I am doing this to refresh progess after some user intercation

## Expected result

No error happens and `.progress` modules are re-initiated.

## Actual result

`TypeError: module.value is undefined`

## Testcase
https://jsfiddle.net/gscaLj35/1/

## Version
2.7.4

exoego:
## Testcase
https://jsfiddle.net/52wqjnxf/

## Closes
#705
#719

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -70,9 +70,9 @@ $.fn.progress = function(parameters) {
       module = {
         helper: {
           sum: function (nums) {
-            return nums.reduce(function (left, right) {
+            return Array.isArray(nums) ? nums.reduce(function (left, right) {
               return left + right;
-            }, 0);
+            }, 0) : 0;
           },
           forceArray: function (element) {
             return Array.isArray(element)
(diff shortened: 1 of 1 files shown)


src/definitions/modules/progress.js
```diff
@@ -70,9 +70,9 @@ $.fn.progress = function(parameters) {
       module = {
         helper: {
           sum: function (nums) {
-            return nums.reduce(function (left, right) {
+            return Array.isArray(nums) ? nums.reduce(function (left, right) {
```
exoego: While investigating this issue, `nums` are sometimes `undefined`,.
So I added safeguard for such situation.

src/definitions/modules/progress.js
```diff
@@ -333,10 +333,10 @@ $.fn.progress = function(parameters) {
           },
 
           percent: function(index) {
-            return module.percent[index || 0] || 0;
+            return module.percent && module.percent[index || 0] || 0;
```
exoego: Note) Not only `module.value` but also `module.pecent` has same issue, so I fixed this too.

lubber-de:
@exoego You can add you changed code directly via jsdelivr from github into your jsfiddle using the folliwing url-syntax:
https://cdn.jsdelivr.net/gh/exoego/fomantic-ui@fix-multiple-progress-timing/src/definitions/modules/progress.js

I changed your jsfiddle accordingly, works now. However the type error mentioned in #705 appears now
https://jsfiddle.net/52wqjnxf/

lubber-de:
Added the working testcase to your description. No type error anymore. All fine :)

src/definitions/modules/progress.js
lubber-de: 👍

src/definitions/modules/progress.js
lubber-de: 👍

exoego:
Added #719 as another issue resolved by this PR

PR #714, opened 2019-04-30
fix(list): align text in lists with icons

### LINKED ISSUE #597: [List] Icons in lists don't line up
# Bug Report

## Steps to reproduce
1. Make a list with different icons like in the List examples

## Expected result
The left side of the text lines up

## Actual result
The left side of the icon lines up

## Testcase

https://jsfiddle.net/4gkvha2b/

## Version
2.7.2

globophobe:
Define min-width for icon in lists, so text is aligned.

Closes #597

## Testcase
https://jsfiddle.net/vk2enzd5/

## Screenshot (when possible)
image

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -107,6 +107,7 @@ ol.ui.list ol li,
 .ui.list .list > .item > i.icon,
 .ui.list > .item > i.icon {
   display: table-cell;
+  min-width: 1.55em;
   margin: 0;
   padding-top: @iconOffset;
   transition: @iconTransition;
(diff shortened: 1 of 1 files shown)


src/definitions/elements/list.less
```diff
@@ -107,6 +107,7 @@ ol.ui.list ol li,
 .ui.list .list > .item > i.icon,
 .ui.list > .item > i.icon {
   display: table-cell;
+  min-width: 1.55em;
```
lubber-de: Could you please move the value into a variable?

hammy2899: I also thought about this, I don't think it should have a variable I think it should work it out from the icon font size so therefore it will change if you change your icon size.

lubber-de: OK, agreed

PR #715, opened 2019-05-01
fix(label): adjust basic ribbon label size and position

### LINKED ISSUE #708: [label] ribbon basic label had wrong calculated position
# Description

When a **basic** ribbon label was used there were slightly mispositioned making it overlap the border

# Screenshoot
## Actual result
ribbon_broken

## Expected result
ribbon_fix

## Testcase
https://jsfiddle.net/dutrieux/h5oapxfk/9/

## Fix 
I propose to remove the border of the **basic** label and use instead a box-shadow: inset, like that all the label will have the same height and fix the wrong calculated position :
https://jsfiddle.net/dutrieux/h5oapxfk/7/
I can do a PR if you want !

## Version
2.7.4

globophobe:
"ui basic ribbon" has a 1px border, but "ui ribbon" does not, so "ui basic ribbon" is 1 pixel wider than "ui ribbon" in all directions:
- subtract 1px from padding on all sides of "ui ribbon basic", so it is same size as "ui ribbon"
- adjust position of "ui.basic.ribbon.label:after"

"ui ribbon"
default

"ui basic ribbon" before:
basic

"ui basic ribbon" after:
Screen Shot 2019-05-01 at 14 14 39

Closes #708

## Testcase
https://jsfiddle.net/eruc460w/

## Screenshot (when possible)
left:
left

right:
right

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -255,7 +255,7 @@ a.ui.label {
   border-color: inherit;
   border-width: @basicBorderWidth 0 0 @basicBorderWidth;
   border-style: inherit;
-  right: e(%("calc(100%% + %d)", @basicBorderWidth));
+  right: @basicBorderFullWidthOffset;
 }
 .ui.basic.tag.labels .label:after,
 .ui.basic.tag.label:after {
@@ -384,6 +384,27 @@ a.ui.label {
   left: @rightRibbonOffset;
   padding-left: @ribbonTriangleSize;
   padding-right: @ribbonDistance;
+}
+.ui.basic.ribbon.label {
+  padding-top: @basicRibbonOffset;
+  padding-bottom: @basicRibbonOffset;
+}
+.ui.basic.ribbon.label:not([class*="right ribbon"]) {
+  padding-left: @basicRibbonTriangleSizeOffset;
+  padding-right: @basicRibbonTriangleSize;
+}
+.ui.basic[class*="right ribbon"].label {
+  padding-left: @basicRibbonTriangleSize;
+  padding-right: @basicRibbonTriangleSizeOffset;
+}
+.ui.basic.ribbon.label::after {
+  top: @basicBorderFullWidthOffset;
+}
+.ui.basic.ribbon.label:not([class*="right ribbon"])::after {
+  left: @basicBorderWidthOffset;
+}
+.ui.basic[class*="right ribbon"].label::after {
+  right: @basicBorderWidthOffset;
 }
 
 /* Right Ribbon */
--- a/src/themes/default/elements/label.variables
+++ b/src/themes/default/elements/label.variables
@@ -105,6 +105,8 @@
 /* Basic */
 @basicBackground: none @white;
 @basicBorderWidth: 1px;
+@basicBorderWidthOffset: -@basicBorderWidth;
+@basicBorderFullWidthOffset: e(%("calc(100%% + %d)", @basicBorderWidth));
 @basicBorder: @basicBorderWidth solid @borderColor;
 @basicColor: @textColor;
 @basicBoxShadow: none;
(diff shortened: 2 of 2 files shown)


dutrieux:
Well done ! 👍

src/definitions/elements/label.less
```diff
@@ -385,6 +385,26 @@ a.ui.label {
   padding-left: @ribbonTriangleSize;
   padding-right: @ribbonDistance;
 }
+.ui.basic.ribbon.label:not([class*="right ribbon"]) {
+  padding-top: calc(@verticalPadding - 1px);
```
lubber-de: Please replace `1px` by `@basicBorderWidth` everywhere

src/definitions/elements/label.less
```diff
@@ -385,6 +385,26 @@ a.ui.label {
...
+  padding-bottom: calc(@verticalPadding - 1px);
```
lubber-de: For saver use of LESS not calculating values in brackets at compilation time (put make sure put the calc straight into CSS), please also move every `calc` values into
expressions, so for example
```less
padding-bottom: calc(@verticalPadding - 1px);
```
becomes
```less
padding-bottom:  e(%("calc(%d - %d)", @verticalPadding, @basicBorderWidth));
```

lubber-de: And , to stay consistent with current codebase, even move that expression into a variable aswell as all the other ribbon settings already do:

```less
@basicRibbonOffset: e(%("calc(%d - %d)", @verticalPadding, @basicBorderWidth));
...
padding-bottom:  @basicRibbonOffset;
```

globophobe: Hopefully, better. It's my first time to use LESS. Thanks for your help so far.

src/themes/default/elements/label.variables
```diff
@@ -128,10 +128,14 @@
...
 /* Ribbon */
 @ribbonTriangleSize: 1.2em;
+@basicRibbonTriangleSize: e(%("calc(%d - %d)", @ribbonTriangleSize, @basicBorderWidth));
+@basicRibbonTriangleSizeOffset: e(%("calc(1rem + %d - %d)", @ribbonTriangleSize, @basicBorderWidth));
 @ribbonShadowColor: rgba(0, 0, 0, 0.15);
 
 @ribbonMargin: 1rem;
 @ribbonOffset: e(%("calc(%d - %d)", -@ribbonMargin, @ribbonTriangleSize));
+@basicRibbonOffset: e(%("calc(%d - %d)", @verticalPadding, @basicBorderWidth));
+@basicRibbonAfterOffset: e(%("-%d", @basicBorderWidth));
```
lubber-de: Expression is not necessary here, only when using "calc", so
```
@basicRibbonAfterOffset: -@basicBorderWidth;
```
is enough here.

src/definitions/elements/label.less
```diff
@@ -385,6 +385,26 @@ a.ui.label {
...
+.ui.basic.ribbon.label {
+  padding-top: @basicRibbonOffset;
+  padding-bottom: @basicRibbonOffset;
+}
+.ui.basic.ribbon.label:not([class*="right ribbon"]) {
+  padding-left: @basicRibbonTriangleSizeOffset;
+  padding-right: @basicRibbonTriangleSize;
+}
+.ui.basic.ribbon.label:not([class*="right ribbon"])::after {
+  top: e(%("calc(100%% + %d)", @basicBorderWidth));
```
lubber-de: please move this into a variable as well to stay consistent (even if only used once) thanks

globophobe: [label.less#L258](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/elements/label.less#L258) has the same expression as my proposed addition:

`@basicRibbonAfterTopOffset: e(%("calc(100%% + %d)", @basicBorderWidth));`

Could they be combined? Perhaps, as @basicLabelOffset or something...

lubber-de: Good catch! Yes, you can combine them as you suggested 👍

globophobe: Combined as @basicBorderFullWidthOffset

hammy2899:
@all-contributors please add @globophobe for code

PR #735, opened 2019-05-09
feat(progress): add indeterminate states

lubber-de:
## Description
This PR adds a proposed solution for an `indeterminate` state to the progress module.
It does not even need javascript and works with any existing FUI progress html code by just applying an additional class 😄 
To visually comply with the current FUI style relied on the `active` state animation style.
Every state is fully compatible to any size or color.

In addtion this PR adds an optional `centered` class to the inner progress text, to provide a functionality to show the percentage as static centered text regardless of the bar width.

I added the following `indeterminate` states :

`pulsating` (currently default, thus not explicitely defined as a class)
`filling`
`sliding`
`swinging`

I would be happy to hear feedback 🙂 

## Testcase
https://jsfiddle.net/nekc8v7m/8/

## Screenshot
indeterminate_progress

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4840

### DIFF
--- a/src/definitions/modules/progress.less
+++ b/src/definitions/modules/progress.less
@@ -487,6 +487,12 @@ each(@colors, {
   transform-origin: center;
   width: 100%;
 }
+.ui.slow.indeterminate.progress .bar::before {
+  animation-duration: @indeterminatePulseDurationSlow;
+}
+.ui.fast.indeterminate.progress .bar::before {
+  animation-duration: @indeterminatePulseDurationFast;
+}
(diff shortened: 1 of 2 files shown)


prudho:
This is nice ! What do you think about make animation time configurable ?

lubber-de:
@prudho Good idea. I added the speed variants `slow` and `fast` now.
See https://jsfiddle.net/65rz3u0m/
indeterminate_progress

PR #737, opened 2019-05-10
fix(button,table): move LESS functions out of LESS file and into variables

### LINKED ISSUE #736: Default theme will be loaded regardless of theme.config
# Help Wanted
What i'm trying to do is have the FUI primary color be bounded to a css var(), the main issue is the usage of pre-processing functions.
So what i did is to copy the site.variables of the default them and copied it to `/site/globals/site.variables`  and modified every primary related var to a css var(), like this:
`@linkColor           : var(--accent-link-color);
@primaryBackground   : var(--accent-bg-color);
`
I then went ahead and tried to compile and got the following:

`{ [Error: error evaluating function `darken`: color.toHSL is not a function in file /home/yahav/fomantic_themeColors/semantic/src/themes/default/globals/site.variables line no. 145]
  message: 'error evaluating function `darken`: color.toHSL is not a function in file /home/yahav/fomantic_themeColors/semantic/src/themes/default/globals/site.variables line no. 145',
  stack: undefined,
  type: 'Runtime',
  filename: '/home/yahav/fomantic_themeColors/semantic/src/themes/default/globals/site.variables',
  index: 3383,
  line: 145,
  column: 23,
  callLine: NaN,
  callExtract: undefined,
  extract: 
   [ '@linkUnderline       : none;',
     '@linkHoverColor      : darken(saturate(@linkColor, 20), 15, relative);',
     '@linkHoverUnderline  : @linkUnderline;' ],
  lineNumber: 145,
  fileName: '/home/yahav/fomantic_themeColors/semantic/src/themes/default/globals/site.variables',
  name: 'Error',
  plugin: 'gulp-less',
  showProperties: true,
  showStack: false,
  __safety: { toString: [Function: bound ] } }
`

so i created a new theme folder, basically copied the default theme and renamed it, i then copied the `/site/globals/site.variables` file to this new theme's site.variables and changed every occurrence of 'default' at theme.config to the new theme's name.
and yet, i still get the same error which means FUI is trying to compile the `themes/default/globals/site.variables` file although i have defined the new theme AND have a `site/globals/site.variables` file overriding it a well.

I can simply overwrite the default theme site.variables and it will work (i had this same setup when i was using SUI) but it won't be wise.
Any ideas?

lubber-de:
## Description
The `button.less` and `table.less` files had hardcoded LESS color math functions included (darken/lighten/saturate/desaturate/screen).
If somebody defines some color variables as CSS Variable strings ( like `var(--mycssvar)`) , any LESS Math functions did not work anymore and caused a LESS compiler error (because LESS always expects a valid color definition as argument)

To comply with the central color definition system we invented in 2.7.0 this PR now replaces all those functions by appropriate variables instead.

This way, if someone decides to change all/some color definitions to strings like mentioned above, the button code won't break anymore (as long as _every_ color related saturate/darken/lighten/desaturate...etc. was adjusted in the *.variables files)

## Closes
#736


Yahav:
that solved the table, next one is Item:

`{ [Error: error evaluating function `lighten`: color.toHSL is not a function in file /home/yahav/fomantic_themeColors/semantic/src/themes/default/views/item.variables line no. 169]
  message: 'error evaluating function `lighten`: color.toHSL is not a function in file /home/yahav/fomantic_themeColors/semantic/src/themes/default/views/item.variables line no. 169',
  stack: undefined,
  type: 'Runtime',
  filename: '/home/yahav/fomantic_themeColors/semantic/src/themes/default/views/item.variables',
  index: 3656,
  line: 169,
  column: 27,
  callLine: NaN,
  callExtract: undefined,
  extract: 
   [ '@invertedContentColor: @invertedTextColor;',
     '@invertedContentLinkColor: lighten(saturate(@linkColor, 30), 25, relative);',
     '@invertedContentLinkHoverColor: @linkColor;' ],
  lineNumber: 169,
  fileName: '/home/yahav/fomantic_themeColors/semantic/src/themes/default/views/item.variables',
  name: 'Error',
  plugin: 'gulp-less',
  showProperties: true,
  showStack: false,
  __safety: { toString: [Function: bound ] } }
`

lubber-de:
@Yahav 
> that solved the table, next one is Item:

No, this time the LESS function is part of a *.variable file (as all the other variables in `site.variables` as well) 
So in this case, as you modified the `@linkColor` variable in your theme, you also have to set the `@invertedContentLinkColor` to your own `var(--)` value, which you probably want anyway.
(remember to set it before changing the `@linkColor`)

PR #738, opened 2019-05-10
[Table] Move LESS Functions out of LESS file and into separat variables instead

Yahav:
## Description
https://github.com/fomantic/Fomantic-UI/issues/736
https://github.com/fomantic/Fomantic-UI/issues/736#issuecomment-491346764
https://github.com/fomantic/Fomantic-UI/pull/737

one of you is going to buy me a new space key..  :)


lubber-de:
I would like to decline this pr because I already made a similar fix to #737 while I was debugging, our error message we discussed 
I also named it "bright" for a more generic usage and made the code fit the current codebase.
However thanks for the effort

Yahav:
kk, thanks.

PR #740, opened 2019-05-12
Added Width per Device for tables

### LINKED ISSUE #222: [Label] Icons should have correct margin when placed after the text

awakenalive:
Copy/paste/update from definitions/collections/grids.less

## Description
I haven't really looked that hard, so I don't know if there was a reason for omitting width-per-device in table column sizes. This _should_ do that... I haven't tested this at all, I just assumed it would work fine based on grids.less.

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -834,6 +834,350 @@ each(@colors, {
   width: @sixteenWide;
 }
 
+/*----------------------
+    Width per Device
+-----------------------*/
+
+/* Mobile Sizing Combinations */
+@media only screen and (min-width: @mobileBreakpoint) and (max-width: @largestMobileScreen) {
+  .ui.table th[class*="one wide mobile"],
+  .ui.table td[class*="one wide mobile"] {
+    width: @oneWide !important;
+  }
+  .ui.table th[class*="two wide mobile"],
+  .ui.table td[class*="two wide mobile"] {
+    width: @twoWide !important;
+  }
+  .ui.table th[class*="three wide mobile"],
+  .ui.table td[class*="three wide mobile"] {
+    width: @threeWide !important;
+  }
+  .ui.table th[class*="four wide mobile"],
+  .ui.table td[class*="four wide mobile"] {
+    width: @fourWide !important;
+  }
+  .ui.table th[class*="five wide mobile"],
+  .ui.table td[class*="five wide mobile"] {
+    width: @fiveWide !important;
+  }
+  .ui.table th[class*="six wide mobile"],
+  .ui.table td[class*="six wide mobile"] {
+    width: @sixWide !important;
+  }
+  .ui.table th[class*="seven wide mobile"],
+  .ui.table td[class*="seven wide mobile"] {
+    width: @sevenWide !important;
+  }
+  .ui.table th[class*="eight wide mobile"],
+  .ui.table td[class*="eight wide mobile"] {
+    width: @eightWide !important;
+  }
+  .ui.table th[class*="nine wide mobile"],
+  .ui.table td[class*="nine wide mobile"] {
+    width: @nineWide !important;
+  }
+  .ui.table th[class*="ten wide mobile"],
+  .ui.table td[class*="ten wide mobile"] {
+    width: @tenWide !important;
+  }
+  .ui.table th[class*="eleven wide mobile"],
+  .ui.table td[class*="eleven wide mobile"] {
+    width: @elevenWide !important;
+  }
+  .ui.table th[class*="twelve wide mobile"],
+  .ui.table td[class*="twelve wide mobile"] {
+    width: @twelveWide !important;
+  }
+  .ui.table th[class*="thirteen wide mobile"],
+  .ui.table td[class*="thirteen wide mobile"] {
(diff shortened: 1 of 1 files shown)


src/definitions/collections/table.less
```diff
@@ -834,6 +834,350 @@ each(@colors, {
...
 
+/*----------------------
+    Width per Device
+-----------------------*/
+
+/* Mobile Sizing Combinations */
+@media only screen and (min-width: @mobileBreakpoint) and (max-width: @largestMobileScreen) {
+  .ui.table th[class*="one wide mobile"],
+  .ui.table td[class*="one wide mobile"] {
+    width: @oneWide !important;
```
exoego: Same `@oneWide` or similar used for all devices, which means width are same across devices.
I think it is required to define `@computerOneWide`, `@tabletEeightWide` and so on to realize responsive width per devices.

awakenalive: Based on what was used in grid.less, I'm not sure that's what is supposed to happen. The widths are percentages, and just give you the ability to have different column widths per device:

`two wide computer one wide mobile`

This would mean that on a "computer screen" the column would be two wide, but on a "mobile screen" the column would only be one wide.

lubber-de:
And: Yes, i checked `grid.less` which indeed has the same equal "senseless" logic by using the same variables for the different breakpoints. 
So what we might need are additional variables for different breakpoints/device widths together with a refactoring of table.
Do you think you can adjust the PR accordingly into that direction? 🙂

awakenalive:
Maybe I'm missing something here, but the these rules seem necessary and actually work for `grid.less`. The rules are surrounded by media queries, so the `@OneWide` or other rules only get applied if the media query matches. This is also the reason for including the `!important`. The css selector for `one wide` will match a rule.

Here's what we would end up with:
```
<td class="one wide mobile two wide computer">column</td>
```

`one wide` and `two wide` would both match the first set of css selectors, so since `one wide` comes first in the file, the styles associated with the rule would be applied. Additionally, if your screen size matches `(min-width: @mobileBreakpoint) and (max-width: @largestMobileScreen)` the below rule would also match:

```
.ui.table th[class*="one wide mobile"],
  .ui.table td[class*="one wide mobile"] {
    width: @oneWide !important;
 } 
```
and the style applied there would end up applying because of `!important`. However, if your screen size matches `(min-width: @computerBreakpoint)`, the below rule would also match:

```
  .ui.table th[class*="two wide computer"],
  .ui.table td[class*="two wide computer"] {
    width: @twoWide !important;
  }
```

and the style applied there would end up applying because of `!important`.

lubber-de:
@awakenalive Alright, got it.. 😃 Please forget about my previous comments  regarding the "same variable"...I thought about it again, and understood now. Thanks for clarification 👍 . I possible need more sleep, but it was really confusing me first.

Anyway, to  adopt the logic for table as well, there still is a bit more testing and some changes needed, i assume. I definately don't think we need the `!important` statements here. The specificity of the media query selcectors should be enough to override the default settings.
I also believe the `!important` could be removed from the `grid.less` media queries as well, but this needs testing.

awakenalive:
@lubber-de cool, point me in the right direction for other changes and tests and I'll make those happen. Do you want me to make the changes to `grid.less` here too?

lubber-de:
@awakenalive Would be really awesome if you could donate some time to investigate. 😉 
I prepared an adjusted semantic.css which basically reduced the whole usage of those many `zzzz wide mobile/tablet/computer/large screen/widescreen`

I think/hope it will turn out that all those

```less
.ui.grid > .row > [class*="seven wide mobile"].column,
.ui.grid > .column.row > [class*="seven wide mobile"].column,
.ui.grid > [class*="seven wide mobile"].column,
.ui.column.grid > [class*="seven wide mobile"].column {
   width: @sevenWide !important;
}

.ui.table th[class*="seven wide mobile"],
.ui.table td[class*="seven wide mobile"] {
   width: @sevenWide !important;
}
```
can be reduced to
```less
.ui.table [class*="seven wide mobile"],
.ui.grid [class*="seven wide mobile"] {
  width: @sevenWide;
}
```

or even (to cover grid/table at the same time, although that would mean this is not element-specific anymore)

```less
.ui [class*="seven wide mobile"] {
  width: @sevenWide;
}
```
The latter i prepare here https://cdn.jsdelivr.net/gh/lubber-de/fomantic-ui@bbdc199f49e64483f08e60ef8c0a81bcb959c369/dist/semantic.css
which you could use for heavy grid+table testing.

I also removed the appropriate `!important` statements for default `width` which were causing the necessarity for all the other `!important` statements inside the media queries. (i probably missed some or they need specificity adjustments (without using`!important`)

hammy2899:
Closing this since it seems stale, if you want to update the PR feel free to reopen.

PR #753, opened 2019-05-20
Update dropdown.js

### LINKED ISSUE #222: [Label] Icons should have correct margin when placed after the text

smartm0use:
Make dropdown compliant with A11Y adding title property.
Added disabled property.

## Closes
-

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -349,8 +349,7 @@ $.fn.dropdown = function(parameters) {
                   .addClass(className.search)
                   .prop({
                       'autocomplete': 'off',
-                      'title': title,
-                      'disabled': module.is.disabled()
+                      'title': title
                   })
                 .insertBefore($text)
               ;
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -48,6 +48,7 @@ $.fn.dropdown = function(parameters) {
           : $.extend({}, $.fn.dropdown.settings),
 
         className       = settings.className,
+        title           = settings.title,
         message         = settings.message,
         fields          = settings.fields,
         keys            = settings.keys,
@@ -345,11 +346,19 @@ $.fn.dropdown = function(parameters) {
             if( module.is.search() && !module.has.search() ) {
               module.verbose('Adding search input');
               $search = $('<input />')
-                .addClass(className.search)
-                .prop('autocomplete', 'off')
+                  .addClass(className.search)
+                  .prop({
+                      'autocomplete': 'off',
+                      'title': title ? title : 'search',
+                      'disabled': module.is.disabled()
+                  })
                 .insertBefore($text)
               ;
             }
+            if(module.is.search() && module.has.search() && title) {
+              module.verbose('Adding title attr to search input');
+              $search.prop("title", title);
+            }            
             if( module.is.multiple() && module.is.searchSelection() && !module.has.sizer()) {
               module.create.sizer();
             }
@@ -2891,7 +2900,7 @@ $.fn.dropdown = function(parameters) {
             }
             module.set.value(newValue, addedValue, addedText, $selectedItem);
             module.check.maxSelections();
-          },
+          }
         },
 
         remove: {
@@ -3854,6 +3863,8 @@ $.fn.dropdown.settings = {
   ignoreDiacritics       : false,      // match results also if they contain diacritics of the same base character (for example searching for "a" will also match "á" or "â" or "à", etc...)
   hideDividers           : false,      // Whether to hide any divider elements (specified in selector.divider) that are sibling to any items when searched (set to true will hide all dividers, set to 'empty' will hide them when they are not followed by a visible item)
 
+  title                  : null,       // input title attribute
+  
   placeholder            : 'auto',     // whether to convert blank <select> values to placeholder text
   preserveHTML           : true,       // preserve html when selecting value
   sortSelect             : false,      // sort selection on init
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -48,6 +48,7 @@ $.fn.dropdown = function(parameters) {
           : $.extend({}, $.fn.dropdown.settings),
 
         className       = settings.className,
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -345,11 +346,19 @@ $.fn.dropdown = function(parameters) {
             if( module.is.search() && !module.has.search() ) {
               module.verbose('Adding search input');
               $search = $('<input />')
-                .addClass(className.search)
-                .prop('autocomplete', 'off')
+                  .addClass(className.search)
+                  .prop({
+                      'autocomplete': 'off',
+                      'title': title ? title : 'search',
```
hammy2899: Instead of checking if title is set just put the default value in the settings instead of the setting being `null` by default.

y0hami (changes requested):

src/definitions/modules/dropdown.js
```diff
@@ -345,11 +346,19 @@ $.fn.dropdown = function(parameters) {
             if( module.is.search() && !module.has.search() ) {
               module.verbose('Adding search input');
               $search = $('<input />')
-                .addClass(className.search)
-                .prop('autocomplete', 'off')
+                  .addClass(className.search)
+                  .prop({
+                      'autocomplete': 'off',
+                      'title': title,
+                      'disabled': module.is.disabled()
```
lubber-de: Adding the `disabled` property to the search input at setup _once_ will break existing code which is possibly adding/removing the `disabled` class to the dropdown within the application at runtime without also removing/adding the disabled property for the search input.
I suggest to remove it here and add it as an enhancement in a separate PR, which will then be postponed for merge until 2.8.x. You should also provide a behavior `set disabled` and `remove disabled` which has to take care of the removement of the class as well as the removement of the property in that mentioned additional PR.

See breaking examples here http://jsfiddle.net/g5to4xks/

dd_search_disabled_property

lubber-de (changes requested):
Please remove the `disabled` property because it is not handled for dynamic changes. See my comment and testcase http://jsfiddle.net/g5to4xks/

smartm0use:
Done, I have removed `disabled` property. Thank you for your tests.

lubber-de (dismissed):
LGTM

lubber-de (commented):
Before merging this, we should vote for using `aria-label` instead of `title`
Because i just recognized the `title` property is visualized by browser default and is not the intended functionality what this PR tries to achieve.
image
https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-label

src/definitions/modules/dropdown.js
```diff
@@ -345,11 +346,18 @@ $.fn.dropdown = function(parameters) {
             if( module.is.search() && !module.has.search() ) {
               module.verbose('Adding search input');
               $search = $('<input />')
-                .addClass(className.search)
-                .prop('autocomplete', 'off')
+                  .addClass(className.search)
+                  .prop({
+                      'autocomplete': 'off',
+                      'title': title
```
lubber-de: As i guess the visual indicator using default browser label is too annoying, we should use the aria-namespace and change the `title` property to `aria-label`
image

src/definitions/modules/dropdown.js
```diff
@@ -345,11 +346,18 @@ $.fn.dropdown = function(parameters) {
...
+                  })
                 .insertBefore($text)
               ;
             }
+            if(module.is.search() && module.has.search() && title) {
+              module.verbose('Adding title attr to search input');
+              $search.prop("title", title);
```
lubber-de: Same here, please change to `aria-label` instead

smartm0use:
Yes, you are right and I agree!

lubber-de (changes requested):
As mentioned the visual indicator could be considered a breaking change, so we should use the aria namespace (`aria-label` in this case), which should be supported by all current screen readers.

Anyway, to make the whole dropdown A11Y compliant, there is much to be done, like adding `role="dropdown"` or `role="option"` to each dropdown entry. This could be covered in a separate PR (which should affect all modules then)

smartm0use:
Yes, I committed again.

src/definitions/modules/dropdown.js
```diff
@@ -3854,6 +3862,8 @@ $.fn.dropdown.settings = {
   ignoreDiacritics       : false,      // match results also if they contain diacritics of the same base character (for example searching for "a" will also match "á" or "â" or "à", etc...)
   hideDividers           : false,      // Whether to hide any divider elements (specified in selector.divider) that are sibling to any items when searched (set to true will hide all dividers, set to 'empty' will hide them when they are not followed by a visible item)
 
+  title                  : 'search',       // input title attribute
```
lubber-de: Sorry for being a grinch again :grin:, but if the think about the A11Y scope a bit further and want to avoid a breaking change in the future, we should name this `aria.searchLabel`, because `title` (better `aria.label `) should be reserved for the whole dropdown element. What about a more future proof setting:
```javascript
aria:{
  title:'dropdown', 
  searchTitle:'search' 
} 
```

lubber-de (changes requested):
title should be reserved for the whole component, see my suggested settings

src/definitions/modules/dropdown.js
lubber-de: Did the typo myself.. Of course the settings should read label instead of title

hammy2899:
@lubber-de Agreed, the reason I haven't reviewed this yet is because there is a lot of a11y attributes which need to be added and especially for the dropdown.

a11y has been an issue for SUI/FUI since v1 and it comes up quite a lot and its a huge task.

hammy2899:
Closing this PR since this is a large task then just this PR.

PR #758, opened 2019-05-21
[Progress] Fix wrong error on validating sum total on multiple progress 

### LINKED ISSUE #757: [Progress] Multiple progress bar is broken although valid total/value have been set.
# Bug Report

## Steps to reproduce
1. Set `total: 1122` and `value: [325, 111, 74, 612]` in multiple progress.

## Expected result
The multiple progress bar is not broken with no console error.

## Actual result

The multiple progress bar is broken like this
スクリーンショット 2019-05-21 12 13 13
with a console error.

スクリーンショット 2019-05-21 12 26 46

## Testcase
https://jsfiddle.net/d8t69y2s/

## Version
2.7.5

exoego:
## Description

1. Validation on sum of percents now is performed before rounding, as proposed by @ryamaguchi0220  in #759.

2.  Add `module.helper.derivePrecision` that calculates small-enough precision for multiple progress.
This is handy for users since they do not need to take care about precision for most cases.
If `precision` option is explicitly specified, it is used.

## Testcase
https://jsfiddle.net/x1ptzvs7/

## Screenshot (when possible)
Multiple progress bars are shown for the data in #757
image

## Closes
#757

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -74,6 +74,41 @@ $.fn.progress = function(parameters) {
               return left + right;
             }, 0) : 0;
           },
+          /**
+           * Derive precision for multiple progress with total and values.
+           *
+           * This helper dervices a precision that is sufficiently small to show minimum value of multiple progress.
+           *
+           * Example1
+           * - total: 1122
+           * - values: [325, 111, 74, 612]
+           * - min ratio: 74/1122 = 0.0659...
+           * - required precision:  0.01
+           *
+           * Example2
+           * - total: 10541
+           * - values: [3235, 1111, 74, 6121]
+           * - min ratio: 74/10541 = 0.0070...
+           * - required precision:   0.001
+           *
+           * @param min A minimum value within multiple values
+           * @param total A total amount of multiple values
+           * @returns {number} A precison. Could be 1, 0.1, 0.1, ... 1e-10.
+           */
+          derivePrecision: function(min, total) {
+            var precisionPower = 0
+            var precision = 1;
+            var ratio = min / total;
+            while (precisionPower > -10) {
+              ratio = ratio / precision;
+              if (ratio > 1) {
+                break;
+              }
+              precisionPower -= 1;
+              precision = Math.pow(10, precisionPower);
+            }
+            return precision;
+          },
           forceArray: function (element) {
             return Array.isArray(element)
               ? element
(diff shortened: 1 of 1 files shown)


lubber-de:
Alright, my suggestion of using 10 as always default precision could lead into never reaching 100% at all.. So your approach is of course better 🙂

exoego:
> Calculation seems right, but bar color/width is messed up

All green is due to `autoSuccess: true (default)`.
If `autoSuccess: false`, bars show each own color even when 100% reached.
Width seems fine for me if `autoSuccess: false`.


I took changes from #759 into this PR.

lubber-de:
@exoego In case you missed my comment for "autoSuccess" , here is a kind reminder... Would like to get this PR approved for the next release :wink:

exoego:
Oops.
Fixed in https://github.com/fomantic/Fomantic-UI/pull/758/commits/1497b6400cacab4fb91919c77509c9e95b25de11

ryamaguchi0220:
@exoego @lubber-de 

Oops. 
In some case, the progress bar is still broken.
Maybe, we should validate sum of the raw value instead of sum of the percents.

The case. 
https://jsfiddle.net/La8twh3u/

スクリーンショット 2019-05-30 11 30 55

スクリーンショット 2019-05-30 11 43 37

exoego:
>  I'm not sure but, I think we should validate sum of the raw value instead of sum of the percents.

True.
Done in https://github.com/fomantic/Fomantic-UI/pull/758/commits/28a29677777cb77d5d904f099cd2d188c70bfa45

src/definitions/modules/progress.js
```diff
@@ -478,34 +503,54 @@ $.fn.progress = function(parameters) {
...
-            });
-            module.percent = percents;
-            if( !module.has.total() ) {
-              module.value = percents.map(function(percent){
-                return (settings.precision > 0)
-                  ? Math.round( (percent / 100) * module.total * (10 * settings.precision)) / (10 * settings.precision)
-                  : Math.round( (percent / 100) * module.total * 10) / 10
-                ;
+            var hasTotal = module.has.total();
+            var totalPecent = module.helper.sum(percents);
```
ryamaguchi0220:
```suggestion
            var totalPercent = module.helper.sum(percents);
```

lubber-de:
Docs corrected by https://github.com/fomantic/Fomantic-UI-Docs/pull/134

PR #759, opened 2019-05-21
[Progress] Fix to validate the sum of the percents before rounding

### LINKED ISSUE #757: [Progress] Multiple progress bar is broken although valid total/value have been set.

ryamaguchi0220:
## Description
In addition to #758 , I think that we should validate the sum of the percents before rounding instead of after rounding, because, as the error message indicates, the validation is for `total` and `value` as entered by the user.
スクリーンショット 2019-05-21 12 26 46

## Testcase
https://jsfiddle.net/e0hvd7xo/
https://jsfiddle.net/562t0f3z/

## Closes
#757

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -411,52 +411,43 @@ $.fn.progress = function(parameters) {
           barWidth: function(values) {
             module.debug("set bar width with ", values);
             values = module.helper.forceArray(values);
-            var total = module.helper.sum(values);
-            if(total > 100) {
-              module.error(error.tooHigh, total);
-            }
-            else if (total < 0) {
-              module.error(error.tooLow, total);
-            }
-            else {
-              var firstNonZeroIndex = -1;
-              var lastNonZeroIndex = -1;
-              var valuesSum = module.helper.sum(values);
-              var barCounts = $bars.length;
-              var isMultiple = barCounts > 1;
-              var percents = values.map(function(value, index) {
-                var allZero = (index === barCounts - 1 && valuesSum === 0);
-                var $bar = $($bars[index]);
-                if (value === 0 && isMultiple && !allZero) {
-                  $bar.css('display', 'none');
-                } else {
-                  if (isMultiple && allZero) {
-                    $bar.css('background', 'transparent');
-                  }
-                  if (firstNonZeroIndex == -1) {
-                    firstNonZeroIndex = index;
-                  }
-                  lastNonZeroIndex = index;
-                  $bar.css({
-                    display: 'block',
-                    width: value + '%'
-                  });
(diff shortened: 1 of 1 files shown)


lubber-de:
The problem is: The rounding issues are a general Javascript problem. If we don't round them before comparing (it's actually correction instead of rounding to avoid values like 100.000000003 because of multiplication and division with the same value at the same time) the comparison  probably always fails

lubber-de:
@ryamaguchi0220  Your PR does not fix the example of your given jsfiddle when the sum is definately exceeding the total value (325+1, 111, 74, 612 =1123, but total given is 1122) , which of course gives a percentage > 100% (regardless of the rounding)

#### Testcases with sum(values) > given total
All fixes break because the sum of all values definaltey exceed the given total sum regardless of any rounding order or precision calculation 
- Your example which uses your fix (breaks)
https://jsfiddle.net/kmbc7fog/

- Your example which uses @exoego 's fix (breaks)
https://jsfiddle.net/kmbc7fog/1/

- Your example which uses both fixes altogether (still breaks)
https://jsfiddle.net/kmbc7fog/2/

#### BUT
... what your PR indeed does fix is some behavior with the precision setting of 10. This currently still breaks breaks in #758 

- Your example which uses your fix (works!)
https://jsfiddle.net/kmbc7fog/5/

- Your example which uses @exoego 's fix (breaks)
https://jsfiddle.net/kmbc7fog/4/

- Your example which uses both fixes altogether (works)
https://jsfiddle.net/kmbc7fog/6/


@exoego You both should join forces for this precision issue 😉

ryamaguchi0220:
@lubber-de Thank you for review!

> Your PR does not fix the example of your given jsfiddle

What I'd like to test in the jsfiddle(https://jsfiddle.net/562t0f3z/) is that if the sum of the `value` exceeds the `total`, the error will occur.
Because I think it is right behavior that the error occurs when the user's input is incorrect.
So it is my expectation that the error occurs in the jsfiddle.
What do you think?

exoego:
Thanks for proposing this changes.
Closing since the changes are merged into #758

lubber-de:
> Because I think it is right behavior that the error occurs when the user's input is invalid.
> So it is my expectation.
> What do you think?

Yes, definately. The console error in that case is correct and does not need a change

PR #761, opened 2019-05-22
build(composer): fix typo, add web property & changed domain to FUI

lubber-de:
## Description
This PR corrects the composer.json files for correct schema usage and fomantic-domain as well as a typo 

## Closes
https://github.com/Semantic-Org/Semantic-UI/pull/6815

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
   },
   "keywords" : [
     "fomantic",
-    "fomaniic-ui",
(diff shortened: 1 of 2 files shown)


lubber-de:
Some people also install JS packages in their PHP projects via composer and it's available via https://packagist.org/packages/fomantic/ui
It's also there since the very beginning (2013), so removing it could lead into a breaking change for some people

PR #764, opened 2019-05-23
fix(icon): don't import default theme overrides

### LINKED ISSUE #763: [Icon] Duplicate icon definition
# Bug Report
Since 2.7.5 (2bfe9a5c78a0434ce0a6eb1425d6b31ad1a8177e) the default theme overrides file is always included. Because the icon classes are in the overrides file, the icon classes are now duplicated.

Additionally it is no longer possible to restrict the icons to a subset, because the default icon.overrides (with all fa icons) is loaded.

## Expected result
Icon classes are defined once

## Actual result
Icon classes are defined multiple times

## Version
2.7.5

ColinFrick:
## Description
Don't automatically import the default theme overrides file for icons.

## Closes
#763

### DIFF
--- a/src/theme.less
+++ b/src/theme.less
@@ -66,8 +66,10 @@
      Overrides
 -------------------*/
 
-.loadUIOverrides() {
-  @import (optional) "@{themesFolder}/default/@{type}s/@{element}.overrides";
+.loadUIOverrides(@importDefault: true) {
+  & when (@importDefault = true) {
+    @import (optional) "@{themesFolder}/default/@{type}s/@{element}.overrides";
+  }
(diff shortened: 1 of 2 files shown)


src/theme.less
```diff
@@ -66,8 +66,10 @@
      Overrides
 -------------------*/
 
-.loadUIOverrides() {
-  @import (optional) "@{themesFolder}/default/@{type}s/@{element}.overrides";
+.loadUIOverrides(@importDefault: true) {
+  & when (@importDefault = true) {
```
lubber-de: Couldn't we just ask if theme is default? So we are prepared for any other similar issue?

` & when not  (@theme = 'default') `

PR #773, opened 2019-05-25
docs(security): add security policy

hammy2899:
## Description

Added a security policy showing which versions we support and how people can get in touch with us to report vulnerabilities or their concerns.

(merging into master so people can see it before release)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -7,8 +7,11 @@
 | Version   | Supported          |
 | --------- | ------------------ |
 | 2.7.x     | :white_check_mark: |
-| > 2.x.x     | :lock:             |
+| > 2.x.x   | :lock:             |
 | 1.x.x     | :x:                |
+
+> All security fixes will be added the next release. Unless the threat is major the fix will be released with in the normal release
+> schedule.
 
 ## Reporting a Vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+:white_check_mark: = Supported   :lock: = Security updates   :x: = Not supported
+
+| Version   | Supported          |
+| --------- | ------------------ |
+| 2.7.x     | :white_check_mark: |
+| > 2.x.x     | :lock:             |
+| 1.x.x     | :x:                |
+
+## Reporting a Vulnerability
+
+We take security very seriously with thousands of users using our library we need the communities help to spot security
+vulnerabilities.
+
+If you think you have found a vulnerability or have any concerns about our security please feel free to reach out to us using
+the details below.
+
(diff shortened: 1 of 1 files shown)


hammy2899:
@exoego Good shout 👍

SECURITY.md
```diff
@@ -0,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+:white_check_mark: = Supported   :lock: = Security updates   :x: = Not supported
+
+| Version   | Supported          |
+| --------- | ------------------ |
+| 2.7.x     | :white_check_mark: |
+| > 2.x.x     | :lock:             |
```
lubber-de: I think this should be >= 2.4 < 2.7.0, because we still only support the FUI codebase. Everything below 2.4 is SUI
Also: As we fixed XSS vulnerabilities for dropdown/search since 2.7.0, wouldn't that mean, we have to provide separate versions for all < 2.7 having those features merged (like 2.5.0.1, 2.5.1.1, 2.6.0.1 ...?
I am a bit confused about what we promising here by this table, as currently every fix we do results in a new minor version patch like 2.7.6, rather than 2.5.1.1 🤔

hammy2899: The table basically says we don't support v1 but we support all of v2 for security fixes and we only support 2.7 for actual releases.

If we didn't support all of v2 for security releases that's basically saying we don't fix any issues we haven't created which isn't true. So if an XSS issue was raised about a feature added in 2.1.0 even though we (FUI) didn't make that release we still need to fix that security issue.

lubber-de: Yes, but it should be somehow clear, that such fixes will be implemented in a new minor version, rather than patching it into the codebase from the original version.
So, if someone still uses 2.5.0 and finds a security issue, we will of course fix it, but the fix will be in 2.7.6 (from current POV) rather than taking the codebase from 2.5.0, patching it there and releasing a separate 2.5.0.1.

hammy2899: I think it would be somewhat obvious that it would be a new patch release since FUI/SUI both us semver (kinda 😄). If you think otherwise I will add a note.

lubber-de: For _me_ its obvious , too 😄. 
I am just thinking about all those people still using SUI 2.2.13 for example (many people havent even upgraded to SUI 2.4.x because of the modal scrolling bug which is still present there) without the desire to upgrade to FUI. For my understanding the above stated "promise" could be understood as "we will backport the fix into the old versions, too" resulting in a 2.2.13.1 (which in some cases could be a lot of effort or even impossible because of the different codebase)...

hammy2899: Yeah its basically impossible because of npm releases.

I will add a note in a sec.

hammy2899: Added

PR #781, opened 2019-05-28
feat(*): add dynamic LESS compilation of component variations

lubber-de:
## Description
This PR is a proposal for a solution to allow users to define which variations of an element will be compiled .
It was often requested to have a possibility to remove certain variations of an element from of the compiled CSS File.

I invented a central file `variation.variables` which is loaded into the default site.variables to enable all variations by default. If now somebody wants to remove for example the `tertiary` variant of the `button` element one just has to add  `@variationButtonTertiary: false` to their custom themes `site.variables`

#### Side effect improvements
-  Sizing logic has been unified, so `divider`, `reveal` and `card` also have size options now
- For the same reason, `header` and `statistic` also received missing size options (mini,big,massive)
- Removed unnecessary !important for `statistic` size options


lubber-de:
Added support for adjustable sizes, so we only have to define all desired sizes in a list
```less
@variationButtonSizes: mini, tiny, small, large, big, huge, massive;
```
This would also allow someone to add even more/custom desired size definition names at a central place as long as the appropriate font-size variables will be also declared in a custom theme's site.variables:

```less
@thin: 0.1em;
@fat: 2em;
@ultra: 10em;
@unbelievable: 15em;
@variationButtonSizes: mini, small, thin, fat, ultra, unbelievable;
```

prudho:
This is something I thought about too, and I think it's dooable. It will provide a great customization level and will allow a better final framework size.

hammy2899:
@lubber-de anything happening with this?

lubber-de:
Nobody, beside @prudho replied so far.. But that also means nobody complained about it 😊
As I also tagged this to 2.8.x I am gonna continue to work on this now as 2.7.7 is released and the new codebase. Shouldn't take too long though

lubber-de:
Finished the PR now. Don't get scared about that many commits, i did a commit for every component. That's why. But this is perhaps making your review easier.

hammy2899:
@lubber-de your branch has conflicts now btw 😉

lubber-de:
@hammy2899 Merged develop and resolved the conflicts 🙂

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/377

PR #785, opened 2019-05-29
fix(dropdown): check for existing observer before accessing it

### LINKED ISSUE #784: [Dropdown] Cannot read property 'observe' of undefined
# Bug Report

## Steps to reproduce
1. Create search dropdown
2. Call `dropdown` on element with values and `allowAdditions: true`
3. Be sad when it doesn't work

## Expected result
Should work all happy

## Actual result
Throws error and dropdown is broken

## Testcase
https://jsfiddle.net/6t70ce54/

## Screenshot (when possible)
image

## Version
2.7.5

## Notes
Was initially reported at Semantic-Org/Semantic-UI/issues/5018, but the stale bot closed it. @nosheenzazafd had a solution as well:

> Looking at the source code, I noticed that every time selectObserver is dereferenced, it is guarded with if(selectObserver), except for the statement at line [165](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/dropdown.js#L165). I believe this statement also needs to be guarded with the same check. I did this and it fixed my problem. If needed I can provide more details to reproduce this error, however it seems to be a logic bug in the code.

lubber-de:
## Description
When a search dropdown is initialized out of a `select` tag with given option values and one of those is set as `allowAddition:true` is set, it will trigger a observer at initialization time, but the observer is not yet initialized. This leads into a TypeError.
This PR adds a check for existing observer object before trying to access it (as it is implemented everywhere else in the code)

### Broken 
https://jsfiddle.net/6t70ce54/

### Fixed
https://jsfiddle.net/7rsge546/

## Closes
#784

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -181,7 +181,7 @@ $.fn.dropdown = function(parameters) {
         },
         observe: {
           select: function() {
-            if(module.has.input()) {
+            if(module.has.input() && selectObserver) {
               selectObserver.observe($module[0], {
                 childList : true,
                 subtree   : true
@@ -189,7 +189,7 @@ $.fn.dropdown = function(parameters) {
             }
           },
           menu: function() {
-            if(module.has.menu()) {
+            if(module.has.menu() && menuObserver) {
               menuObserver.observe($menu[0], {
                 childList : true,
                 subtree   : true


emilioHayes:
I got this problem, and i see this has been happening since 4-3 years,
¿can somebody tell me how is possible semantic-ui.com dont make this simple change?

lubber-de:
@emilioHayes You have to ask the maintainer of SUI to get an answer. It's all about constant maintenance and involving more people into open source ;) The related issue in SUI was closed by a bot after a certain time nobody was taking care of the issue there 🙄

PR #800, opened 2019-06-07
Fix content margin after two attached labels

bborn:
## Description
Non-attached content coming after two top attached labels should have increased margin. Currently this is done only for the case where there is one top attached label. In my use case, I have a card with a top right and top left attached label.

## Testcase
~https://jsfiddle.net/jkb8s3Ly/1/~
https://jsfiddle.net/v9dgfn5y/1/


src/definitions/elements/label.less
```diff
@@ -160,6 +160,7 @@ a.ui.label {
 
 /* Padding on next content after a label */
 .ui.top.attached.label:first-child + :not(.attached),
+.ui.top.attached.label:nth-of-type(2) + :not(.attached),
```
lubber-de: I suggest to change the whole to

```css
  .ui.top.attached.label ~ :not(.attached) {
    margin-top: 2rem !important;
  }
  .ui.bottom.attached.label ~ :not(.attached) {
    margin-top:0;
    margin-bottom: 2rem !important;
  } 
```

That would also cover `.ui-bottom.attached.label` left and right then. 😉 

See your adjusted jsfiddle here https://jsfiddle.net/pqny178m/
 
image

lubber-de: The following is even better, otherwise any element will get the margin:

```css
  .ui.top.attached.label ~ :first-of-type:not(.attached) {
    margin-top: 2rem !important;
  }
  .ui.bottom.attached.label ~ :last-child:not(.attached) {
    margin-top:0;
    margin-bottom: 2rem !important;
  }
```
See https://jsfiddle.net/pqny178m/1/

bborn: WFM.

lubber-de: Unfortunately the provided (for you working) change is not working anymore when an additional `bottom attached` label is added.
See https://jsfiddle.net/v9dgfn5y/

image
My suggestion instead works for this case also https://jsfiddle.net/v9dgfn5y/1/
image

I appreciate the time you spent improving FUI. However, if you want your PR to get merged, you should consider a consistent solution which should work for everybody and not only for yourself  :wink:

bborn: Hey - sorry, that's not what I meant - I was under the impression that you could apply your change to my PR (as a maintainer) - and that's fine with me. If you'd prefer me to update the PR, I'm happy to do that too.

lubber-de: 😄 Alright, then sorry for the misunderstanding. Yes, we usually prefer the authors to update their PRs on their own. 🙂 They of course should always agree to and/or discuss reviewers suggestions, because our suggestions might be wrong, so i advise to test reviewers suggestions also critically before blindly apply the changes. There might be situations i also have overseen

.github/FUNDING.yml
```diff
@@ -0,0 +1,3 @@
+# These are supported funding model platforms
+
+github: [hammy2899]
```
exoego: Changes LGTM, but please rebase your branch to `develop` instead of `master` to avoid unnecessary commits (this file).
We only merge PR's into the develop branch.
Thank you 🙂

lubber-de:
@bborn I updated your PR branch to the latest changes so there are no unnecessary commits anymore
@hammy2899 We all approved, i think we can remove the "on hold" label and merge for v2.7.7 ?

hammy2899:
@all-contributors please add @bborn for code

PR #824, opened 2019-06-19
feat(checkbox): add toggle on ENTER-key option

### LINKED ISSUE #819: Checkbox changes state on ENTER key
Current behavior: Checkboxes change their status, when they are focused and the ENTER-Key is pressed.

Expected behavior: They should only react to space key as native checkboxes do.

Problem: ENTER-Key is used to commit forms.  Now, when you commit a form with ENTER-Key, the value of the checkbox changes unintentionally.

EDIT: Only happens, when checkbox gets initialized via Javascript

lubber-de:
## Description
The usual behavior of a standard (non FUI / html default) checkbox is to also get toggled by the SPACE key, but not by the ENTER key, when the checkbox is focused. 
But if a checkbox is initialized via the checkbox module of FUI it also allows to be toggled via the ENTER key by default.
This could lead into unwanted form submissions.
This PR adds a new setting `enableEnterKey` (default `true` to stay backward compatible), so one can make the FUI behavior match to the html default behavior of a checkbox

## Testcase
https://jsfiddle.net/6o8pvfe4/

## Closes
#819

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -244,7 +244,7 @@ $.fn.checkbox = function(parameters) {
               $input.blur();
               shortcutPressed = true;
             }
-            else if(!event.ctrlKey && ( key == keyCode.space || key == keyCode.enter) ) {
+            else if(!event.ctrlKey && ( key == keyCode.space || (key == keyCode.enter && settings.enableEnterKey)) ) {
               module.verbose('Enter/space key pressed, toggling checkbox');
               module.toggle();
               shortcutPressed = true;
@@ -829,6 +829,7 @@ $.fn.checkbox.settings = {
   // delegated event context
   uncheckable         : 'auto',
   fireOnInit          : false,
+  enableEnterKey      : true,
 
   onChange            : function(){},
 


src/definitions/modules/checkbox.js
```diff
@@ -829,6 +829,7 @@ $.fn.checkbox.settings = {
   // delegated event context
   uncheckable         : 'auto',
   fireOnInit          : false,
+  enableEnterKey      : true,
```
exoego: So this change makes a standard (non FUI / html default) checkbox is also triggered by ENTER Key, which is not triggered right now? 

If so, it sounds breaking change.
I suggest to enable this option only for non-standard checkbox, something like this.
```javascript
enableEnterKey : $module.is("input:not([type=checkbox])")
```

lubber-de: The non Fui / standard html checkbox is unaffected by this change. Just remove the js code. See https://jsfiddle.net/7kqL485z/

lubber-de: Just in case, you are still confused:
> So this change makes a standard (non FUI / html default) checkbox is also triggered by ENTER Key, which is not triggered right now?

No, it's the over way around: html default is to _not_ support ENTER key and this PR makes it possible to also achieve the same behavior for FUI initialized checkboxes as well.
Sorry, if that was not clear enough from the description

Before this change the ENTER key was always supported for FUI initialized checkboxes, that's why the default for `enableEnterKey` is set to `true` to avoid a breaking change.

PR #860, opened 2019-07-08
Tell RTLCSS to ignore flag icon background-positions. Fix #859

### LINKED ISSUE #859: RTLCSS breaks flag elements
# Bug Report

RTLCSS which creates the Right-To-Left language support versions (semantic.rtl.css) incorrectly converts `background-position: 0 ...` to `background-position: 100% ...`. Which breaks the flags.

This is expected behaviour in RTLCSS, see:

https://github.com/MohammadYounes/rtlcss/issues/109

And the correct solution is to tell RTLCSS to ignore that section.

## Steps to reproduce
1. Include the semantic.rtl.css version.
2. Use a `<i class="fr flag"></i>` or any other flag which has a `background-position:0`

## Expected result

The flag in the background of an i tag.

## Actual result

Empty space where the flag should be.

## Testcase
https://jsfiddle.net/1f4qbLa7/
(Although as the RTL version is not available on CDNJS, I copied one of the broken declarations into the CSS file for you to see the problem.

## Version
2.x -> All including b3501f9 at least.

danthedeckie:
RTLCSS which creates the Right-To-Left language support versions (semantic.rtl.css) incorrectly converts background-position: 0 ... to background-position: 100% .... Which breaks the flags.

This is expected behaviour in RTLCSS, see:

MohammadYounes/rtlcss#109

And the correct solution is to tell RTLCSS to ignore that section.

## Description
Adds in rtl:ignore directives see: https://rtlcss.com/learn/usage-guide/control-directives/

## Testcase
https://jsfiddle.net/1f4qbLa7/

## Closes
#859

### DIFF
--- a/src/themes/default/elements/flag.overrides
+++ b/src/themes/default/elements/flag.overrides
@@ -3,6 +3,7 @@
 /*******************************
          Theme Overrides
 *******************************/
+/*rtl:begin:ignore*/
 
 i.flag.ad:before,
 i.flag.andorra:before {
@@ -989,3 +990,5 @@ i.flag.zw:before,
 i.flag.zimbabwe:before {
   background-position: -108px -390px;
 }
+
+/*rtl:end:ignore*/


hammy2899:
@all-contributors please add @danthedeckie for code

PR #863, opened 2019-07-10
Add all-contributors support

hammy2899:
## Description
Added config and relevant files to add support for [all-contributors](https://allcontributors.org/) 🎉

## Screenshot (when possible)
image


package-lock.json
```diff
@@ -4,6 +4,170 @@
   "lockfileVersion": 1,
```
exoego: Jenkins complains that `package-lock.json` has parse error.
Could you re-create `package-lock.json` ?

hammy2899: Fixed, was caused by the merge conflict I fixed.

PR #877, opened 2019-07-12
[CI] Verify build on pull requests

### LINKED ISSUE #876: [CI] Build on every PR
...so that we can be confident that PR does not break build.

Especially, recent overwhelming dependencies update by dependabot could leat to break build.

exoego:
## Description

Run `build` on every pull requests, so that we can verify that PRs does not break build before merge.

## Closes
#876

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,6 @@ cache: npm
 
 before_install:
   - sh preinstall.sh
+
+script:
+  - npm run build


.travis.yml
```diff
@@ -12,3 +12,7 @@ cache: npm
 
 before_install:
   - sh preinstall.sh
+
+script:
+  - npm run build
+  - npm test
```
hammy2899: This won't do anything so you should just remove it

exoego: Deleted commit

PR #884, opened 2019-07-18
[Table] Colored definition table shows overlapping border in top left cell

### LINKED ISSUE #883: [Table] definition and color can't be combined
# Bug Report

## Steps to reproduce
1. Create a `definition` `<table />`
2. Add `color` like `red`

## Expected result
No red border

## Actual result
Weird red border on top of the first row.

image

## Testcase
https://jsfiddle.net/3c4x7a2u/

## Version
2.7.6

lubber-de:
##  Description
A colored `definition table` was still showing the table color border on the top left cell (which is not used on definition tables by design)
I first thought it was enough to change the background, but IE/Edge still showed the border on zooming, thats why i still had to use and adjust the box-shadow

## Testcase
https://jsfiddle.net/gc15zfku/

### Before
883_evil

### Using only background (FYI)
883_bad

### Final Fix
883_good

## Closes
#883 
https://github.com/Semantic-Org/Semantic-UI-React/issues/3710

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -283,7 +283,7 @@
   background: @definitionHeaderBackground;
   font-weight: @definitionHeaderFontWeight;
   color: @definitionHeaderColor;
-  box-shadow: -@borderWidth -@borderWidth 0 @borderWidth @definitionPageBackground;
+  box-shadow: -@coloredBorderSizeCover -@coloredBorderSize 0 @coloredBorderSizeCover @definitionPageBackground;
 }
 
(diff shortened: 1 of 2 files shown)


exoego:
If some issues are found only in Firefox, it might be a bug in Firefox...
I suggest to report an issue in
https://bugzilla.mozilla.org/home
You can now login tonthr issue tracker with GitHub account.

lubber-de:
So glad, i now found a fix for firefox without the need for a `legacy` workaround, so i removed that one again by 0e6de97a5c8c03ca714f9bf3c33d76612f390e68 😃 
Adjusted jsfiddle https://jsfiddle.net/gc15zfku/

PR #895, opened 2019-07-23
[Search] Behavior 'get result:' Support numerical fields and make use of user defined searchFields

### LINKED ISSUE #897: [Search] numeric id not supported
# Bug Report
[Search] $('.ui.search').search("get result",value). Numeric id not supported

## Steps to reproduce
1. Sue numeric id. Ex: ``  const contentOK = [
    { id:1,title: 'United Arab Emirates' }, ``
2. Does not return any value: $('.ui.search').search("get result",1);

## Expected result
This should return  { id:1,title: 'United Arab Emirates' }

## Actual result
false

## Testcase
Fork https://jsfiddle.net/venturaEesc/q3p1b52t/13/

## Version
2.7.6

ventura-eesc:
## Description
When defining custom searchFields the $('.ui.search').search('get result', value)  do not use the user defined values, it uses the default ones.

This is because the default array is passed and the user setting are not used in line 616

``            searchFields = (searchFields !== undefined)
              ? searchFields
              : settings.searchFields
            ;
``
#### Broken
https://jsfiddle.net/venturaEesc/q3p1b52t/13/

#### Fixed
https://jsfiddle.net/2db9gf7u/

#### Broken
https://jsfiddle.net/venturaEesc/q3p1b52t/4/

#### Fixed
https://jsfiddle.net/2jx6bhd8/
## Closes
#897 
https://github.com/Semantic-Org/Semantic-UI/pull/6023

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -436,7 +436,6 @@ $.fn.search = function(parameters) {
           },
           result: function(value, results) {
             var
-              lookupFields = ['title', 'id'],
               result       = false
             ;
             value = (value !== undefined)
@@ -451,7 +450,7 @@ $.fn.search = function(parameters) {
               module.debug('Finding result that matches', value);
               $.each(results, function(index, category) {
                 if(Array.isArray(category.results)) {
-                  result = module.search.object(value, category.results, lookupFields)[0];
+                  result = module.search.object(value, category.results)[0];
                   // don't continue searching if a result is found
                   if(result) {
                     return false;
@@ -461,7 +460,7 @@ $.fn.search = function(parameters) {
             }
             else {
               module.debug('Finding result in results object', value);
-              result = module.search.object(value, results, lookupFields)[0];
+              result = module.search.object(value, results)[0];
             }
             return result || false;
           },
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -632,10 +632,15 @@ $.fn.search = function(parameters) {
             $.each(searchFields, function(index, field) {
               $.each(source, function(label, content) {
                 var
-                  fieldExists = (typeof content[field] == 'string')
+                  fieldExists = (typeof content[field] == 'string') || (typeof content[field] == 'number')
                 ;
                 if(fieldExists) {
-                  var text = module.remove.diacritics(content[field]);
+                  var text;
+                  if (typeof content[field] === 'string'){  
+                      text = module.remove.diacritics(content[field]);
+                  } else {
+                      text = content[field].toString(); 
+                  }
                   if( text.search(matchRegExp) !== -1) {
                     // content starts with value (first in results)
                     addResult(results, content);
@@ -1246,6 +1251,7 @@ $.fn.search.settings = {
 
   // fields to search
   searchFields   : [
+    'id',
     'title',
     'description'
   ],


lubber-de:
Thanks for the PR and basically LGTM...but..
The hardcoded `id` and `title` are expected when `get.value` is called internally.
Please read https://fomantic-ui.com/modules/search.html#retreiving-results where the id generation is explained.
However, to make it as flexible as you suggest we should also add `id` to the default values for settings.searchfields .
I still have to think about that, as the id-generation is part of the search core and expected to be available for each result. Curious about the other opinions

src/definitions/modules/search.js
```diff
@@ -436,7 +436,6 @@ $.fn.search = function(parameters) {
           },
           result: function(value, results) {
             var
-              lookupFields = ['title', 'id'],
```
lubber-de: If we only merge those with the existing searchfield setting here, then this could be all that's needed and the other lines dont have to be changed (and it's not a breaking change :)
```javascript
lookupFields = ['title', 'id'].concat(settings.searchFields).filter(function(item, i, ar){ return ar.indexOf(item) === i; });,
```

ventura-eesc:
Hi,

  Finally I've created a jsfiddle to show up the bug:

     https://jsfiddle.net/venturaEesc/wojfgt3k/5/

   If we don't use custom search fields searchFields the $searchKO.search("get result","Spain"); it works ok but if we change the custom searchFields it does not work. 

> The hardcoded id and title are expected when get.value is called internally.
Are you sure ? because when the input data json is customized the title field may not exists.

 Do not forget that in line 616 if if the passed searchFields argument to the function object() is undefined, the default values will be used. 
``
  searchFields   : [
    'title',
    'description'
  ],
``

Thank-you very much four your help :)

lubber-de:
Thanks for the jsfiddle 👍 

> > The hardcoded id and title are expected when get.value is called internally.
> 
> Are you sure ? because when the input data json is customized, the title field may not exists.

ok, right, but at least `id` is expected/needed when working with categories, because it is automatically generated for each result entry internally
https://github.com/fomantic/Fomantic-UI/blob/1fd94d2479f297c906bed557f0567d1d3f0a4d2c/src/definitions/modules/search.js#L849


The internal call where the lookupfields are currently expected is here
https://github.com/fomantic/Fomantic-UI/blob/1fd94d2479f297c906bed557f0567d1d3f0a4d2c/src/definitions/modules/search.js#L228

> 
> Do not forget that in line 616 if if the passed searchFields argument to the function object() is undefined, the default values will be used.

Yes, but that doesn't contain the `id` by default, that's why i suggested to just merge the array by 
```javascript
lookupFields = ['title', 'id'].concat(settings.searchFields).filter(function(item, i, ar){ return ar.indexOf(item) === i; }),
```

Another option is to add `id` (and also `title`, otherwise it is a breaking change) to the default `settings.searchFields`

Basically it is looking into all of the fields but at least one of them needs to match.

ventura-eesc:
Hi,

   Just debugging the code I found that in the line:

> result  = $result.data(metadata.result) || module.get.result(value, results) 

The first part of the expression try to find the information into the JQuery data structure, where we can found the values, in this case id, **titulo** (I changed the json) 
image

and just only if JQuery.data() fails, it will try to use the second part ``module.get.result(value, results)`` . I guess this should never happen because the values should always be in the jQuery data.

Options:
1. Removing The lookupFields should work, but I agree this is a breaking change and there will be a risk.
2. As you propose,  remove the lookupFields but adding id to the default searchFields like this:
``
  searchFields   : [
    'id',
    'title',
    'description'
  ],
``
3. And also your seconds option,  just removing the title from the lookupFields. Like this:
``
lookupFields = [ 'id'].concat(settings.searchFields).filter(function(item, i, ar){ return ar.indexOf(item) === i; }),
``
  This adds a small overhead because we lookup for the field "id" and may be it is useless.

Kind regards

hammy2899:
@all-contributors please add @ventura-eesc for code

PR #918, opened 2019-08-05
[Form] Check for existing calendar, dropdown or checkbox modules

### LINKED ISSUE #916: UI selectors matching when component not installed
I am not sure if this is exactly a bug, but would like to document this none the less. Especially as this only came up after the latest 2 releases.

## Steps to reproduce
1. Install fomantic ui without calendar
2. Use class "ui calendar" class for your custom calendar
3. Submit a form

## Expected result
Form is submitted without JS errors

## Actual result
JS errors will be present

## Error message
$calendar.calendar is not a function (coming from form validation)

## Version
2.7.7

# Why could it be considered a bug?
As mentioned, only came a problem now. 
Semantic UI does not include a calendar, so we use were free to use what we wanted and use the same namespace as that is what we started out with. As Fomantic UI calendar is included so on switch we just removed that component. However now it broke since it has a check for if it can find elements with the calendar selector. If it included without the component then it can find matching elements and cause problems - like we had. Our solution is to override the selector, but could be good if the component wasn't included that it wouldn't include the selector either (use something that wouldn't match anything). Could be useful to others as well if you switch from SUI to FUI and have your own calendar.

lubber-de:
## Description
In case somebody does not use the module functionality of calendar, dropdown or checkbox the form validation should check for the existing FUI JS module before trying to call any methods on it to prevent JS errors.
The current behavior is only checking for matching selectors but should also check if the relevant FUI module exists.

## Closes
#916


lubber-de:
> Mmmh, maybe we should display an error when we detect an non existent module, to warn the user
Implemented by 2e33b54ae1c08bc0ec6532c93de7ee3cd7023c69 🙂

src/definitions/behaviors/form.js
```diff
@@ -795,6 +795,16 @@ $.fn.form = function(parameters) {
 
         },
 
+        can: {
+            useElement: function(element){
+               if ($.fn[element] !== undefined) {
+                   return true;
+               }
+               module.error(error.noElement.replace(/{element}/,element));
```
exoego: I think costly RegExp is not required here.

```suggestion
               module.error(error.noElement.replace("{element}",element));
```

lubber-de: Very good catch! Thx 👍

PR #920, opened 2019-08-05
[Dropdown] Existing 'required' attribut for <select> prevents form submission

### LINKED ISSUE #917: [Form] Form cannot submit if a form contains select dropdown with required attribute and leaves empty
# Bug Report
The form failed to submit when it contains the HTML select element with the "required" attribute which uses Fomantic dropdown and the select left empty.

## Steps to reproduce
1. Create a form including the select element with the attribute "required"
2. Add the options for the select element. Include also for the empty option.
3. Invoke the dropdown plugin to the select element.
4. Submit the form.
5. The form fails to submit. In the browser console, the error "An invalid form control with name='' is not focusable" appears.

## Testcase
Here's the fiddle https://jsfiddle.net/ndpuks2y/2/

## Version
2.7.6

Having this issue on Firefox (68) and Chrome (75.0.3770.90) on Ubuntu. Firefox doesn't show any error but prevents the form submission.

## Note
According to this [post](https://stackoverflow.com/questions/7168645/invalid-form-control-only-in-google-chrome/7264966#7264966) on Stackoverflow, the select should remove the attribute "required" which is hidden by Javascript, in our case, the select element with dropdown component shouldn't have the attribute "required".

lubber-de:
## Description
When a dropdown was initialized out of a select tag and that tag had the 'required' attribute set, it is not possible to submit the form with an empty value, because the select-tag stays hidden. Instead a console warning is shown "An invalid form control with name='' is not focusable"

Basically we could argue "Just don't add 'required' to the select tag and use FUI form validation instead". But there might be situations where the form is generated by some backend framework where the "required" attribute is set automatically, thus i just removed the required attribute in any case as a proposal to solve the situation.

> This PR possibly breaks existing code, in case some developer have used the "required" attribute explicitly to gain the current behavior (and don't care about the console warning)
> __Therefore and also because of the possible argument "use fui form validation" (see above) I am  open to get this PR rejected if the other contributors do not agree__  

> ### I am a bit unsure myself if we should really merge this, so please share your opinions

## Testcase
- Select the empty entry of the dropdown
Try to submit the form via the blue "submit" button

### Broken
Does not submit, console warning instead
https://jsfiddle.net/t6sa9gv2/2/

### Working
Form submits
https://jsfiddle.net/t6sa9gv2/1/

## Closes
#917


ko2in:
@lubber-de Thank you. Exactly, I'm having this issue when working with Symfony form component. The form generates by Symfony dynamically and the required attribute is set when a field is set as required.

I can set as false with required property in Symfony form library, but which remove also from form field container, so, there is no indication that the field is required in UI. Because the field doesn't show the asterisk without the class name 'required'.

I think this change will affect the use of FUI form validation only. But in the real world project, no one would rely on client-side validation alone, and the server-side validation is required anyway. So, it makes no difference for me.

lubber-de:
@ko2in In your case you could also try to use "novalidate" at Symfony form creation

```php
{{ form_start(form, {'attr': {'novalidate': 'novalidate'}}) }}
```
This should make the browser ignore he "required" form field validation in general.

ko2in:
@lubber-de Wow, that does the trick :+1: . I've never been used 'novalidate' attribute and so, I've no clue to try with that. Thanks.

I'm not sure if there anyone else having a similar problem. If no one ever been mentioning, you may omit to merge this change. My problem solves with 'novalidate' attribute alone. So, I don't need this fix yet.

PR #946, opened 2019-08-21
Move away from travis and to GitGub Actions

hammy2899:
## Description
Move away from travis and use GitHub Actions instead.

My reasoning for this is because then our CI is in the same place as our code and also because GitHub Actions allows for more possibilities then travis.


hammy2899:
This should be good to review 👌

prudho:
This looks nice ! So it try building with 4 versions of node, but what happens if one build (let's say Node 8), triggers an error ? And is it possible to grab build artifacts (to push them into BrowserStack, or to make some tests on my machine for example) ?

hammy2899:
@prudho Yeah we could do that like I said anything is possible with Actions.

The only issue with your proposal is we don't have any automated tests to run in BrowserStack 😄

PR #947, opened 2019-08-22
Fix CI and build errors

hammy2899:
## Description

Fix the CI building twice and also throw the build errors so the script exists with an error instead of success


hammy2899:
I am going to force merge this since the CI is technically correct 😄

PR #954, opened 2019-08-25
fix(progress): wrong error on validating sum of rounded percents

### LINKED ISSUE #953: [Progress] Multiple progress shows an error although valid total/value have been set
# Bug Report

## Steps to reproduce
1. Set `total: 6`, `value: [1, 4, 1, 0]` and `precision: 0.1` in multiple progress
2. Invoke `progress()` twice

## Expected result
No console error is shown.
## Actual result
A console error is shown.
## Testcase
https://jsfiddle.net/qtnux0s6/

## Screenshot (when possible)
スクリーンショット 2019-08-25 18 12 10

## Version
2.7.7

ryamaguchi0220:
## Description

I propose to store unrounded percents in the `data-percent` attribute instead of rounded percents.

Here is what causes the wrong error.
① Store rounded percents in the `data-percent` attribute [here](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/modules/progress.js#L483)
② Get the rounded percents [here](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/modules/progress.js#L165)
③ Validate sum of the rounded percents  [here](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/modules/progress.js#L513).
    As [this comment](https://github.com/fomantic/Fomantic-UI/blob/master/src/definitions/modules/progress.js#L514) indicates, it is not expected to pass rounded percents to the `percent()`.

## Testcase
https://jsfiddle.net/yxLr6mut/

## Closes
#953

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -468,7 +468,7 @@ $.fn.progress = function(parameters) {
                   width: value + '%'
                 });
               }
-              return parseInt(value, 10);
+              return parseFloat(value);
             });
             values.forEach(function(_, index) {
               var $bar = $($bars[index]);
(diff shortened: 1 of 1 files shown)


hammy2899:
@all-contributors please add @ryamaguchi0220 for code and bug

PR #977, opened 2019-08-28
feat(tab): auto activate first tab if no active specified

### LINKED ISSUE #914: Make first tab active if not specified
# Feature Request

## Description
How about making the first tab active when not active tag added to the html code? 

I the below example no active tag is added, the content of the first tab is first shown when you click on it. 

## Testcase (when possible)
https://jsfiddle.net/t6we9gv8/

prudho:
## Description
This PR activate automatically the first tab if no active tab is specified trough the `active` class.

## Testcase
Before: [JSFiddle](https://jsfiddle.net/t6we9gv8/)
After: [JSFiddle](https://jsfiddle.net/kLzqex81/1/)

## Closes
#914

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -694,7 +694,7 @@ $.fn.tab = function(parameters) {
               if( $tab.hasClass(className.active) ) {
                 var
                   tabPath = $(this).data(metadata.tab),
-                  $anchor = $allModules.filter('[data-' + metadata.tab + '="' + module.escape.string(tabPath) + '"]');
+                  $anchor = $allModules.filter('[data-' + metadata.tab + '="' + module.escape.string(tabPath) + '"]')
                 ;
 
                 if( $anchor.hasClass(className.active) ) {
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -100,6 +100,11 @@ $.fn.tab = function(parameters) {
             initializedHistory = true;
           }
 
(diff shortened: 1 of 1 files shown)


src/definitions/modules/tab.js
```diff
@@ -679,6 +684,29 @@ $.fn.tab = function(parameters) {
           }
         },
 
+        determine: {
+          activeTab: function() {
+            var activeTab = null;
+
+            $tabs.each(function(_index, tab) {
```
exoego: `Array.forEach` is in efficient for this purpose since it iterate all elements (tabs).
We can use `Array.find` to find first element that satisy predicate when we drop IE support.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

For now, we can use `Array.some` that can stop iteration when first satisfying element found.

```javascript
var activeTab = null;
$tabs.some(function(_index, tab) {
    var $tab = $(tab);
    if( $tab.hasClass(className.active) ) {
        var
            tabPath = $(this).data(metadata.tab),
            $anchor = $allModules.filter('[data-' + metadata.tab + '="' + module.escape.string(tabPath) + '"]')
        ;

        if( $anchor.hasClass(className.active) ) {
            activeTab = tabPath;
            return true;
        }
    }
    return false;
});
```

prudho: Using the `some` function will make me convert my `$tabs` to an array, which is not wanted since I'll need to get the tab object by another way. I think that actually it's not the best option to use, but it's the most convenient.

PR #978, opened 2019-08-29
feat(slider): add auto adjusted label display

### LINKED ISSUE #913: Make the step of 'labeled slider' configurable
# Feature Request

## Description
Currently, a 'labeled' slider is labeled at every possible value. With large sliders, especially on mobile devices, the labeling is confusing. 
It even get's worse with floating-point steps - then, the labeling is unreadable, even on large screens.
This could be solved by allowing the step of the labeling to be configurable so that not every value is labeled, but maybe only every 4th.

Also, i think it would be great to be able to control the labeling-step independently from the ticks-step (of a `labeled *ticked* slider`)!

## Example/Screenshots
`ui large labeled slider` from 1 to 16 with step=1 on a mobile device
image  

`ui large labeled slider` from 1 to 16 with step=0.5 on desktop
image

prudho:
## Description
This PR adds dynamically adjusted labels to sliders, which will avoid unreadable labels when there's too many on screen.

It works by calculating the labels which respect a minimal distance between them (adjustable trough the new setting `labelDistance`; default `100`). All intermediate labels still exists, but they don't display a value, and their tick height is half sized.

This new feature can be disabled by setting `autoAdjustLabels` to `false`.

Oh, and the labels are also adjusted on window resizing too.

## Testcase
Before: [JSFiddle](https://jsfiddle.net/Lopwyghd/)
After: [JSFiddle](https://jsfiddle.net/xj827hfw/)

## Closes
#913

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -23,6 +23,7 @@ $.fn.slider = function(parameters) {
 
   var
     $allModules    = $(this),
+    $window        = $(window),
 
     moduleSelector = $allModules.selector || '',
 
@@ -86,6 +87,7 @@ $.fn.slider = function(parameters) {
         offset,
         precision,
         isTouch,
+        gapRatio = 1,
 
         module
       ;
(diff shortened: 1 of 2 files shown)


src/definitions/modules/slider.js
```diff
@@ -273,6 +282,9 @@ $.fn.slider = function(parameters) {
               $(document).on('mousemove' + eventNamespace, module.event.move);
               $(document).on('mouseup' + eventNamespace, module.event.up);
             }
+          },
+          windowEvents: function() {
+            $window.on('resize' + eventNamespace, module.event.resize);
```
lubber-de: Please add the appropriate unbind event in `module.unbind.events()`

```
$window.off('resize' + eventNamespace);

prudho: Oh right, didn't think about that ! Thanks !

PR #982, opened 2019-08-30
feat(calendar): add onBeforeChange callback and fix a date value issue

### LINKED ISSUE #417: [Date Picker] The date picker doesn't return the correct value when submitting the form with 'onChange' callback
# Bug Report
I've a form with date picker and I setup to submit the form as soon as the date is picked and send the form data to the server. But, the form always sends the previous selected date.

## Testcase
You can test in this fiddle (http://jsfiddle.net/m7wn3xus/3/). The date picker has `onChange` callback, which triggers the submit event of the form. You can see the serialized array of form data below. The form always send the previous value from the date picker.

## Current Workaround Solution
I've to update the input field value with the selected date before triggering the submit event. You can test in this fiddle (http://jsfiddle.net/yka8fpsr/1/). The form will be sent the correct date to the server.

In my real world project, I'm implementing the search filters with several date fields and send to the server via Ajax request when the fields have changed the values.

The browser native `onChange` event has no effect on the date picker field. Looks like the date picker module unbind the default events. Try this fiddle (http://jsfiddle.net/54gsqa27/). The two text fields listen the change event, but the date picker field doesn't. 

So, the only way to trigger the change event is using `onChange` callback from the date picker API, which causes the issue.

## Version
2.7.1

prudho:
## Description
This PR actually provide two things:
- a new callback, called `onBeforeChange`, which allow to cancel a change before it's applied to calendar logic. The old `onChange` callback is now moved further.

- a fix to the `onChange` callback. His position was bad, since the date wasn't really changed internally when it was called, so any call the one of the behaviors or using the input resulted in a bad value.

⚠️ This is a breaking change ! The new `onChange` callback, doesn't allow to cancel change anymore, so we might warn users when we'll publish this update.

## Testcase
Before: [JSFiddle](https://jsfiddle.net/a28ruv0x/1/)
After: [JSFiddle](https://jsfiddle.net/rL6ukpz1/)

## Closes
#417 
https://github.com/mdehoog/Semantic-UI-Calendar/issues/124

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -674,7 +674,8 @@ $.fn.calendar = function(parameters) {
 
             var mode = module.get.mode();
             var text = formatter.datetime(date, settings);
-            if (fireChange && settings.onChange.call(element, date, text, mode) === false) {
+
+            if (fireChange && settings.onBeforeChange.call(element, date, text, mode) === false) {
               return false;
             }
 
@@ -693,6 +694,10 @@ $.fn.calendar = function(parameters) {
 
             if (updateInput && $input.length) {
               $input.val(text);
+            }
+
+            if (fireChange) {
+              settings.onChange.call(element, date, text, mode);
             }
           },
           startDate: function (date, refreshCalendar) {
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/166

Issue #985, opened 2019-09-01
[Form Validation] Built-in rule to validate a field based on the value of another field


rogervanmars:
# Feature Request
A built-in rule to validate a form field based on the value from another field of the same form.

## Description
Now it is possible to validate a form field based on the current value of another field by writing a custom rule. That would look like: 
```
// custom form validation rule
$.fn.form.settings.rules.customRule = function(value) {
  return $(".ui.form").form("get value", "field2") == 'alpha' ? Number(value) : true;
};

$('.ui.form')
  .form({
    fields: {
    	field1: 'customRule',
        field2: 'empty'
    }
  })
;
```
It would be nice to have a built-in rule, like the ones describe [here](https://fomantic-ui.com/behaviors/form.html#validation-rules), to achieve the same result.

## Example
That could look like:
```
$('.ui.form')
  .form({
    fields: {
    	field1: 'fieldValue[field2, alpha, number]', 
        field2: 'empty'
    }
  })
;
```
where `fieldValue[field2, alpha, number]` means: _apply 'number' rule to field1 when field2's value is 'alpha'_

## Testcase (when possible)
not applicable

## Screenshot (when possible)
not applicable

PR #988, opened 2019-09-01
[Form Validation] Add form validation rule: condition

vibridi:
## Description
This PR adds the `condition` validation rule. It allows to apply an arbitrary rule to a field value depending on another field value.

## Test Case
After: [JSFiddle](https://jsfiddle.net/7ndsx3th/2/)


src/definitions/behaviors/form.js
```diff
@@ -1818,6 +1819,74 @@ $.fn.form.settings = {
...
+      if(rule.includes('condition')) {
+        // Don't support nested condition rules
+        return false;
+      }
+
+      var
+          fieldValue,
+          fieldElement
+      ;
+      if((fieldElement = $module.find('[data-validate="'+ identifier +'"]')).length > 0 ) {
```
lubber-de: I think you should use `module.get.value(identifier)` to fetch the fieldValue instead, because it will take care of all possible form types like checkbox, radio or calendar which need special handling to get the value.

vibridi: Hi @lubber-de , thank you for your advice! I've updated both PR and JSFiddle.

src/definitions/behaviors/form.js
```diff
@@ -1823,6 +1824,58 @@ $.fn.form.settings = {
...
+      if(params.length !== 3) {
+        return false;
+      }
+
+      let
+          identifier  = params[0].trim(),
+          condition   = params[1].trim(),
+          rule        = params[2].trim()
+      ;
+      if(rule.includes('condition')) {
```
lubber-de: Please use `indexOf('condition') >-1` instead as we are still supporting IE11

vibridi: Done
Updated PR and JSFiddle. I tested with a n IE11 emulator, it looks like it's working. 
Please let me know if I missed anything

lubber-de (dismissed):
LGTM 🙂 👍

src/definitions/behaviors/form.js
```diff
@@ -1823,6 +1824,58 @@ $.fn.form.settings = {
       ;
     },
 
+    // applies rule if another field's value satisfies a condition
+    condition: function(value, ancillary, $module) {
```
exoego: I have a little difficulty to understand the new syntax...

My understanding is `condition[FIELD_NAME, RULE1, RULE2, ... RULEn]`.
Am I on the right track?
If so, I feel it is a bit strange to treat `FIELD_NAME` and `RULE`s  in the same way.
I suggest a JSON-like syntax like `{ FIELD_NAME: [ RULE ], FIELD_NAME2: [RULE1, RULE2] }` that will cover use case to have conditions on multiple fields.

hammy2899: I agree, the current way will be too confusing for users and doesn't allow for expansion of the rule. I think your suggestion of making it JSON like will be our best shot to making it easy to use and easy to improve in the future.

vibridi: @exoego @hammy2899 Hello guys, based on what the author of this issue (#985) writes:
> A built-in rule to validate a form field based on the value from another field of the same form.

I intended to lay out the parameters as follows, basically using the syntax proposed by the issue author:
`condition[FIELD_NAME, VALUE_TO_CHECK_AGAINST, RULE_TO_APPLY]`
which means: "if THAT_FIELD's value is VALUE, then apply RULE to this field". 

Also I intended to set a hard limit on what `RULE_TO_APPLY` can be, namely one single existing rule, on purpose. I was looking to not risk having recursive rules, or excessively complicated config.

With that said I'm absolutely open to improve this. If you think JSON-like syntax is the way to go, I'll change accordingly 👍

hammy2899: I think making this JSON like is still the best option we have especially for future support.

vibridi: @hammy2899 please can you set out a small spec of how you envision this being JSON-like, then I'll try to take it from there. Thanks

hammy2899: > I have a little difficulty to understand the new syntax...
> 
> My understanding is `condition[FIELD_NAME, RULE1, RULE2, ... RULEn]`.
> Am I on the right track?
> If so, I feel it is a bit strange to treat `FIELD_NAME` and `RULE`s in the flat way.
> 
> I suggest a JSON-like syntax like `{ FIELD_NAME: [RULE], FIELD_NAME2: [RULE1, RULE2] }`.
> This syntax distinguishes `FIELD_NAME` and `RULE`s.
> This will also can cover use-case to have conditions on multiple fields.

Like how @exoego said

lubber-de: @vibridi Are you still interested in continuing to adjust the PR ?

vibridi: @lubber-de I'm sorry, I don't think I will be able to work on this at this time. Feel free to close or take over. Thanks!

Issue #986, opened 2019-09-01
Calendar flips Date and Month around


OssiPesonen:
# Bug Report

For reasons unknown our web app, which is using Fomantic UI, keeps flipping the month and date around when the input value is set to format 'j.n.Y H:i' and calendar formatter to day.month..year.

**This only happens on Chromium browsers, Chrome, Opera...** Not certain if it is locale specific (I am using Finnish)

For clarification, I am using PHP date() function format. j = day, n = month, Y = year if someone does not understand.

## Steps to reproduce
1. Set input value as php date `j.n.Y H:i`
2. Open view with Google Chrome which then formats the value and calls (at the bottom of the page), $('input').calendar()
3. Input (UI) shows n.j.Y H.i even though dev tools report j.n.Y H:i. This can only happen when date is below 13th making it eligible to pass as a month. When form is submitted the wrong format gets passed forward.

The calendar is formatted as such:

```    
$('#time').calendar({
    type: 'datetime',
    today: true,
    ampm: false,
    firstDayOfWeek: 1,
    text: {
      days: ['S', 'M', 'T', 'K', 'T', 'P', 'L'],
      months: ['Tammikuu', 'Helmikuu', 'Maaliskuu', 'Huhtikuu', 'Toukokuu', 'Kesäkuu', 'Heinäkuu', 'Elokuu', 'Syyskuu', 'Lokakuu', 'Marraskuu', 'Joulukuu'],
      monthsShort: ['Tam', 'Hel', 'Maa', 'Huh', 'Tou', 'Kes', 'Hei', 'Elo', 'Syy', 'Lok', 'Mar', 'Jou'],
      today: 'Tänään',
      now: 'Nyt',
      am: 'AM',
      pm: 'PM'
    },
    monthFirst: false,
    formatter: {
        date: function (date, settings) {
            if (!date) return '';
            var day = date.getDate();
            var month = date.getMonth() + 1;
            var year = date.getFullYear();
            return day + '.' + month + '.' + year;
        }
    }
});
```

The input field (Laravel blade view) is printed as so:

```
<input type="text" placeholder="{{ __("Aika") }}" name="time" value="{{ !empty($obj->time) ? date('j.n.Y H:i', strtotime($obj->time)) : '' }}">
```

## Expected result

Input (UI) field should contain the time format as j.n.Y H:i

## Actual result

Input shows date in format n.j.Y H:i 

## Testcase
https://jsfiddle.net/eow4Lqg5/

[...]

PR #995, opened 2019-09-04
[Calendar] Fix broken dates when keyboard typed

### LINKED ISSUE #829: [Calendar] breaks dates that entered from the keyboard
# Bug Report
The calendar breaks dates that entered from the keyboard, it changes the day and month. The option monthFirst: false does not help. I'm not sure that the defined formatter makes sense for this bug, but I faced it using formatter.
PS: It is better to have the possibility to make custom date parse function like formatter.
## Steps to reproduce
1. Open Fomantic-UI Calendar -> Example -> Custom format page https://fomantic-ui.com/modules/calendar.html#custom-format
2. Enter from the keyboard 1/8/2012
## Expected result
1/8/2012
## Actual result
8/1/2012
## Version
2.7.6

prudho:
# Description
This PR aim to fix (or at least drastically reduce) the broken date format when it's typed through keyboard. It's a bit hacky, since it only reverse supposed days and month when the setting `monthFirst` is defined to `true`, but I think this little piece of code will cover 98% of cases.

## Testcase
Before: [JSFiddle](https://jsfiddle.net/musjvb63/)
After: [JSFiddle](https://jsfiddle.net/z6ghfcv9/)

## Closes
#829 
#986

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1277,7 +1277,7 @@ $.fn.calendar.settings = {
         return null;
       }
       // Reverse date and month in some cases
-      text = settings.monthFirst ? text : text.replace(/([0-9]+)[\/\-]([0-9]+)/,'$2/$1');
+      text = settings.monthFirst ? text : text.replace(/([0-9]+)[\/\-\.]([0-9]+)/,'$2/$1');
       var textDate = new Date(text);
       if(!isNaN(textDate.getDate())) {
         return textDate;
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1276,6 +1276,8 @@ $.fn.calendar.settings = {
       if (text.length === 0) {
         return null;
       }
+      // Reverse date and month in some cases
+      text = settings.monthFirst ? text : text.replace(/([0-9]+)[\/\-]([0-9]+)/,'$2/$1');
       var textDate = new Date(text);
       if(!isNaN(textDate.getDate())) {
         return textDate;


lubber-de:
🤔 Isn't the change assuming the provided text is always in  "wrong" order when monthFirst = false? It would exchange those two values all the time.
I think we should rather replace the whole parser by a proper method which always uses a given dateFormat setting without the need to rely on heavy extra libraries like moment.js

I think adopting something like the following tiny but effective method is suitable
http://blog.stevenlevithan.com/archives/date-time-format

prudho:
For users who use the "dd/mm/yyyy" format, they'll type as it, so the date will be always be corrected for them. People that use `monthFirst: false` are, for the vast majority, the same people that use the "dd/mm/yyyy" format (let's say 98% 😃).

I know that it's not the best solution, but write a brand new date parser, which will handle all dates specificities (timezones, format...), will need a heavy reflection (and we must not forget that calendar can also handle minutes, month and years only).

src/definitions/modules/calendar.js
```diff
@@ -1276,6 +1276,8 @@ $.fn.calendar.settings = {
       if (text.length === 0) {
         return null;
       }
+      // Reverse date and month in some cases
+      text = settings.monthFirst ? text : text.replace(/([0-9]+)[\/\-]([0-9]+)/,'$2/$1');
```
lubber-de: Lets also support a dot

```
text.replace(/([0-9]+)[\/\-\.]([0-9]+)/
```

, just because most of the people which are used to dd.mm.yyyy are most likely using a dot and won't enter 03-12-2019
Without the dot support it is still failing.

prudho: Enhanced with by, 2359db3, thank you @lubber-de !

PhilippGrashoff:
Hi,
I just tried the fix with the German (and I assume for many other european countries, too) standard Date format DD.MM.YYYY. So a dot instead of a slash. In this case the fix does not work, but should be easy to implement, since only the seperator sign is different?

PhilippGrashoff:
> I also changed the jsfiddle with a new one to prove the dot usage is also fixed

Hi, I tested the jsFiddle https://jsfiddle.net/z6ghfcv9/, but its failing as in lubber_de's gif. It picks  12th of october if a dot is used as seperator

lubber-de:
@PhilippGrashoff 
😱 I confirm, it still happened when using at least Edge/IE11 ...
I created another Hotfix #1004 which should work in any browser now
Please try this new jsfiddle https://jsfiddle.net/9caf67pv/

Issue #1000, opened 2019-09-05
[Slider] onChange ran when initialized


hammy2899:
# Bug Report

## Steps to reproduce
1. Create a slider with `onChange` method
2. View the page

## Expected result
Nothing should happen in the `onChange` method and the slider should be initialized as normal

## Actual result
Slider is initialized but `onChange` is ran

## Testcase
https://jsfiddle.net/jgwkd75c/

## Screenshot
![asbYBqd5Eu](https://user-images.githubusercontent.com/11588822/64341103-91115680-cfdf-11e9-8ed4-83147a58e8c0.gif)

## Version
2.7.8

hammy2899:
This is because `set.value()` is ran in the `read.settings()` method here https://github.com/fomantic/Fomantic-UI/blob/develop/src/definitions/modules/slider.js#L1002

and `read.settings()` is ran in the `initialize` method here https://github.com/fomantic/Fomantic-UI/blob/develop/src/definitions/modules/slider.js#L112

GammaGames:
I always assumed this was a feature, dropdown does it too. I do see that a few other modules have a `fireOnInit` setting, but I'm not sure if it's working on the Rating module. https://jsfiddle.net/x8ue7dmq/

I think not calling `onChange` by default and offering a `fireOnInit` for every input would be the most consistent.

PR #1006, opened 2019-09-09
[Rating] Fix unworking `fireOnInit`

prudho:
# Description
Following @GammaGames's comment on #1000, i found out that `fireOnInit` parameter didn't work (ever ?). So here's a fix.

I could have done it like this :
```js
if(!module.is.initialLoad() || settings.fireOnInit) {
  settings.onRate.call(element, rating);
}
```
But all the `initialLoad` logic is actually useless, so why not clean the code instead !

## Testcase
Before: [JSFiddle](https://jsfiddle.net/s04wc895/)
After: [JSFiddle](https://jsfiddle.net/wmkcarj2/)

### DIFF
--- a/src/definitions/modules/rating.js
+++ b/src/definitions/modules/rating.js
@@ -57,7 +57,6 @@ $.fn.rating = function(parameters) {
         $module         = $(this),
         $icon           = $module.find(selector.icon),
 
-        initialLoad,
         module
       ;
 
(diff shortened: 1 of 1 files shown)


prudho:
You know what ? I'm dumb and tired, so let's close this one and forget it in the deep of the interwebz...

PR #1007, opened 2019-09-09
[Rating] Fix unworking `fireOnInit`

prudho:
## Description
Basically the same as #1006, but done after a coffee and brain usage...

### DIFF
--- a/src/definitions/modules/rating.js
+++ b/src/definitions/modules/rating.js
@@ -268,7 +268,7 @@ $.fn.rating = function(parameters) {
                   .addClass(className.active)
               ;
             }
-            if(!module.is.initialLoad()) {
+            if(!module.is.initialLoad() || settings.fireOnInit) {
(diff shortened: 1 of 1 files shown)


exoego:
> #1006
> But all the initialLoad logic is actually useless, so why not clean the code instead 

So, is this line wrong ?

prudho:
@exoego yes, it's wrong. Actually the code can be (very) simplified, but we must keep this flag.

prudho:
@hammy2899 please wait a little before merging it, my next PR will show how i can improve the code.

prudho:
Please take a look at my PR #1008. Do you want it to be simplified like that ?

hammy2899:
@prudho Should just close this and use the other PR.

prudho:
Sure ! Will do.

ColinFrick:
@prudho so this PR should be closed right?

PR #1011, opened 2019-09-11
[Container] increased specificity to correctly override grid container

### LINKED ISSUE #1010: Mixing grid and container (vs nested usage) results in differing container width
# Bug Report

At a cursory glance, it appears there was a move away from negative margins on grid containers in a recent update. The result of this is that when [mixing `grid` and `container`](https://fomantic-ui.com/collections/grid.html#containers) the resulting width of that container is different from what we'd get if we just nested a `grid` inside a `container`. This resulted in a visual change for me when upgrading from Semantic-UI to Fomantic-UI.

## Steps to reproduce
1. Create both of these:
```html
<div class="ui three column grid container">
   <div class="column"></div>
</div>
```

```html
<div class="ui container">
   <div class="ui three column grid">
      <div class="column"></div>
   </div>
</div>
```

## Expected result

Both versions should have the same width.

## Actual result

When viewed with Semantic UI v2.4.1 they will both have the same overall width. When viewed with Fomantic UI v2.7.8 the combined version will have a lesser width (2rem from what I can tell reading the CSS).

## Version
2.7.8

lubber-de:
## Description
The specificity of `grid container` was too low resulting in not applied widths

### Broken
https://jsfiddle.net/xr324Lkg/1

### Fixed
https://jsfiddle.net/xr324Lkg/

## Closes
#1010


hammy2899:
This is good but is there a reason on mobile FUI has more horizontal margin? I can't remember any reason for this?

lubber-de:
> This is good but is there a reason on mobile FUI has more horizontal margin? I can't remember any reason for this?

🤔 At least this is the same behavior as with SUI 2.4.1  https://jsfiddle.net/50ndz9cs/

Let me check if we can fix it together in this PR as well

lubber-de:
@hammy2899 
Mmh, since SUI ages `@mobileGridWidth` had the same value as `@mobileWidth` (which is `auto`). All other breakpoints have a different explicit calculations. That's why.
We could fix it by changing the calculation for `@mobileGridWidth` according the the other breakpoint adjustments for example https://jsfiddle.net/xr324Lkg/2/
I think that additional change would be worth a separate discussion so lets have a separate PR for it if necessary.

https://github.com/fomantic/Fomantic-UI/blob/4b5cd6bc0dd17cf2645d3a5f1564cfe20691b7a3/src/themes/default/elements/container.variables#L35-L39

brendon:
Thanks for the prompt action everyone! :)

Issue #717, opened 2019-05-02
Large modals detached from DOM are centered strangely


Merlin04:
# Bug Report

## Steps to reproduce
1. Create a modal with more contents than can fit on screen
2. Initialize the modal with `detachable: false`
3. Open it

## Expected result

Modal starts at top of screen

## Actual result

Modal starts at middle of screen and also has no bottom padding

## Testcase

http://jsfiddle.net/pwhxv4o2/2/

## Version
Not sure the number but the latest version

Merlin04:
A temporary patch for this is to set the CSS property `top` to a low percentage, but this still leaves a large empty space below the modal.

PR #1013, opened 2019-09-12
fix(modal): vertical alignment on detachable modal was incorrect

### LINKED ISSUE #1009: Modal not centered correctly when using detachable: false
# Bug Report

Hello, it seems that the modal is not centering correctly when using option detachable: false and we have a long scrollable content.

My guess is that it is not calculating the top property correctly, it seems to be using the body height instead of viewport height.

Note 1: If you remove detachable : false option it works.
Note 2: If you remove like a paragraph of text you will then see the modal, but it won't be centered though.

## Steps to reproduce
1. Open the fiddle and click on show modal button

## Expected result
Modal should be centered.

## Actual result
Modal is not centered. 

## Testcase
[jsFiddle link](http://jsfiddle.net/7j38xdbw/1/)

My original Semantic-ui repository issue [link](https://github.com/Semantic-Org/Semantic-UI/issues/6866)

## Version
2.7.8

prudho:
## Description
First, i'm not sure this is the best solution, but it seems to fix the problem.
Second, this needs to be heavily tested, so don't approve it until you've tested many cases, please.

When modal is not detachable, it use the legacy display system, not flex based. So it calculates the position of the modal based on his window position, then apply a margin to center it. It works, when you don't have scrolled...

For me, the margin trick is not the best solution, so i tried to center the modal by calculating the top property instead (hope what I write is understandable), but with including the scroll position.

This is not a breaking change, but regarding the nature of the bug I think that we need to be careful with this PR.

Et voilà !

## Testcase
Before: [JSFiddle](http://jsfiddle.net/7j38xdbw/1/)
After: [JSFiddle](http://jsfiddle.net/dgzLjxwv/)

## Closes
#1009 
#717

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -806,17 +806,13 @@ $.fn.modal = function(parameters) {
             }
           },
           modalOffset: function() {
-            var
-              width = module.cache.width,
-              height = module.cache.height
-            ;
             $module
               .css({
-                marginTop: (!$module.hasClass('aligned') && module.can.fit())
-                  ? -(height / 2)
-                  : 0,
-                marginLeft: -(width / 2)
-              })
+                top: (!$module.hasClass('aligned') && module.can.fit())
+                  ? parseInt(($(window).height() / 2) + $(document).scrollTop() - ($module.height() - settings.padding))
+                  : $(document).scrollTop() + (settings.padding / 2),
+                marginLeft: -(module.cache.width / 2)
+              }) 
             ;
             module.verbose('Setting modal offset for legacy mode');
           },
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -694,10 +694,7 @@ $.fn.modal = function(parameters) {
 
         can: {
           useFlex: function() {
-            return (settings.useFlex == 'auto')
-              ? settings.detachable && !module.is.ie()
-              : settings.useFlex
-            ;
+            return settings.useFlex && settings.detachable && !module.is.ie();
           },
           fit: function() {
             var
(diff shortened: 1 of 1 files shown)


src/definitions/modules/modal.js
```diff
@@ -806,18 +803,25 @@ $.fn.modal = function(parameters) {
...
-            ;
+            if (!settings.detachable) {
+              $module
+                .css({
+                  top: (!$module.hasClass('aligned') && module.can.fit())
+                    ? parseInt(($(window).height() / 2) + $(document).scrollTop() - ($module.height() - settings.padding))
+                    : $(document).scrollTop() + (settings.padding / 2),
+                  marginLeft: -(module.cache.width / 2)
+                }) 
+              ;
```
lubber-de: Unfortunately this part is somehow not correctly working when the modal is bigger and the background body has less content
See http://jsfiddle.net/bkh17Lgy/
Without this PR change that example was working http://jsfiddle.net/bkh17Lgy/1/

@prudho Could you please take a look again at this and prepare another PR? 😉

lubber-de: Opened #1079 for this

testinfected:
@lubber-de We have a use case where we using `detachable:false` and `useFlex: true` to drag the Modal around. We recently upgraded to 2.8.2 and it no longer works.

Any idea on how we could force `useFlex` with `detachable:false` like the old behavior did ?

lubber-de:
@testinfected  You can override the internal method to decide for the flex-setting and always return true:
```javascript
var myModal = $('.ui.modal').modal({
  detachable: false
});
myModal.modal('internal',{
  can: {
    useFlex: function(){
    	return true; 
    }
  }
});
myModal.modal('show');
```

See https://jsfiddle.net/a2ukrjme/1/

testinfected:
Will do that, thanks

PR #1027, opened 2019-09-19
Update dimmer.less

justkey007:

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -63,7 +63,7 @@
 
 
 /* Loose Coupling */
-.ui.segment > .ui.dimmer {
+.ui.segment > .ui.dimmer:not(.page) {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dimmer.less
```diff
@@ -63,7 +63,7 @@
 
 
 /* Loose Coupling */
-.ui.segment > .ui.dimmer {
+.ui.segment > .ui.dimmer:not(.page) {
```
exoego: What is a motivation to this change?

justkey007: When the dimmer is inside a segment and we add the page class to extend it on the screen, it inherits the border radius property of the segment, which is unpleasant. Once extended to the page it must be fully overlaid on the page.

hammy2899:
@all-contributors please add @justkey007 for code

PR #1028, opened 2019-09-19
[Calendar] Don't limit word length for month parsing to support all languages

### LINKED ISSUE #1026: [calendar] only considers the first 3 letters to identify month words
# Bug Report

When entering a date using words, calendar only considers the first 3 letters of the month, which in most languages is probably sufficient, but not in french. As a result `juillet` (july) is always interpreted as `juin` (june).

## Steps to reproduce
1. Configure calendar with:
```javascript
$('.ui.calendar').calendar({
    type: 'date',
    text: {
        days: ["D","L","M","M","J","V","S"],
        months: ["janvier","février","mars","avril","mai","juin","juillet","août","septembre","octobre","novembre","décembre"],
        monthsShort: ["janv.","févr.","mars","avr.","mai","juin","juill.","août","sept.","oct.","nov.","déc."],
        today: "Aujourd\u0027hui",
        now: "\u0027Maintenant\u0027",
        am: "a.m.", 
        pm:"p.m."
    },
    monthFirst: false,
    formatter: {
        date: function (date, settings) {
            if (!date) return '';
            return date.toLocaleDateString('fr-FR');
        }
    }
});
```

2. Enter `5 juillet 2019` (July 5th)

## Expected result

`05/07/2019` (July 5th)

## Actual result

`05/06/2019` (June 5th)

## Testcase
https://jsfiddle.net/xwmz7ha5/

## Version
2.7.6

lubber-de:
## Description
When entering a date using words, calendar only considers the first 3 letters of the month, which in most languages is probably sufficient, but not in french. As a result juillet (july) is always interpreted as juin (june).
It's also an issue for many other languages according to https://library.princeton.edu/departments/tsd/katmandu/reference/months.html

## Testcase
Enter "5 juillet 2019" (July 5th)

#### Broken
https://jsfiddle.net/bdanwq7t
-> 05/06/2019 (June 5th)

#### Fixed
https://jsfiddle.net/bdanwq7t/2/
-> 05/07/2019 (July 5th)

## Closes
#1026


src/definitions/modules/calendar.js
```diff
@@ -1300,10 +1304,10 @@ $.fn.calendar.settings = {
 
         //time with ':'
         for (i = 0; i < numbers.length; i++) {
-          var number = numbers[i];
```
exoego: why removed `var`?
It seems that the variable `numbers` defined in `header` function scope, but no `number` defined, so it might be leaked into window (global) scop.e

src/definitions/modules/calendar.js
```diff
@@ -1324,14 +1328,14 @@ $.fn.calendar.settings = {
       if (!isTimeOnly) {
         //textual month
         for (i = 0; i < words.length; i++) {
-          var word = words[i];
+          word = words[i];
```
exoego: It seems that `var` is necessary as same in https://github.com/fomantic/Fomantic-UI/pull/1028/files#r326366079

src/definitions/modules/calendar.js
lubber-de: I moved the "number" (as well as "word") var declaration to the top of the function so it still has function scope.

      var words = text.split(settings.regExp.dateWords);
      var numbers = text.split(settings.regExp.dateNumbers);
      var words = text.split(settings.regExp.dateWords), word; //<-- Moved declaration here
      var numbers = text.split(settings.regExp.dateNumbers), number;  //<-- Moved declaration here

I removed it here, because its being unnecessarily redefined on each for loop iteration (micro optimation)

src/definitions/modules/calendar.js
lubber-de: See above, i moved all of the declarations to the top of the function 😉

src/definitions/modules/calendar.js
```diff
@@ -1176,6 +1176,7 @@ $.fn.calendar.settings = {
   centuryBreak       : 60,         // starting short year until 99 where it will be assumed to belong to the last century
   currentCentury     : 2000,       // century to be added to 2-digit years (00 to {centuryBreak}-1)
   selectAdjacentDays : false,     // The calendar can show dates from adjacent month. These adjacent month dates can also be made selectable.
+  monthParseLength   : 3,
```
exoego: I thinkt adding new setting may complicate calendar modules for non-English users and for multi-lingual apps.
Instead of new settings, how about making the parsing logic works with any-length ?

lubber-de: Agree 👍 I removed the extra setting now and also removed the length calculation limit now by fb81e3e62089f6f0184c77b0fce96963a2fa9018, much easier. Thx for the suggestion!

PR #1049, opened 2019-10-01
feat(table): add colored left & right row border

### LINKED ISSUE #1030: [Table] Coloured Border Left
# Feature Request
Essentially a bit like https://fomantic-ui.com/collections/table.html#colored but with the ability to apply it to the left (or right) table row border. A transparent option may also be required.

## Example
https://jsfiddle.net/5for3buz/

## Screenshot

image

lubber-de:
## Description
This PR adds optional `left marked` or `right marked` to  table row `tr`.
This way a colored table row can be displayed with a small border bar instead of the whole row background

## Testcase
https://jsfiddle.net/j0x1cfqo/8/

## Screenshot
image

## Closes
#1030

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -141,9 +141,6 @@
   padding: @cellVerticalPadding @cellHorizontalPadding;
   text-align: @cellTextAlign;
 }
-.ui.table tr.left > :first-child {
-  padding-left: @cellHorizontalPadding - @coloredBorderSize;
-}
 
 /* Icons */
 .ui.table > .icon {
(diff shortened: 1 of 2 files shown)


hammy2899:
Maybe we should also add this to individual columns. We could put it in a separate PR to keep this to the scope of the fixed issue.

lubber-de:
> Maybe we should also add this to individual columns. We could put it in a separate PR to keep this to the scope of the fixed issue.

Providing it for cells as well is a matter of a few LOC only... I can do another PR of course when this is merged (which will increase my Hacktoberfest counter 😄 )
or i do it right here and everybody needs to approve again... i don't care

hammy2899:
@lubber-de I think another PR so this one stays to scope.

Issue #834, opened 2019-06-25
[Form] Unexpected gap between the checkbox and other field


ko2in:
# Bug Report
The checkbox set top margin inappropriately which causes unexpected gap between the checkbox and other form fields.

## Expected result
The checkbox shouldn't display too far from the other field.

![Checkboxes-2](https://user-images.githubusercontent.com/930315/60098044-b7627c80-977a-11e9-97e3-83bda4d043e8.png)

## Actual result
The checkbox displays too far from the other field above. Here's the [fiddle](https://jsfiddle.net/yfesupjh/2/).

![Checkboxes](https://user-images.githubusercontent.com/930315/60096448-fd1d4600-9776-11e9-99bc-ac8eea54f9c1.png)

I've tested with version 2.7.4 and older versions, but the checkbox stays too close to other field above. Here's the [fiddle](https://jsfiddle.net/jneLo378/1/).

![Checkboxes-3](https://user-images.githubusercontent.com/930315/60098326-44a5d100-977b-11e9-87a2-42852ccb4d6e.png)

I've a test version with some changes which produces the best output with more complex form layout. You can test in this [fiddle](https://jsfiddle.net/stz9mfbr/2/).

I can send PR later base on this fiddle. I'm busy for my project at the moment, so, it may take longer to send my PR. If someone would like to contribute, please feel free to use my fiddle and send PR.

## Version
2.7.6

Darking360:
I can work on this 🚀 I'm cloning and setting everything up 🙏

lubber-de:
@all-contributors please add @ko2in for doc, review and bug

amansaurabh:
Is it still open?

lubber-de:
> Is it still open?

Yes, the aproached PR #1054 did not fix it (see comments there)

PR #1054, opened 2019-10-02
[Checkbox] Fix gap in form fields

Darking360:
## Description

This PR resolves a big gap that was inside `.field` `divs` due to `form.less` styling on last items from `.fields`, removing the line for the last child of `.fields`,  adding more precedence on the selectors for `.ui.[class].checkbox` and adding a `margin-bottom` of `14px` to the `.field` inside a `.ui.form` resolves the issue with the examples attached below:

## JS Fiddles

- [Way too separated](https://jsfiddle.net/yfesupjh/2/)
- [Too close](https://jsfiddle.net/jneLo378/1/)

## Before

image

## After

image

Much better spacing.

## Closes

https://github.com/fomantic/Fomantic-UI/issues/834

### Original

Credits and shout out to @ko2in for coming with the JSFiddle to resolve this 🚀

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -47,14 +47,17 @@
   margin: @fieldMargin;
 }
 
-.ui.form .field:last-child,
-.ui.form .fields:last-child .field {
+.ui.form .field:last-child {
   margin-bottom: 0;
 }
 
 .ui.form .fields .field {
   clear: both;
   margin: 0;
+}
+
+.ui.form .fields .field {
+  margin-bottom: 14px;
 }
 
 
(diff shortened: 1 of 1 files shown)


lubber-de:
🤔 It works in the mentioned use cases, but only if all checkboxes are in a row.
If only one column does have a usual input field, it breaks the view.
See https://jsfiddle.net/wn8p0krx/

#### Before your fix
image

#### After your fix
image

src/definitions/collections/form.less
```diff
@@ -57,6 +56,10 @@
   margin: 0;
 }
 
+.ui.form .fields .field {
+  margin-bottom: 14px;
```
lubber-de: Use the existing variable `@rowDistance`  instead 

```css
margin-bottom: @rowDistance;
```

lubber-de (changes requested):

Darking360:
Ohhh I see, holly, then it's trickier, we'd need to know that the `.field` children of that `.fields` container only contain as last child the checkbox to add the style to separate them a little less, but if we encounter a different item, then apply the usual `margin` separation 🤔 Should this be accomplished only using `less`/`css` or is ok to use Javascript as well? I don't know if scripting should be avoid to fix this. If not, I think that then the separation reported is the usual behavior for this column fields right? @lubber-de 🤔

lubber-de:
@Darking360 Yes, it should be accomplished using CSS/LESS only.

PR #1056, opened 2019-10-03
fix close modal incorrect behaviour

### LINKED ISSUE #1055: Mouse down event inside modal and mouse up outside modal should not close modal dialog when detachable false
# Bug Report

Mouse down event inside modal and mouse up outside modal should not close modal dialog.

It works with detachable true; It works on firefox but not chrome. 

It seems that the mousedown button is not triggering when detachable false because modal div is not inside the dimmer div anymore.

## Steps to reproduce
1. Open fiddle 
2. Click inside the modal selecting the text input
3. Release the mouse button outside of the modal

## Expected result
Modal should not be closed if mouse down event were triggered inside modal dialog

## Actual result
Modal is being closed on mouse up event

## Testcase
[Fiddle](http://jsfiddle.net/1hekjqrb/3/)

## Version
component modal 2.7.8

rmarchiori:
## Description
Since the modal element is not inside dimmer when detachable false, mouse down event was not being triggered. I added mouse down event to modal too for theses cases. 

## Testcase
[fixed fiddle](http://jsfiddle.net/1eLyqtk0/3/)

## Closes
#1055


hammy2899:
@all-contributors please add @rmarchiori for code

PR #1066, opened 2019-10-05
fix(calendar): prevent toggle range class at today button

### LINKED ISSUE #901: "Today" in calendar popup shares className with range cells
# Bug Report
Apparently "Today" button in calendar popup shares _className_ with range cells.

## Steps to reproduce
1. Select a date as "startDate" of the range, from the first input
2. Open calendar popup for the second input
3. Try to mouseover cells _BEFORE_ and _AFTER_ **today** cell, as you would select the "endDate" for the range

## Expected result
"Today" button should not change color, not depending on the date you select as "endDate" of the range

## Actual result
"Today" button has different backgroud and text color based on the date you select: if after **today** it takes the same colors.

## Testcase
https://jsfiddle.net/a0re4nyd/1/

## Version
2.7.6

iwgx:
## Description

As reported in #901 that today button is also affected when select the end date because it's toggle when that's in range or not.

So I created helper `isTodayButton`, that check text inside the cell if that same with `settings.text.today`. If return true then don't toggle the range class.

### Before
https://jsfiddle.net/t7f50Lk8/

### After
https://jsfiddle.net/t7f50Lk8/1/

### Before
before

### After
after

## Closes
#901

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -416,6 +416,10 @@ $.fn.calendar = function(parameters) {
                 ((!!startDate && module.helper.isDateInRange(cellDate, mode, startDate, rangeDate)) ||
                 (!!endDate && module.helper.isDateInRange(cellDate, mode, rangeDate, endDate)));
               cell.toggleClass(className.focusCell, focused && (!isTouch || isTouchDown) && (!adjacent || (settings.selectAdjacentDays && adjacent)) && !disabled);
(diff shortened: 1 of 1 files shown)


lubber-de:
Thanks for the PR 👍 I added appropriate Testcase jsfiddles to the PR description

hammy2899:
@all-contributors please add @iwgx for code

PR #1067, opened 2019-10-05
feat(form): accept 1 in addition to true as checked value for checkboxes

### LINKED ISSUE #987: .form('set values', recordObject) with a checkbox
# Help Wanted

## Problem
I'm using a service which provides JSON data which represents boolean values as the numbers 0 and 1 instead of true and false. Is there a way to make .form('set values', recordObject) work with a checkbox with data coming in this format?

lubber-de:
## Description
When using the behavior `set value` or `set values`, checkbox elements were only supporting a real boolean `true` to check any checkbox.
This PR additionally supports the integer value `1` to have a checkbox checked now.

#### Not working
https://jsfiddle.net/6eL31k90

#### Working
https://jsfiddle.net/6eL31k90/1/

## Closes
#987

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1053,7 +1053,7 @@ $.fn.form = function(parameters) {
                 }
                 else if(isCheckbox) {
                   module.verbose('Setting checkbox value', value, $element);
-                  if(value === true) {
+                  if(value === true || value === 1) {
                     $element.checkbox('check');
                   }
                   else {


src/definitions/behaviors/form.js
```diff
@@ -1053,7 +1053,7 @@ $.fn.form = function(parameters) {
                 }
                 else if(isCheckbox) {
                   module.verbose('Setting checkbox value', value, $element);
-                  if(value === true) {
+                  if(value === true || value === 1) {
```
exoego: I think limiting to boolean is reasonable to avoid confusing on truthy values.
For numbers, an usage-site type conversion like `$chekbox.checkbox("set value", value === 1)` is suffiient.

I am proposing this to avoid limitless widening acceptable values in case someone request "Please accept string 'true' too".

lubber-de: Good point, but for checkbox in this only case, i wouldn't care so much, as it's optional and 0/1 used as false/true is widely used. 
I tested before using `Boolean(value) === true`, but then anything else also was true (even '0' is true then!), So i would definaltey not accept strings here, just because its not common.

But i am fine to decline the issue request and don't merge this if most of the other contributors also think we should stick with pure boolean values.

hammy2899: I agree both but I think leaving it as just boolean and 1/0 should be fine and should reduce any issues we could face.

esmondb: I've ended up trying to write my own version roughly based on this example [https://6pac.github.io/SlickGrid/examples/example-composite-editor-item-details.html](url). It's more of a form controler.
(my original fix was a !!)

PR #1073, opened 2019-10-07
fix(dropdown): action button has incorrect border radius with dropdown

### LINKED ISSUE #1072: [Dropdown] action button has incorrect border radius
# Bug Report

## Expected result
Dropdown should not have border radius between the action.

## Actual result
Dropdown has border radius as it normally would.

## Testcase
https://jsfiddle.net/pxstuqo9/

## Screenshot (when possible)
image

## Version
2.7.8

## Related issues
https://github.com/Semantic-Org/Semantic-UI/issues/5314

RShalman:
## Description
Fixes issue #1072 + refactors CSS keeping same logic but less misguiding.

### Before
https://jsfiddle.net/qrv7xw2k/

### After
https://jsfiddle.net/7gf135dc/

## Closes
#1072

### DIFF
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -487,8 +487,12 @@
 }
 
 /* Input Focus */
-.ui.ui.action.input:not([class*="left action"]) > input:focus {
+.ui.action.input:not([class*="left action"]) > input:focus {
   border-right-color: @focusBorderColor;
+}
+
+.ui.ui[class*="left action"].input > input:focus {
+  border-left-color: @focusBorderColor;
 }
 
 /*--------------------
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -456,14 +456,14 @@
 }
 
 /* Input when ui Left*/
-.ui[class*="left"].action.input > input {
+.ui[class*="left action"].action.input > input {
   border-top-left-radius: 0 !important;
   border-bottom-left-radius: 0 !important;
   border-left-color: transparent !important;
 }
 
 /* Input when ui Right*/
-.ui.action.input:not([class*="left"]) > input {
+.ui.action.input:not([class*="left action"]) > input {
   border-top-right-radius: 0 !important;
   border-bottom-right-radius: 0 !important;
   border-right-color: transparent !important;
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -456,17 +456,17 @@
 }
 
 /* Input when ui Left*/
-.ui[class*="left action"].action.input > input {
-  border-top-left-radius: 0 !important;
-  border-bottom-left-radius: 0 !important;
-  border-left-color: transparent !important;
+.ui[class*="left action"].input > input {
+  border-top-left-radius: 0;
+  border-bottom-left-radius: 0;
+  border-left-color: transparent;
 }
 
 /* Input when ui Right*/
 .ui.action.input:not([class*="left action"]) > input {
-  border-top-right-radius: 0 !important;
-  border-bottom-right-radius: 0 !important;
-  border-right-color: transparent !important;
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
+  border-right-color: transparent;
 }
 
 /* Button and Dropdown */
@@ -487,8 +487,8 @@
 }
 
 /* Input Focus */
-.ui.action.input:not([class*="left action"]) > input:focus {
-  border-right-color: @focusBorderColor !important;
+.ui.ui.action.input:not([class*="left action"]) > input:focus {
+  border-right-color: @focusBorderColor;
 }
 
 /*--------------------
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -455,8 +455,15 @@
   margin: 0;
 }
 
-/* Input */
-.ui.action.input > input {
+/* Input when ui Left*/
+.ui[class*="left"].action.input > input {
+  border-top-left-radius: 0 !important;
+  border-bottom-left-radius: 0 !important;
+  border-left-color: transparent !important;
+}
+
+/* Input when ui Right*/
+.ui.action.input:not([class*="left"]) > input {
   border-top-right-radius: 0 !important;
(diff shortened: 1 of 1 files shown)


RShalman:
> Your complete removement of `left action` breaks existing, non-dropdown, inputs:
> 
> * Right input border is lost
> * Left input border has border radius
> 
> #### Before your change
> https://jsfiddle.net/8hvjwr1p
> image
> 
> #### After your change
> https://jsfiddle.net/8hvjwr1p/1/
> 
> image

Thanks for feedback! 
Fixed that issue also. Thanks for pointing on that!

src/definitions/elements/input.less
```diff
@@ -455,20 +455,34 @@
   margin: 0;
 }
 
-/* Button on Right */
-.ui.action.input:not([class*="left action"]) > input {
+/* Input when ui Left*/
+.ui[class*="left"].action.input > input {
```
lubber-de: Please change to
```css
.ui[class*="left action"].input > input {
```
Because it won't work with `<div class="ui right action input left icon">` otherwise

src/definitions/elements/input.less
```diff
@@ -455,20 +455,34 @@
...
+  border-top-left-radius: 0 !important;
+  border-bottom-left-radius: 0 !important;
+  border-left-color: transparent !important;
+}
+
+/* Input when ui Right*/
+.ui.action.input:not([class*="left"]) > input {
```
lubber-de: Same here, please change to
```css
.ui.action.input:not([class*="left action"]) > input {
```
to not interfere with `<div class="ui right action input left icon">`

RShalman:
> Please see my code comments: (basically just turn `class*="left"` into `class*="left action"`)
> 
> When using `<div class="ui right action input left icon">` your current changes have the following effect because `left` could be used for a different behavior than just `action`
> 
> * Left input border is lost
> * Right input border has border radius
> 
> #### Before your change
> https://jsfiddle.net/4L7zvsbc
> 
> image
> 
> #### After your change
> https://jsfiddle.net/4L7zvsbc/1/
> 
> image

Fixed this, as you've asked! 
However, as for my opinion, those CSS constructions + class applications to inputs' containers are a thin thing. If only you change the order of the classes applied you will also need to change your CSS otherwise it will become broken. The only thing that changes if we add/remove `right` class in this case is an icon near the cost, so in my humble opinion such class can be settled deeper in the DOM tree.

src/definitions/elements/input.less
```diff
@@ -455,20 +455,34 @@
   margin: 0;
 }
 
-/* Button on Right */
+/* Input when ui Left*/
+.ui[class*="left action"].action.input > input {
+  border-top-left-radius: 0 !important;
+  border-bottom-left-radius: 0 !important;
+  border-left-color: transparent !important;
```
lubber-de: Please remove the ` !important` s. They are not needed (See my testcase)

src/definitions/elements/input.less
```diff
@@ -455,20 +455,34 @@
...
+}
+
+/* Input when ui Right*/
 .ui.action.input:not([class*="left action"]) > input {
   border-top-right-radius: 0 !important;
   border-bottom-right-radius: 0 !important;
   border-right-color: transparent !important;
```
lubber-de: Please also remove `!important`, also not needed

src/definitions/elements/input.less
```diff
@@ -477,27 +491,6 @@
...
-}
-.ui[class*="left action"].input > .dropdown:first-child,
-.ui[class*="left action"].input > .button:first-child,
-.ui[class*="left action"].input > .buttons:first-child > .button {
-  border-radius: @borderRadius 0 0 @borderRadius;
-}
-/* Input Focus */
-.ui[class*="left action"].input > input:focus {
-  border-left-color: @focusBorderColor !important;
-}
```
lubber-de: Don't delete this. It's needed. but remove `!important` and increase specificity by one additional `ui` instead, so it becomes

```css
/* Input Focus */
.ui.ui[class*="left action"].input > input:focus {
  border-left-color: @focusBorderColor;
}
```

See https://jsfiddle.net/4L7zvsbc/1/
Try to click into the very bottom input, so it gets focus.

|Your change|Old bevavior|
|-|-|
|image|image|

src/definitions/elements/input.less
```diff
@@ -455,20 +455,34 @@
   margin: 0;
 }
 
-/* Button on Right */
+/* Input when ui Left*/
+.ui[class*="left action"].action.input > input {
```
lubber-de: the additional `.action` can be removed ([class*="left action"] implies it already)
->
```css
.ui[class*="left action"].input > input {
```

RShalman:
> Almost done 😉
> Please see my comments.
> I implemented your changes together with my requested changes here
> https://jsfiddle.net/7gf135dc/ to see it is working

Committed those requested changes :)

src/definitions/elements/input.less
```diff
@@ -455,47 +455,40 @@
...
+.ui.action.input > .buttons:last-child > .button {
+  border-radius: 0 @borderRadius @borderRadius 0;
 }
+
 /* Input Focus */
-.ui[class*="left action"].input > input:focus {
-  border-left-color: @focusBorderColor !important;
+.ui.ui.action.input:not([class*="left action"]) > input:focus {
+  border-right-color: @focusBorderColor;
 }
```
lubber-de: Sorry, but that was the wrong code line (although its good, you removed the !Important there aswell)


There is an additional `/* Input Focus */` existing, which (additional `.ui` only need for the second declaration!
In the end we need to have 
```css
/* Input Focus */

.ui.action.input:not([class*="left action"]) > input:focus {
  border-right-color: #85B7D9;
}
.ui.ui[class*="left action"].input > input:focus {
  border-left-color: #85B7D9;
}
```

RShalman: Committed that.

RShalman:
> LGTM, thanks for your patience 🙂

No problem! Was happy to help the project :)

hammy2899:
@all-contributors please add @RShalman for code

PR #1074, opened 2019-10-07
[Rating] Render partial/non-integer ratings

### LINKED ISSUE #554: Rating: display average ratings
# Feature Request
display average ratings

## Description
Would be nice if there where some thoughts like this:
https://github.com/Semantic-Org/Semantic-UI/issues/1048#issuecomment-403869204

zombietfk:
## Description
Render partial ratings when given a decimal value

Had to do some weird stuff to get this working guys and I've tried and tried to get it working on edge/ie but no success 🤷‍♀ but previous functionality is at least retained in those environments, let me know if you have any ideas

## Testcase
https://jsfiddle.net/pqkavxrd

## Closes
#554


hammy2899:
This won't work in IE as mentioned because of the use of CSS variables. We could get around this by applying the background gradient as a direct style but that seems too hacky. I think if we leave this and add a message on the docs saying it doesn't work in IE it should be fine unless anyone has any other way to fix it.

hammy2899:
@zombietfk you still want to work on this?

zombietfk:
@hammy2899 sorry, been a bit busy recently, I will pick this up again this week

zombietfk:
Discovered jquery only supports css variables in ^3.2.0, I'll need [this ](https://github.com/fomantic/Fomantic-UI-Docs/pull/164) in before I make further changes!

lubber-de:
It's also not working in firefox because of the -webkit settings
image

src/definitions/modules/rating.less
```diff
@@ -69,14 +69,23 @@
   color: @activeColor;
 }
 
+/* Partially Active Icon */
+.ui.rating .icon.partial.active {
+  background: linear-gradient(to right, @activeColor 0% var(--full), @inactiveColor var(--full) 100%);
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: transparent;
```
exoego: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke

According to MDN, those webkit features may be a subject to change.
I think web-standard features should be used.

zombietfk: Hi, thanks for taking the time to review my PR. I have updated the CSS to remove all non-standard props other than `-webkit-text-stroke`. Using `text-shadow` unfortunately won't seem to work due to the alpha value of the text color of non-active icons, so it shows through 😕

I have created an additional jsfiddle with text-stroke removed in order to show the effect of removing `-webkit-text-stroke`. 
image
https://jsfiddle.net/pqkavxrd
image
https://jsfiddle.net/pqkavxrd/1/

While I agree it's not ideal to use CSS features which are currently non-standard, I would like to point out that it is not without precedent in other components, i.e.

`font-smooth` in the flag and icon elements
`-webkit-tap-highlight-color` in search module and button element
`-webkit-overflow-scrolling` in the dropdown module

zombietfk:
@lubber-de I believe I have resolved the firefox compatibility issue and have updated the [linked fiddle](https://jsfiddle.net/pqkavxrd)

hammy2899:
@all-contributors please add @zombietfk for code

hammy2899:
@all-contributors please add @zombietfk for docs

hammy2899:
@all-contributors please add @RenonDis for docs

(from docs repo not this PR)

hammy2899:
@all-contributors please add @aexvir for docs

(from docs repo not this PR)

hammy2899:
@all-contributors please add @josejbarneto for docs

(from docs repo not this PR)

hammy2899:
@all-contributors please add @aardbol for docs

(from docs repo not this PR)

hammy2899:
@all-contributors please add @HypertextPP for docs

(from docs repo not this PR)

hammy2899:
@all-contributors please add @ryamaguchi0220 for docs

(from docs repo not this PR)

PR #1082, opened 2019-10-10
[Slider] preventCrossover setting for range sliders

### LINKED ISSUE #1021: Prevent slider from swapping handles
# Feature Request

Slider should have option to  block "max handle" when trying to go below "min handle"

## Description
The slider handles can be reversed, so if i drag max handle to the left of min handle, values get inverted.

There should be option in settings to prevent this.

## Example
ui-slider-bug

ColinFrick:
## Description
Adds a new setting called `preventCrossover` which prevents the lower thumb to crossover the upper thumb.

This is set to `true` by default, making this a breaking change.

Does not prevent calling `$("#slider").slider("set rangeValue", 50, 25);`

## Testcase
https://jsfiddle.net/nb4x7ukm/2/

## Screenshot (when possible)
Peek 2019-10-10 10-40

## Closes
#1021

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -658,6 +658,9 @@ $.fn.slider = function(parameters) {
               secondThumbPos = parseFloat(module.determine.thumbPos($secondThumb)),
               secondThumbDelta = Math.abs(eventPos - secondThumbPos)
             ;
+            if(thumbDelta === secondThumbDelta && module.get.thumbValue() === module.get.min()) {
+              return $secondThumb;
+            }
             return thumbDelta <= secondThumbDelta ? $thumb : $secondThumb;
           },
           closestThumbPos: function(eventPos) {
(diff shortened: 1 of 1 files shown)


ColinFrick:
@prudho True I'll change it accordingly

ColinFrick:
@prudho I'll changed it, so that - if both handles are on the lowest value - the second handle is always focused when dragging. This is not tied to the preventCrossover setting

ColinFrick:
@prudho currently the active handle is determined as soon as you press the mouse button.
 I'll investigate if that could be changed, f.e. keeping it in an undetermined state until the mouse is moved.

ColinFrick:
@prudho I changed the PR accordingly.

An alternative to this would be that we change the active thumb on the fly - as soon as it'd pass the other thumb. Would that be better?

hammy2899:
@fomantic/core you can all just re-approve I just fixed merge conflicts

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/162

PR #1084, opened 2019-10-10
[Calendar] Popup arrow color doesn't match container background

### LINKED ISSUE #744: [Calendar] Popup :before element color doesn't match container background
# Bug Report

## Expected result
Popup's `:before` pseudo-element should match popup color

## Actual result
Popup's `:before` pseudo-element doesn't match popup background color when opening down.

One solution I can think of is to apply a temporary style to the calendar's div when it is opening down. The correct color looks like `#f9fafb`.

## Testcase
https://jsfiddle.net/nfc9gxqu/

## Screenshot
image

## Version
2.7.4

ColinFrick:
## Description
Added a new header background color variable to calendar, which is used for the table header and the popup arrow. Timepicker are excluded, because they don't have a header in the popup.

## Testcase
https://jsfiddle.net/72q8Lxrw/

## Closes
#744

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -137,7 +137,7 @@ $.fn.calendar = function(parameters) {
                   domPositionFunction = $activatorParent.closest(selector.append).length !== 0 ? 'appendTo' : 'prependTo';
               $container = $('<div/>').addClass(className.popup)[domPositionFunction]($activatorParent);
             }
-            $container.addClass(className.calendar);
+            $container.addClass(className.calendar).addClass(settings.type);
             var onVisible = settings.onVisible;
             var onHidden = settings.onHidden;
             if (!$input.length) {
--- a/src/definitions/modules/calendar.less
+++ b/src/definitions/modules/calendar.less
@@ -28,6 +28,10 @@
   user-select: none;
 }
 
+.ui.calendar .ui.popup:before {
+  background: @headerBackground;
+}
+
 /*******************************
             Calendar
 *******************************/
@@ -131,6 +135,11 @@
   color: @adjacentTextColor;
   background: @adjacentBackground;
 }
+
+.ui.calendar .ui.table > thead > tr > th {
+  background: @headerBackground;
+}
+
 
 /*--------------
      States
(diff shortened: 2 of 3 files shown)


lubber-de:
Depending on the PoupPosition or existing selected cell, the color had to be adjusted to fix this perfectly:

#### top left/right
- arrow should be white again (slightly recognizable with the default colors though)
image

#### If a time was already selected 
- Arrow should be same cell selection color, beware of different possible in combination with selected cell (top left/right, bottom left/right) arrow position as well here
image

ColinFrick:
The more I look at this, the more I think we should either leave it like it is now (handling all eventualities in the css seems impossible) or change the arrow to a triangle element (instead of the rotate rectangle it is now).
The arrow would still have a different color to the element, but wouldn't overlay it any more thus mitigating the problem.

ColinFrick:
I played around with the css a little, problem with using a triangle is that it'd require both :before and :after to add a border to it. Because popups already uses :after for tooltip positioning we can't go that route.
What I tried was to use clip-path to hide the part that is rendered over the popup: https://jsfiddle.net/w5cduzpy/ (only top / bottom popups work here).
No support for IE11 and Edge: https://jsfiddle.net/w5cduzpy/

Any suggestions how to proceed @fomantic/core ?

lubber-de:
@ColinFrick 🤔  Adjusting the general bottom and possible left/right center arrows was easy, but some JS is needed whenever a date is selected to find out if one of the corner cells has been selected and give the parent calendar node some extra class, so the popup arrow can be exactly selected via CSS.
I prepared a draft here, it's not perfect, but you will get the base idea.
https://jsfiddle.net/hp1k4r7f/3/
image

ko2in:
I think we don't need JS to add extra class to achieve this. We can simply adopt the same method from popup for the calendar and add the arrow to cover the popup arrow with specific background color.

I created a fiddle to illustrate this. Here's the fiddle (https://jsfiddle.net/cLfbjy0a/). Works for both top and bottom popups with left arrow. I've not yet tested for right and the center.

lubber-de:
@ko2in 
Very nice approach. However, some situations still have to be fetched, like
image

lubber-de:
@ko2in Sorry, my bad, ignore my previous post (which i deleted already...:wink:)

PR #1098, opened 2019-10-16
feat(form): generalize state variations for field

### LINKED ISSUE #1097: [Form] More field states
# Feature Request

## Description

`ui form` supports several states: `loading`, `success`, `error` and `warning`.

Form's child element `field` supports only [`error` state](https://fomantic-ui.com/collections/form.html#field-error).
It would be great if `field` also support `warning` state, so warning-level check (can be submited but not recommended) can be visually expresed.

I don't think `loading` and `success` arenecessary for `field`, since it is unlikely to have success/loading in single field.

exoego:
## Description

* Add `info` state variation to `form` element (all other 3 are already implemented).
* Add `info`, `success` and `warning` state variations to...
    * `field` element
    * `input` element
    * `dropdown` element
* Show/hide `.ui.message.{STATE}` depending on field-state (addresses #1124)

TODO:
- [x] Define central mapping to iterate
- [x] Set default colors for each states 
- [x] Update docs in https://github.com/fomantic/Fomantic-UI-Docs/pull/160

## Screenshot (when possible)
image

## Closes
#1097
#1124


src/definitions/collections/form.less
```diff
@@ -448,196 +414,214 @@
...
+  .ui.form .field.@{state} input[type="datetime-local"],
+  .ui.form .field.@{state} input[type="email"],
+  .ui.form .field.@{state} input[type="number"],
+  .ui.form .field.@{state} input[type="password"],
+  .ui.form .field.@{state} input[type="search"],
+  .ui.form .field.@{state} input[type="tel"],
+  .ui.form .field.@{state} input[type="time"],
+  .ui.form .field.@{state} input[type="text"],
+  .ui.form .field.@{state} input[type="file"],
+  .ui.form .field.@{state} input[type="url"] {
```
lubber-de: :thinking: i wonder if we could reduce all the the specific input selectors everywhere to

```css
.ui.form .field.@{state} input:not([type="checkbox"]):not(type="radio"),
```

?

exoego: I would like to keep this PR focusing on adding state variations, to avoid regressions.
Opened up an issue for refactoring them.

I think `hidden`, `button`, `file` and `submit` should be excluded using `:not` too.
However, I am afraid that performance impact of many `:not` selector, especially on low-perf devices.
https://www.sitepoint.com/optimizing-css-id-selectors-and-other-myths/
https://dev.to/gmartigny/comment/8h17

src/definitions/collections/form.less
```diff
@@ -448,196 +414,227 @@
...
 /*--------------------
-        Success
+        States
 ---------------------*/
+each(@formStates, {
+  @state: replace(@key, '@', '');
+  @c: @formStates[@@state][color];
+  @bg: @formStates[@@state][background];
+  @bdc:  @formStates[@@state][borderColor];
+  //@l: @colors[@@color][light];
```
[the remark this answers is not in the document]
exoego: Oops.
Removed in https://github.com/fomantic/Fomantic-UI/pull/1098/commits/8451e255671bb5cf640544fbd0eb9c87ac1d2a91

PR #1100, opened 2019-10-17
fix(dropdown): use regular event handler for allowAdditions input blur

### LINKED ISSUE #997: [Dropdown] Escaped element gets duplicated
# Bug Report

When I add a new element with characters which will be html escaped, it wil get duplicated under certain conditions.

I think this is intruduced with: https://github.com/fomantic/Fomantic-UI/commit/f7c24ace480ef0f98350d8192caff44f42adf0a9

## Steps to reproduce
1. Open https://jsfiddle.net/yzvmobrw/1/
2. Add the following value: <div>
3. Press enter
4. Click in the white area under the element

## Expected result
The new element should be added.

## Actual result
The new element will be added twice.

## Testcase
https://jsfiddle.net/yzvmobrw/1/

## Screenshot (when possible)
dropdown

## Version
2.7.8

aexvir:
Closes https://github.com/fomantic/Fomantic-UI/issues/997

## Description
The blur event for dropdowns that allowed additions wasn't calling the regular event handler, and it was bypassing some of the preprocessing, checks and side effects, like sanitization or flushing the values.

## Testcase
broken: https://jsfiddle.net/Lxv71m8k/
fixed: https://jsfiddle.net/cyL5r09k/

## Closes
#997

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -981,17 +981,12 @@ $.fn.dropdown = function(parameters) {
               : $activeItem,
             hasSelected = ($selectedItem.length > 0)
           ;
-          if(hasSelected && !module.is.multiple()) {
+          if(settings.allowAdditions || (hasSelected && !module.is.multiple())) {
             module.debug('Forcing partial selection to selected item', $selectedItem);
             module.event.item.click.call($selectedItem, {}, true);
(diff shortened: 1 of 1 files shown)


prudho:
Seems good to me too, but I wonder why this case was bypassed before.

PR #1101, opened 2019-10-17
feat(dropdown): add case sensitivity setting for search

### LINKED ISSUE #932: [Dropdown] ignoreCase does not change text search behavior
# Bug Report

## Steps to reproduce
1. Focus Dropdown
2. Type alabama on search box

## Expected result
No option is to be suggested, since search is case-insensitive only with `ignoreCase: on` .

## Actual result
Search is case-insensitive even with `ignoreCase` defaulting to off, Alabama option is suggested.

## Testcase
https://jsfiddle.net/9nymb3g1/2/

## Version
2.7.7

aexvir:
## Description
`ignoreCase` was only used for affecting the label creation. ~This MR makes the filtering respect that setting, for both searches, on text and value.~

Instead or reusing the same variable, as that would introduce a breaking change where none of the current users will have case insensitive search, this MR adds a new setting that allows setting the case sensitivity for the item filtering

## Testcase
before: https://jsfiddle.net/3we8nb7j/
after: https://jsfiddle.net/ye1hg6c4/

## Closes
#932

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -832,7 +832,8 @@ $.fn.dropdown = function(parameters) {
             ),
             results          =  null,
             escapedTerm      = module.escape.string(searchTerm),
-            beginsWithRegExp = new RegExp('^' + escapedTerm, 'igm')
+            regExpFlags      = (settings.ignoreSearchCase ? 'i' : '') + 'gm',
+            beginsWithRegExp = new RegExp('^' + escapedTerm, regExpFlags)
           ;
           // avoid loop if we're matching nothing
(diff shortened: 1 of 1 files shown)


hammy2899:
I agree with @lubber-de, keeping `ignoreCase` as backwards compatible is the way to go because changing its behavior can't be justified for the impact it would have on users. Also adding this setting allows the user more control over the search case.

aexvir:
Yep, totally agree

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/161

hammy2899:
@all-contributors please add @aexvir for code

Issue #341, opened 2019-01-01
Commit message format seems required but documented as optional


exoego:
# Help Wanted

## Problem
Commitlint bot marks PR ❌ if commit message does not follow commit message format.
![image](https://user-images.githubusercontent.com/127635/50572998-8a0dc500-0e0e-11e9-9263-e9955da1755b.png)

However, the commit message format is documented as not required.
> https://github.com/fomantic/Fomantic-UI/blob/7f6bd29db53f17a2c4e2c47e2dc0e81568d1ea5f/CONTRIBUTING.md#pencil-pull-requests-guide
> - Please commit using the [AngularJS Git Commit Message Convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153) this is more of a nice to have and isn't required.

I initially did not know what is valid `type` for commit message...

I agree that enforcing message format is required to make history useful, 
but the CONTRIBUTING.md should be updated. 
It is good to include commit message format and examples.

hammy2899:
The bot which checks your PR commits doesn't block merging and is just a check mainly for us maintainers.

We might want to consider self hosting the bot and customizing who's PR's it checks etc.

PR #1110, opened 2019-10-19
chore(ci): Add job to enforce commit message convention

aexvir:
## Description

Seeing that fomantic ui requests following the Angular git commit message convention, which is precisely what was used as base for the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) proposal, I wanted to suggest using a CI tool for making sure that the commits follow the convention.

`commitsar` is a small go binary that does just that, I personally know and work with the developer, I don't think it matters but I just don't want to hide it. I think it could be a good and lightweight solution for https://github.com/fomantic/Fomantic-UI/issues/341

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,3 +19,11 @@ jobs:
       run: npm install --ignore-scripts
     - name: pre fomantic install & gulp build
       run: npx gulp install
+  commitsar:
+    name: Verify commit messages
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v1
+      - name: Run commitsar
+        uses: docker://commitsar/commitsar:0.8.0


hammy2899:
We had something like this in the past and like you mentioned this issue https://github.com/fomantic/Fomantic-UI/issues/341 was created. I don't think we need to lint commit messages because it adds confusion to outside contributors and to be honest we don't care the format of a commit message because we will just format it correctly when we merge it.

That being said I think we should just leave this to make contributing easy for any new contributors.

@fomantic/core whats your 2 cents?

lubber-de:
I agree with @hammy2899

aexvir:
Actually, thanks to this PR we found some small issue in the branch targeted on GitHub actions. Fixed now 🙂

hammy2899:
I'm going to close this because the last time we had something like this it caused a few issues with some people and I don't want to increase the complexity for new contributors.

@aexvir This is a good PR however as I said above this does add confusion for new contributors and we can't afford that in the current state of FUI. (maybe for v3 😉)

PR #1112, opened 2019-10-19
fix(dropdown): value on dynamically loaded select based dropdowns

### LINKED ISSUE #927: [Dropdown] 'values' and 'get value'
hi
when i use `dropdown` with `values` option, `get values` method of `form` return `null` for `dropdown`!

see : http://jsfiddle.net/2m6b53dk/

fomantic-ui : 2.7.7

aexvir:
## Description
When HTML selects don't have `option` child elements we can't get or set its value. Fomantic allows making an empty select and passing the data on `dropdown`'s init, but those values never get added to the `select`, so, when trying to get the value, it returns `null`.

This MR solves that by "persisting" the values that get changed via Fomantic's `dropdown` module to the DOM, as select options, allowing to interact with the select's value.

## Testcase
broken: http://jsfiddle.net/bh162r9p/
fixed: http://jsfiddle.net/mq5L71ne/

## Closes
#927


src/definitions/modules/dropdown.js
```diff
@@ -1007,6 +1007,15 @@ $.fn.dropdown = function(parameters) {
...
             });
+
+            if($input.is('select')) {
+              $input.html('');
+              $input.append('<option disabled selected value></option>');
+              $.each(values, function(index, item) {
+                $input.append('<option value="' + item.value + '">' + item.name + '</option>');
+              });
+              module.refresh();
+            }
```
lubber-de: Please change the code to the following because of:
- Use existing functions
- avoid code recursion when the selectobserver is still running (which would unnecessarily recreate the menu each time you append an option to the select) 
- Make sure a probably omitted  `name` attribut is fetched
- sanitize the given values
- The refresh at the end is not necessary , because $menu is not changed (because we stopped the observer before)
```javascript
            if(module.has.selectInput()) {
              module.disconnect.selectObserver();
              $input.html('');
              $input.append('<option disabled selected value></option>');
              $.each(values, function(index, item) {
                $input.append('<option value="' + settings.templates.deQuote(item[fields.value]) + '">' + settings.templates.escape(item[fields.name] || item[fields.value], settings.preserveHTML) + '</option>');
              });
              module.observe.select();
            }
```
Please also adjust one line in the `templates.menu` code:
https://github.com/fomantic/Fomantic-UI/blob/8151ae62b6729dd441433a049b2311e4ccff03d7/src/definitions/modules/dropdown.js#L4145
to 
```javascript
html +=   escape(option[fields.name] || option[fields.value],preserveHTML);
```
in case the name attribute is not given in the `change values` behavior

aexvir: Thanks for the review @lubber-de !

I was actually looking for a function that would return if the input was select, I even searched for `is('select')` in the code, but I guess I missed it somehow 😅 

I didn't knew much about the observer logic, I'll take a deeper look at it the next time

And you are right about sanitization and missing `name` attribute

Issue #566, opened 2019-03-12
[toast]Toast with Actions - approve, deny 


apple-idn:
# Feature Request
on Actions - Approve and deny and right image(Whatever it is)
## Description
feature same OSX Notification Center with active Approve and deny

## Example
![image](https://user-images.githubusercontent.com/33564576/54219315-403a0d00-452a-11e9-8359-15afabbdb1af.png)

lubber-de:
Duplicate of #160

lubber-de:
Possible by #1129

Issue #316, opened 2018-12-18
Toast usability


romaninsh:
# Usability report - Toast

I have interviewed few people regarding the "Toast" to collect the following usability feedback. Hopefully this can inspire someone to further improve "Toast".

(I compared with MacOS notifier, see "terminal-notifier")

# 1. Copy text from the toast.

In some cases user may want to interact with a toast - take a screenshot of it or copy text. Developer is faced with a choice - hide toast automatically or make it sit there forever. There is no middle option.

PROBLEM: toast can disappear preemptively

FIX: Allow user to prevent toast from disappearing by hovering mouse over it.

FIX: Set default timeout based on the length of title/message, so longer messages would be displayed longer.

# 2. Attention impact and readability

Thinking on how toast is used: It captures user's attention when it appears and user scans the message. If message is not important, user will re-focus back on the page.

PROBLEM: toast perform more animation when disappearing rather than when appearing.

FIX: Run animation when toast appears. Use fade-out when disappearing.

# 3. Interrupted and difficulty to follow

Imagine user reads a toast. Two events can occur. Either new toast appear or another toast disappear. Which of those should encourage user to "stop" reading and re-focus?

PROBLEM: toast disappearing is more disruptive than appearance of new toast due to repositioning of the text. 

FIX: new toasts should appear next to the edge of the screen and move existing toasts. Old toasts should fade-out subliminally.

# 4. Text selection

Attempt to select text for copy-paste from a toast makes it disappear (even if it's not timed).

FIX: click-drag-release should not hide toast. 

# 4. Shadow

Toast appears on a layer ABOVE the interface. All other UI elements that appear on top feature shadow in the default theme.

FIX: toast should also cast shadow.

# 5. Bouncy effect

User may continue reading message while it slides. Bouncy effect makes it very hard to follow. Using ease-in-ease-out slide is much more easy on the eyes.

FIX: replace animation with smooth reposition

# 6. Default progress

When progress goes forward it indicates that some work is being performed. For example when you are installing update, progress bar moves forward. Typically user interaction will cancel action (e.g. copying files).

When action is "paused" such as waiting 30 seconds before reboot or displaying dialog that will disappear, progress goes backwards (count-down seconds or reverse progress bar). This should be recommended and default action.

FIX: change default

# 7. Icon positioning

It looks like icon is not properly aligned with the text. 

FIX: position icon in the middle vertically

# 8. Title

[...]

Issue #315, opened 2018-12-18
Toast needs to have shadow.


romaninsh:
# Bug Report

Shadow is used by various elements, which are "floating", e.g.:

<img width="309" alt="image" src="https://user-images.githubusercontent.com/453929/50170047-e937e980-02e6-11e9-8a2a-4cf9b776fac2.png">

However toast does not have shadow. 

<img width="367" alt="image" src="https://user-images.githubusercontent.com/453929/50170069-f3f27e80-02e6-11e9-9200-8582f22be9a7.png">

It's reasonable to expect it to be styled consistently to tooltips and pop-overs.

## Steps to reproduce
1. https://fomantic-ui.com/modules/toast.html
2. Run Code

## Version
2.6.4

romaninsh:
I wrote a more detailed analysis on #316.

lubber-de:
Implemented by #1129

Issue #280, opened 2018-12-02
[Toast] Build from html 


hakeem86:
## Description
It would be great if it was possible to build toast from HTML.. Would be useful when you have POST some data to the server and you want to show an error/info toast message due to user input error etc. 

## Example
Maybe something like this or another syntax that make more sense. 
>     <div class="toast success compact visible" data-position="top left" data-displayTime="5000">
>         <div class="icon info ui toast">
>             <i class="info icon"></i>
>             <div class="content">
>                 <div>Hello world</div>
>             </div>
>         </div>
>     </div>

> $('.toast').toast();

lubber-de:
We are already working on this feature :wink: Stay tuned.

lubber-de:
Finally possible by #1129

PR #1129, opened 2019-10-28
feat(toast): add actions, behaviors and other variations

### LINKED ISSUE #160: [Toast] Action Text
# Feature Request

## Description
Add the option of allowing action text (like when sending/deleting an email in Gmail) to the toast prompt with a corresponding callback for when this part is specify clicked (eg. onActionClick)

An alternative could potentially allow action buttons like a MacOS notification, with primary/secondary button callbacks (eg. onButtonPrimary), however, on the other hand, buttons may not be quite suited for a toast

## Example
Action Text:
https://www.howtogeek.com/wp-content/uploads/2016/12/00_lead_image_undo_send.png

Action Button(s):
https://www.jamf.com/jamf-nation/discussions/26683/macos-notification-prompts

lubber-de:
## Description
This PR adds a **lot** of new and often requested features to the toast component. 
> See it as a special Halloween gift contribution because i am celebrating my _**one year membership**_ 🎉 as a Fomantic UI maintainer 😃 🚀 

#### New Features
- Image support just like icon
- displayTime 'auto' calculates time by amount of words
- Progressbars can have its own individual color now 
- A toast can be created out of existing dom nodes or markup now $('.ui.toast').toast() with clone support;
- Toast is pausable on hovering to be able to mark/copy text or just hold on to have more time to read it
- closeIcon can be displayed to the left 
- Support for `card` and `message` layouts

#### Behaviors
- get or check for existing toasts
- stop/start display timer
- get the remaining Time

#### Actions
- a toast can have a configurable amount of action buttons
- support for default button selectors as in modal
- displayable at the bottom of the toast as in modal or full width attached to the modal
- also displayable to the right

### Changed defaults (aka breaking changes)
- Toast is floating (aka has a shadow)
- Progressbar decreases
- Close-Easing is smoother instead of bouncy (still available for those who like it)
- Icon is centered vertically
- Toast display in neutral white background without icon (looks good, as it has a shadow now)
- A possible given title is a bit larger than before (by using standard "ui header")
- floating and hoverfloating moved from toast to toast-box to cover the whole toast including possible attached actions

### Core
- Changed progressbar from css transition to css animation to be able to pause scrollbar on hover via css only
- Return instance of created toast even when created dynamically via body so you can interact with and reference it via javascript afterwards
- Support behaviors because of instantiation
- Optimized LESS Code usage, refactored JS 

## Testcase
Take some ☕️ and have fun playing with the following playground (you may need to scroll down to see all trigger buttons)
### Toast
https://jsfiddle.net/dw5Lon4t/1/
https://jsfiddle.net/dw5Lon4t/1/show (play fullscreen)

### Card + Message
https://jsfiddle.net/vm9xgjpu/
https://jsfiddle.net/vm9xgjpu/show (play fullscreen)

#### New smooth fadeout and default `neutral` white toast with shadows
toast2fadeout
### Pausable progress on hover
toast2pause
### Image support (`tiny`,`mini` and `avatar` sizes are supported)
image
### Actions with tons of possibilities 🙂 
image
image
image
### Card 
image
image
image
image
### Message
image
image

## Closes
#160 
#280 
#315 
#316 
#566


src/definitions/modules/toast.js
```diff
@@ -119,91 +141,198 @@ $.fn.toast = function(parameters) {
...
+            if(settings.closeOnClick || settings.closeIcon) {
+              (settings.closeIcon ? $close : $toast)
+                  .on('click' + eventNamespace, module.event.click)
+              ;
+            }
+            if($animationObject) {
+              $animationObject.on('animationend' + eventNamespace, module.close);
+            }
+            $toastBox
+              .on('click' + eventNamespace, selector.approve, module.event.approve)
```
prudho: Shouldn't you enable these events only if toast has action ?

lubber-de: The `selector.approve` makes sure the events will only created when appropriate action buttons exist.
I used the same code as for modal.

src/definitions/modules/toast.js
```diff
@@ -47,68 +47,90 @@ $.fn.toast = function(parameters) {
...
+        $actions,
+        $progress,
+        $progressBar,
+        $animationObject,
+        $close,
         $context        = (settings.context)
           ? $(settings.context)
           : $('body'),
 
+        isComponent     = $module.hasClass('toast') || $module.hasClass('message'),
```
exoego: Why it is needed to distinguish `toast` from `message` ?

`isToastComponent` may be a better name to express intent.

lubber-de: Because the styling of the message component was supported before as an alternative to style toasts (and needed some css adjustments)
See https://fomantic-ui.com/modules/toast.html#use-message-style

lubber-de: @exoego  changed the variable Name to `isToastComponent` as suggested while also adding support for card layout as well 😃 by 2fe990b8eb8c07c229227e3b2670eb934f1ab872

lubber-de:
I added support for `message` and `card` layout. Changed description, jsfiddles and screeshots
Try this https://jsfiddle.net/86xgthyp/show

jamessampford:
Can the attached buttons be equally spaced out, as an option, just to look tidier? Or is it just the way the fiddle was made as a sample?

lubber-de:
> Can the attached buttons be equally spaced out, as an option, just to look tidier? Or is it just the way the fiddle was made as a sample?

Yes, it's possible, just provide the number of actions to the `classActions` setting. Attached actions are usual [button groups](https://fomantic-ui.com/elements/button.html#buttons) so they already support the technique of [equal width](https://fomantic-ui.com/elements/button.html#equal-width) 🙂 

See https://jsfiddle.net/m0fLasr9/

### Normal
using `classActions: 'attached'`
image

### Equal Width
Suppose you have three action buttons then use `classActions: 'three attached'`
image

hammy2899:
I think on this toast (clicking does not close) the cursor shouldn't be a pointer because the toast has no click action 
image

Although this might be difficult to fix so we might have to just leave this and also the toast may have a callback action. Let me know what you think @fomantic/maintainers

lubber-de:
@hammy2899 Adjusted the cursor as suggested 😃 
See https://jsfiddle.net/a68kdy3v/

btw: Callback functions are already there 😉:
 `onShow`, `onHide`, `onHidden`

hammy2899:
@lubber-de Nice, il give it another review when I'm at a PC 👍

PR #1143, opened 2019-11-02
fix(slider): offset for reversed vertical slider

### LINKED ISSUE #1142: [Slider] Reversed Vertical Slider's thumb is offset
# Bug Report

## Expected result
Reversed vertical slider thumb should line up with slider

## Actual result
It doesn't

## Testcase
https://jsfiddle.net/82gy40a5/

## Screenshot (when possible)

image

## Version
2.7.8

## Note
This is fixed by removing the following rule:

```css
.ui.reversed.slider .inner .thumb {
    left: auto;
}
```

A solution could be to make the reversed slider thumb style not apply to vertical ones with the selector `.ui.reversed.slider:not(.vertical) .inner .thumb`

GammaGames:
# Description
Fixed reversed vertical slider's thumb, it used to be offset.

## Testcase
https://jsfiddle.net/rt4zqeak/

## Screenshot (when possible)
image

## Closes
#1142

### DIFF
--- a/src/definitions/modules/slider.less
+++ b/src/definitions/modules/slider.less
@@ -116,7 +116,7 @@
 }
 
 .ui.reversed.vertical.slider .inner .thumb {
-  left: @verticalSliderThumbOffset;
+  left: @thumbVerticalSliderOffset;
 }
 
 .ui.labeled.reversed.slider > .labels .label {
--- a/src/definitions/modules/slider.less
+++ b/src/definitions/modules/slider.less
@@ -116,8 +116,8 @@
 }
 
 .ui.reversed.vertical.slider .inner .thumb {
-    left: 0.03em;
-  }
+  left: @verticalSliderThumbOffset;
+}
 
 .ui.labeled.reversed.slider > .labels .label {
   transform: translate(-100%, -100%);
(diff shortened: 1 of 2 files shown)


GammaGames:
@lubber-de Like now? I don't have my windows computer handy to test atm

src/definitions/modules/slider.less
```diff
@@ -110,11 +110,15 @@
...
 }
 
-.ui.reversed.slider .inner .thumb {
+.ui.reversed.slider:not(.vertical) .inner .thumb {
   left: auto;
   right: 0;
 }
 
+.ui.reversed.vertical.slider .inner .thumb {
+    left: 0.03em;
```
lubber-de: Please use a variable like `@verticalSliderThumbOffset` and define it in slider.variables instead of hardcoding it here

src/definitions/modules/slider.less
```diff
@@ -110,11 +110,15 @@
...
+  }
```
lubber-de: Please fix spacing so it matches the selector position 🙂

lubber-de:
Basically yes, this works . Please use a variable instead and slightly correct the closing brace spacing and this is done from my perspective 😁👍🏻

src/definitions/modules/slider.less
GammaGames: I wonder how that happened! I also fixed the indentation to 2 spaces, my bad

PR #1148, opened 2019-11-05
feat(icons): update icons to FA 5.11.2

hammy2899:
## Description
Update icons to FontAwesome 5.11.2

@lubber-de I have made the arrow down change which is required 😉


mkodrats:
LGTM

PR #1149, opened 2019-11-05
[Slider] Fix touch events handling when jquery 2 is used 

### LINKED ISSUE #1147: Keep getting javascript error on mobile - Slider
# I'm getting javascript error while I'm using the fomantic UI with WordPress.

## Javascript error on slider's node movement in mobile.

hi, I tried to create a slider using fomantic-ui and this works fine on desktop but throws an error on mobile view in chrome.

as far as I know, this works well even in mobile but as I fused this to WordPress site it keeps throwing an error in the mobile view.

here is live site link : [Demo](http://101.53.136.134/demo/jewellery/product-category/diamond/?EO_WBC=1&BEGIN=diamond&STEP=1)

in-order to get an error toggle to mobile view and click on filter then carat and you can see a slider but when you try to change the slider's nodes position it will throw the following error in console.

```
Uncaught TypeError: Cannot read property '0' of undefined
    at Object.eventPos (semantic.min.js?ver=5.1.3:16114)
    at Object.down (semantic.min.js?ver=5.1.3:15703)
    at HTMLDivElement.<anonymous> (semantic.min.js?ver=5.1.3:15653)
    at HTMLDivElement.dispatch (jquery.js?ver=1.12.4:3)
    at HTMLDivElement.r.handle (jquery.js?ver=1.12.4:3)
```

lubber-de:
## Description
When jquery 2 is used together with FUI the `changedTouches` or `touches` property was not attached to the event object, but still exists at the `event.originalEvent` Object.
Slider was always expecting it at the event object which results in a JS error when trying to move the slider using touch devices.
This is especially the case when trying to simulate touch devices using chrome developer tools.

I also removed the completely unused second originalEvent parameter in all event handlers , just because it is never provided by any jquery version and always undefined. instead `originalEvent` is always provided as a child of `event` itself.

Why support jquery 2? Because this might be bundled with other packages like wordpress.

## Testcase
- Open any of the below testcases in chrome
- open developer tools 
- switch to device view
- probably reload the page 
- watch console

|jquery version|testcase|result|
|-|-|-|
|jquery 2|https://jsfiddle.net/xju4Lqhe/3/|breaks|
|jquery 2+this PR fix|https://jsfiddle.net/xju4Lqhe/2/|works|
|jquery 3|https://jsfiddle.net/xju4Lqhe/4/|works|

## Screenshot
slider_simulated_touch_bad

## Closes
#1147

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -317,11 +317,11 @@ $.fn.slider = function(parameters) {
         },
 
         event: {
-          down: function(event, originalEvent) {
+          down: function(event) {
             event.preventDefault();
             if(module.is.range()) {
               var
-                eventPos = module.determine.eventPos(event, originalEvent),
+                eventPos = module.determine.eventPos(event),
                 newPos = module.determine.pos(eventPos)
               ;
               // Special handling if range mode and both thumbs have the same value
@@ -336,12 +336,12 @@ $.fn.slider = function(parameters) {
               module.bind.slidingEvents();
             }
           },
-          move: function(event, originalEvent) {
+          move: function(event) {
             event.preventDefault();
-            var value = module.determine.valueFromEvent(event, originalEvent);
+            var value = module.determine.valueFromEvent(event);
             if($currThumb === undefined) {
               var
-                eventPos = module.determine.eventPos(event, originalEvent),
+                eventPos = module.determine.eventPos(event),
                 newPos = module.determine.pos(eventPos)
               ;
               $currThumb = initialPosition > newPos ? $thumb : $secondThumb;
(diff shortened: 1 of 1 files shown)


src/definitions/modules/slider.js
```diff
@@ -756,17 +756,19 @@ $.fn.slider = function(parameters) {
...
+                touches = touchEvent.changedTouches[0] ? touchEvent.changedTouches : touchEvent.touches,
+                touchY = touches[0].pageY,
+                touchX = touches[0].pageX
               ;
               return module.is.vertical() ? touchY : touchX;
             }
             var
-              clickY = event.pageY || originalEvent.pageY,
-              clickX = event.pageX || originalEvent.pageX
+              clickY = event.pageY || event.originalEvent.pageY,
```
exoego: Can we assume `originalEvent` always exists ?
If not, I am afraid that Null Pointer Exception.

lubber-de: Yes, we can. pageX/pageY should always be available  in a non jquery event. Only the fallback is to use event.originalEvent which should only happen when jquery is used...and FUI still needs jquery anyway to make the slider work.
As of https://api.jquery.com/category/events/event-object/ we can rely on the event.originalEvent existance

exoego: Understood 👍

PR #1159, opened 2019-11-12
[List] Fix horizontal divided list first item extra padding

### LINKED ISSUE #1158: [List] Horizontal divided list has an extra left padding for its first child
# Bug Report
When using a list that is `horizontal` and `divided` the first child has `0.5em` on left padding, and the whole list moves to the right that looks a bit weird.

## Steps to reproduce
Simply make this list
```
<div class="ui container">
  <div class="ui divided horizontal list" style="border: 1px solid red">
    <div class="item">
      <div class="content">
        Item1
      </div>
    </div>
    <div class="item">
      <div class="content">
        Item2
      </div>
    </div>
    <div class="item">
      <div class="content">
        Item3
      </div>
    </div>
  </div>
</div>
```
There are related css classes that apply to the items:
```
.ui.divided.horizontal.list > .item:not(:last-child) {
  padding-right: 0.5em;
}
```
and
```
.ui.divided.horizontal.list > .item {
  padding-left: 0.5em;
}
```

## Expected result
Same as just `horizontal` list the `horizontal divided` list should have no extra padding on the left.

## Actual result
When list has `divided` class it's first child has extra `padding-left: 0.5em` style.
The list looks shifted to the right.

## Testcase
https://jsfiddle.net/sergei_krylov/zu39pxo5/2/

## Version
2.7.8

sergei-krylov:
## Description
Fixes horizontal divided list first item extra padding.
All the details here in bug: #1158 

## Testcase
https://jsfiddle.net/sergei_krylov/zu39pxo5/2/

## Closes
#1158

### DIFF
--- a/dist/components/list.css
+++ b/dist/components/list.css
@@ -754,7 +754,7 @@ ol.ui.suffixed.list li:before,
 .ui.divided.horizontal.list {
   margin-left: 0;
 }
-.ui.divided.horizontal.list > .item {
+.ui.divided.horizontal.list > .item:not(:first-child) {
   padding-left: 0.5em;
 }
 .ui.divided.horizontal.list > .item:not(:last-child) {
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -764,7 +764,7 @@ ol.ui.suffixed.list li:before,
 .ui.divided.horizontal.list {
   margin-left: 0;
 }
-.ui.divided.horizontal.list > .item {
+.ui.divided.horizontal.list > .item:not(:first-child) {
   padding-left: @horizontalDividedSpacing;
 }
 .ui.divided.horizontal.list > .item:not(:last-child) {


y0hami (approved):
LGTM

exoego (changes requested):
LGTM

lubber-de:
`:not(:first-child)` was already there for horiznotal divided lists.
We removed it by #353 to make sure those items will still align identically when wrapped on mobile devices, so i am against merging this PR.

Look at your adjusted jsfiddle and try to reduce the screen width until the list wraps, then your fix introduces the old wrapping bug again which, imho, is more important to fix
https://jsfiddle.net/Lcht5mpa/
image

lubber-de:
Here is a proposal for goodness 😉 
If we implement a new `unstackable` class and add your suggested fix for the fist-childs left padding, we can satisfy both of us

See https://jsfiddle.net/6u1pmhsc/1/
image

sergei-krylov:
@lubber-de thanks! #353 is a good point that I missed. This definitely makes sense and important for equal-width items so they wrap into straight columns, and probably less important when the list contains items of various widths.

Adding such `unstackable` will add variety to lists, good proposal btw, but it's kind of separate thing as it still does not solve the issue if I have some long horizontal list and really want it to be "stackable".

Let's not go further with these changes so far. I'm using FUI on my project and maybe will get into another solution for both use cases.

_Yeah, I will update only LESS files next time_ :pray:

Issue #1185, opened 2019-11-18
[Input left action] in form


dutrieux:
# Bug Report

## Expected result
Left corners **aren't** rounded when the input is in form
![image](https://user-images.githubusercontent.com/1622751/69054519-064fcd80-0a0c-11ea-8782-8c8f51d032c0.png)

## Actual result
Left corners are rounded when the input is in form
![image](https://user-images.githubusercontent.com/1622751/69054356-ae18cb80-0a0b-11ea-9a96-6421b33064b1.png)

## Testcase
https://jsfiddle.net/9tzLg5vq/1/

## Version
2.8.1

KarlaChen:
Hi, @dutrieux , can I fix it?

dutrieux:
@KarlaChen of course, I have not time at moment

lubber-de:
Their already is a pr from @exoego which is supposed to also handle the left actions. See my review comment there https://github.com/fomantic/Fomantic-UI/pull/1197

exoego:
@KarlaChen 
Thanks for taking a atime to resolve this issue 👍 
I have already updated a PR to resolve this issue in #1197.

It would be very helpful if you deal with other issues.
Issues labeled with `good first issue` are good as entry point.
https://github.com/fomantic/Fomantic-UI/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen++label%3A%22%3A1st_place_medal%3A+good+first+issue%22+

PR #1197, opened 2019-11-27
fix(form): action input had wrong border radius in some cases

### LINKED ISSUE #1184: [Input action] right corners are rounded when input have focus
# Bug Report

## Expected result
When input have focus : right corners are rounded
image

## Actual result
When input have focus : right corners aren't rounded
image

## Testcase
https://jsfiddle.net/9tzLg5vq/

## Version
2.8.1

exoego:
## Description

`.ui.input` in `.ui.form` has border-radius at all corners.
https://github.com/fomantic/Fomantic-UI/blob/2f3312f0f075e3a570dc96725be17e965b3b0964/src/definitions/collections/form.less#L411
Unfortunately, this override brings border-radius to `.ui.action.input`.

This PR removes border-radius from `.ui.action.input` in `.ui.form` where radius not needed.

## Testcase
https://jsfiddle.net/qdp6gxLy/1/

## Screenshot (when possible)
image

## Closes
#1184
#1185

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -413,20 +413,36 @@
   box-shadow: @inputFocusBoxShadow;
 }
 & when (@variationInputAction) {
-  .ui.form .ui.action.input input:not([type]):focus,
-  .ui.form .ui.action.input input[type="date"]:focus,
-  .ui.form .ui.action.input input[type="datetime-local"]:focus,
-  .ui.form .ui.action.input input[type="email"]:focus,
-  .ui.form .ui.action.input input[type="number"]:focus,
-  .ui.form .ui.action.input input[type="password"]:focus,
-  .ui.form .ui.action.input input[type="search"]:focus,
-  .ui.form .ui.action.input input[type="tel"]:focus,
-  .ui.form .ui.action.input input[type="time"]:focus,
-  .ui.form .ui.action.input input[type="text"]:focus,
-  .ui.form .ui.action.input input[type="file"]:focus,
-  .ui.form .ui.action.input input[type="url"]:focus {
+  .ui.form .ui.action.input:not(.left) input:not([type]):focus,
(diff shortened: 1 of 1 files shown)


exoego:
@lubber-de 
Updated a PR in consideration of `.left.action.input`, fixing #1185
https://jsfiddle.net/3fno0rga/

src/definitions/collections/form.less
```diff
@@ -412,6 +412,39 @@
...
+  .ui.form .ui.action.input.left input[type="datetime-local"],
+  .ui.form .ui.action.input.left input[type="email"],
+  .ui.form .ui.action.input.left input[type="number"],
+  .ui.form .ui.action.input.left input[type="password"],
+  .ui.form .ui.action.input.left input[type="search"],
+  .ui.form .ui.action.input.left input[type="tel"],
+  .ui.form .ui.action.input.left input[type="time"],
+  .ui.form .ui.action.input.left input[type="text"],
+  .ui.form .ui.action.input.left input[type="file"],
+  .ui.form .ui.action.input.left input[type="url"] {
```
prudho: Why those lines does not have the `:focus` attribute like above ?

exoego: Because `.left.action.input` have left radius by default (not focused).
image

prudho: Ok, thank for the explanation @exoego !

dutrieux:
@exoego 
Do you consider the case with .labeled.input or .right.labeled.input

I update your jsfiddle to see this case : https://jsfiddle.net/dutrieux/4ets2cgm/

exoego:
@dutrieux 
Thanks for reporting.
Could you open an issue ?

My first feeling is that the issue looks complicated and may require a lot effort, so should be addressed in different PR.
I keep this PR focusing on border radius, so get merged soon.

PR #1210, opened 2019-12-06
fix(dropdown): values are incorrect if they contain some special chars

### LINKED ISSUE #1207: [Dropdown] Values are parsed incorrectly when they contain <, ≤, etc
# Bug Report

## Steps to reproduce
1. Create a dropdown
2. Set its values to `<`, `>`, `≤`, and `≥`
3. Get values and inspect the result

## Expected result
Get the values for `<` and company

## Actual result
Get `&;amp;amp;amp;lt;` and company. The number of `&amp;` might be related to number of selected values?

## Testcase
https://jsfiddle.net/4cqtnf0d/

## Screenshot (when possible)
image

## Version
2.8.2

lubber-de:
##  Description
Dropdown values which are supposed to be encoded into htmlentities like `&lt;` were wrongly encoded whenever the values changed. Especially when working with mulitple selection dropdowns.
Reason for this was the overseen fact that an ampersand `&` was also encoded into `&amp;`. As htmlentities are in general encoded by starting with an ampersand aswell this was totally messing up the value string.

As the escape routine was also implemented in other modules, this PR fixes the logic there as well

### Broken
https://jsfiddle.net/4cqtnf0d/

### Fixed
https://jsfiddle.net/yfgwph04/

## Screenshot
|Before|After|
|-|-|
|image|image|

## Closes
#1207

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2099,7 +2099,7 @@ $.fn.dropdown = function(parameters) {
                       optionValue = optionValue.toLowerCase();
                       value = value.toLowerCase();
                     }
-                    if( String(optionValue) == String(value)) {
+                    if(module.escape.htmlEntities(String(optionValue)) === module.escape.htmlEntities(String(value))) {
                       module.verbose('Found select item by value', optionValue, value);
                       $selectedItem = $choice;
                       return true;
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2079,9 +2079,8 @@ $.fn.dropdown = function(parameters) {
                   if(optionValue === null || optionValue === undefined) {
                     return;
                   }
-                  optionValue = module.escape.htmlEntities(optionValue);
                   if(isMultiple) {
-                    if($.inArray( String(optionValue), value) !== -1) {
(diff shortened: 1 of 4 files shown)


exoego:
XML-like values (e.g. `<xss>`) are not rendered in dropdown both in current and this PR.
Is this as-designed for security reason?
https://jsfiddle.net/2cp5gL7o/

If `<` and `>` are rendered in dropdown, I think `<xml-like>`  should be rendered too.

lubber-de:
> Since `<` and `>` are rendered in dropdown, I think `<xml-like>` should be rendered too.
> This might be trickier, so could be addressed in later.

@exoego `<xml>` is valid html, thus it's parsed and and used as such by default. If you want to display such html code, it works by using `preserverHTML:false`
See https://jsfiddle.net/uLejd364/1/
image

lubber-de:
@hammy2899 I had to fix the remove label selection as @GammaGames  mentioned , so please review again. I updated the jsfiddle in the description accordingly https://jsfiddle.net/uLejd364/

lubber-de:
@exoego @hammy2899  Please review again, i also fixed the situation when  `allowAdditions: true` was set (entries came up twice then... 🙄 )
See here:
#### Broken
https://jsfiddle.net/wckbtaey/

#### Fixed
https://jsfiddle.net/wckbtaey/1/

PR #1215, opened 2019-12-11
feat(modal): allow close icon to be inside header for small modals

lubber-de:
##  Description
On larger screens an existing close icon of a modal is always displayed outside of the modal itself.
Only when using `fullscreen` it is put directly into the header by default.

To make this behavior possible for small modals on large screens as well i invented an optional class `inside` which can be added to the `close icon`

## Testcase
https://jsfiddle.net/m59p0wq1/

## Screenshot
|Before|After|
|-|-|
|image|image|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/2873

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -442,7 +442,7 @@
   }
 }
 
-& when (@variationModalFullscreen) or (@variationModalOverlay) {
+& when (@variationModalFullscreen) or (@variationModalOverlay) or (@variationModalCloseInside) {
   /*--------------
      Full Screen
   ---------------*/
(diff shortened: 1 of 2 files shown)


prudho:
Do you think it is possible to move the close button closer to the top right corner? It's not a big deal but IMHO it's a little too far actually.

lubber-de:
@prudho Well, when we simply use the same value/variable as for mobile view, which is 0.5rem for top/right, it would look better, but isnt inline with the header anymore  (which i personally would not care about 😛 )
### Current
image

### Adjusted to use `@mobileCloseTop/Right = 0.5rem`
image

PR #1216, opened 2019-12-11
feat(popup): add variant to ensure width if parent container is to small

### LINKED ISSUE #1213: [Popup] Showing popup out of very narrow container
# Feature Request

## Description

I want to use `.popup` as rich tooltip inside narrow `.column`.
However, width of `.popup` are limited to its container width, so popup content in narrow column are bad in terms of readability.

### Current: Too narrow popup 💥 
image

### Expected: Relaxed popup out of column 🏆  
image

## Testcase (when possible)
https://jsfiddle.net/yqhpg25n/

lubber-de:
## Description
When a popup is used directly from any html source, it inherits its default width by its parent container. width.
This leads into situations where popups appear too small and text wraps looking ugly

The only solution without changing existing html code is to give such popups a fixed width.
Therefore i added a new optional `fixed` variant.

This PR also supports css-only tooltips , so if `data-variation="fixed"` is used on `data-tooltip` the behavior is the same as for popups

#### Before
https://jsfiddle.net/baukrwxq/

#### After
https://jsfiddle.net/pfdLt8aq/

### Popup Before
popup_width_normal

### Popup using `fixed`
popup_width_fixed

### Tooltip Before
image

### Tooltip using `data-variation="fixed"`
image

## Closes
#1213
https://github.com/Semantic-Org/Semantic-UI/issues/6434

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -721,6 +721,12 @@
   }
 }
 
+& when (@variationPopupFixed) {
+  .ui.fixed.popup {
+    width: @maxWidth;
+  }
+}
+
 & when (@variationPopupWide) {
   /*--------------
        Wide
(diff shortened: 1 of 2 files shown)


lubber-de:
Added tooltip support which will also close some very old SUI issues

PR #1219, opened 2019-12-13
fix(dropdown): handle text metadata on select build dropdowns

### LINKED ISSUE #1103: [Dropdown] Setup select does not convert data-text to Menu items
# Feature Request

## Description
When converting select dropdowns to dropdown menus with `$('select').dropdown()` the data-text metadata is not copied over to the menu dropdown.

Having this feature makes it possible to create select with optgroups, where the visible chosen value as label also contains the optgroup label.

## Testcase
https://jsfiddle.net/6hdtzybr/

prudho:
## Description
This PR preserve the `data-text` attribute when creating the dropdown from a select, so the displayed text can be different than the in-dropdown text.

## Testcase
Before: [JSFiddle](https://jsfiddle.net/6hdtzybr/)
After: [JSFiddle](https://jsfiddle.net/1x4caqfm/2/)

## Closes
#1103

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1975,6 +1975,9 @@ $.fn.dropdown = function(parameters) {
                     value    = ( $option.attr('value') !== undefined )
                       ? $option.attr('value')
                       : name,
+                    text     = ( $option.data('text') !== undefined )
+                      ? $option.data('text')
+                      : name,
                     group = $option.parent('optgroup')
                   ;
                   if(settings.placeholder === 'auto' && value === '') {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -1975,6 +1975,9 @@ $.fn.dropdown = function(parameters) {
                     value    = ( $option.attr('value') !== undefined )
                       ? $option.attr('value')
                       : name,
+                    text     = ( $option.data('text') !== undefined )
```
exoego: Key `'data'` should be defined in `metadata` so one can configure it as other data attribute do so.

lubber-de: It indeed already exists
https://github.com/fomantic/Fomantic-UI/blob/2f3312f0f075e3a570dc96725be17e965b3b0964/src/definitions/modules/dropdown.js#L3989

prudho: Indeed, should be okay now !

PR #1220, opened 2019-12-13
fix(dropdown): clearable icon is not clickable on mobile

### LINKED ISSUE #647: Dropdown Clearable not working on mobile
# Bug Report

## Steps to reproduce
On Android Chrome or iPhone Safari. Try to clear dropdown. It does not work.
But when the dropdown is search dropdown it works fine.

## Testcase
https://fomantic-ui.com/modules/dropdown.html#clearable-selection

## Version
2.7.4

lubber-de:
## Description
On mobile devices a clearable icon was not  working. The mobile browsers do not seem to fetch/interpret the existing "click" event as a "touchstart" event.
I centralized the eventbinding and whenever a touchdevice is recognized it (in most cases) used `touchstart` instead of `click` not, which makes the clearable icon work now :)

### Hint for review
I had to leave one `click` event as it is to prevent a item selection on mobile when you try to scroll the menu list

## Testcase
- Open in a mobile device, especially iOS
### Current behavior
Clearable icon does not work
https://jsfiddle.net/tnk79L1q/show

### Fixed
Clearable works now
https://jsfiddle.net/tnk79L1q/1/show

## Screenshot
ios_clearable

## Closes
#647

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -658,9 +658,9 @@ $.fn.dropdown = function(parameters) {
               }
             }
             $menu
-              .on('mouseenter' + eventNamespace, selector.item, module.event.item.mouseenter)
+              .on((hasTouch ? 'touchstart' : 'mouseenter') + eventNamespace, selector.item, module.event.item.mouseenter)
               .on('mouseleave' + eventNamespace, selector.item, module.event.item.mouseleave)
-              .on(clickEvent   + eventNamespace, selector.item, module.event.item.click)
+              .on('click'      + eventNamespace, selector.item, module.event.item.click)
             ;
           },
           intent: function() {
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -31,6 +31,10 @@ $.fn.dropdown = function(parameters) {
     moduleSelector = $allModules.selector || '',
 
     hasTouch       = ('ontouchstart' in document.documentElement),
+    clickEvent      = hasTouch
+        ? 'touchstart'
+        : 'click',
+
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -671,7 +658,7 @@ $.fn.dropdown = function(parameters) {
               }
             }
             $menu
-              .on('mouseenter' + eventNamespace, selector.item, module.event.item.mouseenter)
+              .on((hasTouch ? 'touchstart' : 'mouseenter') + eventNamespace, selector.item, module.event.item.mouseenter)
               .on('mouseleave' + eventNamespace, selector.item, module.event.item.mouseleave)
```
exoego: Is it OK not adding `touchend` to this line in a symmetric fashion ?

lubber-de: I don't think so, because `mouseleave` would hide a possible submenu. This is ok when using a real mouse but unwanted when using touch devices (one would need to swipe between dropdown submenu entreies instead of touching each entry)
It also wasn't used before this PR (mabe because of the above mentioned reason). Only `mouseenter` had an additional touch event, so i only replace that one.

lubber-de: @exoego Just tested for clarification and i am right, here is the proof: Submenus won't work anymore (because hidden when `touchend` is used instead of keeping `mouseleave`)
https://jsfiddle.net/d0snvrxb/1/show
touchend_bad

Using only the changes of this PR (keep `mouseleave`), it works as before and expected
https://jsfiddle.net/d0snvrxb/show
mouseleave_good

exoego: Understood

PR #1223, opened 2019-12-15
fix(dimmer): borders on partial segment dimmers are incorrect

lubber-de:
## Description
The change in #1027 changed the specificity. This lead into the border correction for partial dimmers in segments to not work anymore.
This PR increases their specificity so they are adjusted again.

I also removed the unnecessary `!important` settings for the `segment > dimmer` border settings

####  Broken corners
https://jsfiddle.net/q1Lfgpk2

#### Fixed again and `!important` removed
https://jsfiddle.net/hq8tcs6w/

#### Broken
image

### Fixed
image

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -295,14 +295,14 @@ body.dimmable > .dimmer {
     -webkit-transform: translateY(calc(-50% - .5px));
   }
 
-  .ui.segment > .ui[class*="top dimmer"] {
+  .ui.segment > .ui.ui[class*="top dimmer"] {
       border-bottom-left-radius: 0 !important;
       border-bottom-right-radius: 0 !important;
   }
-  .ui.segment > .ui[class*="center dimmer"] {
+  .ui.segment > .ui.ui[class*="center dimmer"] {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dimmer.less
```diff
@@ -295,14 +295,14 @@ body.dimmable > .dimmer {
     -webkit-transform: translateY(calc(-50% - .5px));
   }
 
-  .ui.segment > .ui[class*="top dimmer"] {
+  .ui.segment > .ui.ui[class*="top dimmer"] {
       border-bottom-left-radius: 0 !important;
```
exoego: Are these !important still needed?

lubber-de: I'll check that and possibly create another PR. I guess dimmer borders are declared in other components as well

lubber-de: You were right 👍 Turns out, there was no need for the `!important` settings after checking all other components. To avoid merge conflicts i decided to make their removedment part of this PR, so please review again.
I also updated the fixed jsfiddle testcase  in the description accordingly

PR #1228, opened 2019-12-17
[Form] `add rule` was overwriting existing rules

prudho:
## Description
While investigating on adding some autoset rules on required fields, I found that adding a rule trought the `add rule` behavior was overwritting existing rules.

This PR fixes this bad behavior and add some checks in the related function (checks if args are defined, check is a rule already exists...).

## Testcase
Before: [JSFiddle](https://jsfiddle.net/ohvu2q84/)
After: [JSFiddle](https://jsfiddle.net/jsmvop9h/)

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -818,26 +818,36 @@ $.fn.form = function(parameters) {
             module.add.field(name, rules);
           },
           field: function(name, rules) {
+            // Validation should have at least a standard format
+            if(validation[name] === undefined || validation[name].rules === undefined) {
+              validation[name] = {
+                rules: []
+              }
+            }
             var
-              newValidation = {}
+              newValidation = {
+                rules: []
+              }
             ;
             if(module.is.shorthandRules(rules)) {
               rules = Array.isArray(rules)
                 ? rules
                 : [rules]
               ;
-              newValidation[name] = {
-                rules: []
-              };
-              $.each(rules, function(index, rule) {
-                newValidation[name].rules.push({ type: rule });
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -818,26 +818,36 @@ $.fn.form = function(parameters) {
...
             }
             else {
-              newValidation[name] = rules;
+              newValidation.rules = rules.rules;
             }
-            validation = $.extend({}, validation, newValidation);
-            module.debug('Adding rules', newValidation, validation);
+            // For each new rule, check if there's not already one with the same type
+            $.each(newValidation.rules, function (_index, rule) {
+              if (validation[name].rules.findIndex(function(item) { return item.type == rule.type }) == -1) {
```
lubber-de: Unfortunately `findIndex` is not supported in IE11 😢 Try `$.each()` instead

src/definitions/behaviors/form.js
```diff
@@ -818,26 +818,36 @@ $.fn.form = function(parameters) {
             module.add.field(name, rules);
           },
           field: function(name, rules) {
+            // Validation should have at least a standard format
+            if(validation[name] === undefined || validation[name].rules === undefined) {
+              validation[name] = {
+                rules: []
+              }
```
lubber-de: Missing semicolon (although valid and working, but we have it everywhere else...)

src/definitions/behaviors/form.js
prudho: I always forgot to check for IE compatibility (usually babel do it for me...), so it should be fixed now.. but i did not used `$.each()` 😃

PR #1229, opened 2019-12-18
feat(dropdown): add unfilterable item class for search input

arkanovicz:
## Description

Let some items of a dropdown be resilient to filtered text searchs.

### Without the PR
https://jsfiddle.net/arkanovicz/s6k9mqd1/35/

### With this PR included
https://jsfiddle.net/0fo1bas3/

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -4077,7 +4077,7 @@ $.fn.dropdown.settings = {
     header      : 'header',
     divider     : 'divider',
     groupIcon   : '',
-    unfiltered  : 'unfiltered
+    unfiltered  : 'unfiltered'
   }
 
 };
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -848,6 +848,10 @@ $.fn.dropdown = function(parameters) {
                   text,
                   value
                 ;
+                if($choice.hasClass('unfiltered')) {
+                  results.push(this);
+                  return true;
+                }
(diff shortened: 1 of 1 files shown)


arkanovicz:
@lubber-de The example I used in the jsfiddle use another embedded menu, yes. But I can see other potential usecases (like global selectable group items that you want always present), and I came to this solution because I didn't find any other way to do it.

@prudho  I commited the requested change, it now uses the className map.

src/definitions/modules/dropdown.js
```diff
@@ -4076,7 +4076,8 @@ $.fn.dropdown.settings = {
     delete      : 'delete',
     header      : 'header',
     divider     : 'divider',
-    groupIcon   : ''
+    groupIcon   : '',
+    unfiltered  : 'unfiltered
```
prudho: You've missed the closing quote 😃

arkanovicz: Damn! Fixed...

lubber-de:
I would basically approve this now. However, i think we should name the new class `unfilterable` instead of `unfiltered` , because, by the name, all items are "unfiltered" by default.

So if we want to declare that some items should be **unfilterable**, we should name it like this, because it describes a *property* to the item (just like  `stackable` , `unstackable`, `pushable`, `pausable` in other components)  and not a temporary **state** (like `filtered` is)

arkanovicz:
@lubber-de done - 'unfilterable' instead of 'unfiltered'. Yes, more consistent.

PR #1232, opened 2019-12-19
fix(slider): preventCrossover should only affect range sliders

### LINKED ISSUE #1231: [Slider] [commit 4b1d548] broke smooth: true with preventCrossover: true
At the moment the DOCS of Fomantic show examples of smoothly repositioned slider.
It however it does not move smoothly, and the slider thumb moves in discreet steps to where the click was registered (even in cases of `step: 0`).

The root cause is the [commit 4b1d548](https://github.com/fomantic/Fomantic-UI/commit/4b1d54894c640896e6554358140e4578159a9bf4#diff-52cb3744b99612b0af91190a82e69b59)
Commit message says
> BREAKING CHANGE: `preventCrossover ` is set to true by default, making this a breaking change.

Current state of affairs: `smooth` works only when `preventCrossover` is set to false

Ideas for possible solutions:
(a) handle both `smooth: true` and `preventCrossover: true`,
(b) immediately - do not set `preventCrossover: true` and/or add information to Docs that prevent `preventCrossover: true` renders `smooth: true` invalid.

ColinFrick:
## Description
Fixes a regression introduces in  4b1d54894c640896e6554358140e4578159a9bf4
The preventCrossover setting affected single sliders as well, which prevented them to slider correctly.

## Testcase
Before: https://jsfiddle.net/kj2m03f6/
After: https://jsfiddle.net/e5yt26bg/

## Closes
#1231

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -353,13 +353,13 @@ $.fn.slider = function(parameters) {
                 thumbSmoothVal = module.determine.smoothValueFromEvent(event)
               ;
               if(!$currThumb.hasClass('second')) {
-                if(settings.preventCrossover) {
+                if(module.is.range() && settings.preventCrossover) {
                   value = Math.min(secondThumbVal, value);
                   thumbSmoothVal = Math.min(secondThumbVal, thumbSmoothVal);
                 }
                 thumbVal = value;
               } else {
-                if(settings.preventCrossover) {
+                if(module.is.range() && settings.preventCrossover) {
                   value = Math.max(thumbVal, value);
                   thumbSmoothVal = Math.max(thumbVal, thumbSmoothVal);
                 }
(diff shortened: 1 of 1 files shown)


src/definitions/modules/slider.js
```diff
@@ -353,13 +353,13 @@ $.fn.slider = function(parameters) {
                 thumbSmoothVal = module.determine.smoothValueFromEvent(event)
               ;
               if(!$currThumb.hasClass('second')) {
-                if(settings.preventCrossover) {
+                if(module.is.range() && settings.preventCrossover) {
```
exoego: I think that evaluating boolean flag `settings.preventCrossover` is faster than `module.is.range()`.

So I propose to flip them like `if(module.is.range() && module.is.range()) {`

hammy2899: `module.is.range()` is only checking for a class https://github.com/ColinFrick/Fomantic-UI/blob/fix/1231/src/definitions/modules/slider.js#L473 the performance benefit from switching these would be basically invisible but change as you wish @ColinFrick

ColinFrick: Done

PR #1244, opened 2019-12-27
[Form] Auto add empty rule validation for required fields

### LINKED ISSUE #1094: [Form Validation] Automatically add empty rule validation for required fields
Not sure if it exists, don’t believe it does, but how about automatically adding form fields (input/textarea/select) that have the `required=“required”` attribute on them as a validation rule for the empty check? Some other `input` could also have automatic rules, such as `type=“number”` with the min/max attributes, etc etc. Could potentially have it apply with the required+field/fields classes too

I should think Listeners may be required so that if a field were to become no longer required/now required it would be removed/added to the rules. Same may need to be done if a field were added to/removed from the DOM, and something similar if a hidden field becomes visible (?)

If a field were in a segment tab, then when validating the tab should be switched to the tab with the first incomplete field (I don’t think this exists at the moment)

I should also think that a submit button would have to also trigger the validation of whatever form it is inside of the fields that belong to that form

I think if this were done on the fly, then could save some effort in building validation for every form, and only additional validation needs to be manually added, or overrides added in.

prudho:
## Description
This PR adds the ability to add automatically an 'empty' rule for fields that got the `required` class (or `checked` for checkboxes). It's disabled by default since I don't want it to be a breaking change, but can be enabled at initialization with `autoCheckRequired: true` or any time with the `.form('set auto check')` behavior.

It can handle all fields types, grouped fields as well, and is able to detect disabled fields or groups. Rule is added to others if there's already some defined, and isn't duplicated if already set.

## Testcase
[JSFiddle](https://jsfiddle.net/m2xaLgsz/)

## Closes
#1094

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -95,6 +95,9 @@ $.fn.form = function(parameters) {
             module.verbose('Initializing form validation', $module, settings);
             module.bindEvents();
             module.set.defaults();
+            if (settings.autoCheckRequired) {
+              module.set.autoCheck();
+            }
             module.instantiate();
           }
         },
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -1106,6 +1109,32 @@ $.fn.form = function(parameters) {
...
+            $field.each(function (_index, el) {
+              var
+                $el        = $(el),
+                $elGroup   = $(el).closest($group),
+                isCheckbox = ($el.filter(selector.checkbox).length > 0),
+                isRequired = $el.prop('required') || $elGroup.hasClass(className.required) || $elGroup.parent().hasClass(className.required),
+                isDisabled = $el.prop('disabled') || $elGroup.hasClass(className.disabled) || $elGroup.parent().hasClass(className.disabled),
+                validation = module.get.validation($el),
+                hasEmptyRule = validation
+                  ? (validation.rules.findIndex(function(rule) { return rule.type == "empty" }) > -1)
```
lubber-de: `findIndex` is not supported in IE11 unfortunately 😢  You may use the `$.grep()` approach as you did in #1228

lubber-de: @prudho reminder... 😉

prudho: Yup, i'll try to fix this tomorrow.

prudho: Done !

prudho: Done !

PR #1266, opened 2020-01-14
fix(dropdown): get item function now working properly with array of i…

### LINKED ISSUE #1265: [Dropdown] "get item" function not working when value is an array of integers
# Bug Report
The function `get.item(value, strict)` does only implicitly convert `value` to String when `value` is not an array.
Found this bug trying to set a multiple dropdown value to an array of integer (didn't work), tried the same command with an array of strings and it worked.

## Steps to reproduce
https://jsfiddle.net/b9ko3sfn/1/

## Expected result
The result of `get.item(["2"])` and `get.item([2])` should be the same

## Actual result
`get.item([2])` returns false

## Version
2.8.3

natoli66:
Fixes #1265

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2075,7 +2075,7 @@ $.fn.dropdown = function(parameters) {
                     return;
                   }
                   if(isMultiple) {
-                    if($.inArray(module.escape.htmlEntities(String(optionValue)), value) !== -1) {
+                    if($.inArray(module.escape.htmlEntities(String(optionValue)), value.map(function(v){return String(v);})) !== -1) {
                       $selectedItem = ($selectedItem)
                         ? $selectedItem.add($choice)
                         : $choice


src/definitions/modules/dropdown.js
```diff
@@ -2075,7 +2075,7 @@ $.fn.dropdown = function(parameters) {
                     return;
                   }
                   if(isMultiple) {
-                    if($.inArray(module.escape.htmlEntities(String(optionValue)), value) !== -1) {
+                    if($.inArray(module.escape.htmlEntities(String(optionValue)), String(value)) !== -1) {
```
lubber-de: You are casting an array (`value` is an array here!) to a string. This results in wrong results if the value has more than one item
```console
String([1,2,3,4]) = "1,2,3,4"
$.inArray("2","1,2,3,4") -> 2 -> wrong
```
You have to cast each single possible item in the array instead

```javascript
if($.inArray(module.escape.htmlEntities(String(optionValue)), value.map(function(v){return String(v);})) !== -1) {
```

natoli66: Idk how I ended up casting this way, got it working so I guess it was fine for me,
anyway, fixed the fix, and got my branch cleaned

Thanks for the review though

PR #1275, opened 2020-01-18
[Modal] Border radius for inner dimmer should inherit from the modal

### LINKED ISSUE #1274: CSS bug multiple modals border-radius missing on second+ overlay.
Description: Where multiple modals are displayed (allowMultiple = true) when the second modal opens over the first, then the faint dimmer div that is placed over the first modal is missing the css border-radius setting. Because modals have a border-radius, this has the effect of putting square corners on modal 1. 

Repro of issue: https://jsfiddle.net/fsge8zo9/3/ Look at bottom left corner of first modal when second modal has been opened.

lubber-de:
## Description
The inner dimmer which covers a parent modal in case `allowMultiple: true` is used, didn't have a border radius. This results in visually removing the border from the modal itself.

## Testcase
Reomve the CSS to see the issue
https://jsfiddle.net/10dbpftg/

## Screenshot 
Look at the top left corner 

|Before|After|
|-|-|
|image|image|

## Closes
#1274

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -53,6 +53,10 @@
   border-bottom-right-radius: @borderRadius;
 }
 
+.ui.modal > .ui.dimmer {
+  border-radius: inherit;
+}
+
 /*******************************
             Content
 *******************************/


VanquishedWombat:
I confirm the supplied fiddle illustrating the fix at https://jsfiddle.net/10dbpftg/ looks good to me.

PR #1298, opened 2020-01-31
[Form, Calendar] Calendar inline fields did not display properly

### LINKED ISSUE #1296: [Calendar] inline form fields
# Bug Report
Calendar bad rendering on inline form fields

## Testcase
https://jsfiddle.net/1h2yntv3/

## Screenshot
image

## Version
2.8.3

lubber-de:
## Description
When a `ui calendar` is used within a `inline field`, its display was broken (shrunk and not inline)
I had to use a proper direct em setting to fix this, as the input field is inside the calendar component making the usual `width: auto` not work as for the other standard input form fields.

## Testcase
- remove the CSS to see the issue
https://jsfiddle.net/3payxo5h/1/

### Before
image

### After
image

## Closes
#1296

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -1097,6 +1097,16 @@
     font-size: @inlineInputSize;
   }
 
+  .ui.form .inline.fields .field .calendar:not(.popup),
+  .ui.form .inline.field .calendar:not(.popup) {
+    display:inline-block;
+  }
+
+  .ui.form .inline.fields .field .calendar:not(.popup) > .input > input,
+  .ui.form .inline.field .calendar:not(.popup) > .input > input {
+    width: @inlineCalendarWidth;
+  }
+
   /* Label */
   .ui.form .inline.fields .field > :first-child,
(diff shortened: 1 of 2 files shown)


lubber-de:
@prudho I was trying but without success. The 13.11em was the best result I could get. (at least those 1pixel difference is better than the current result 😁) 
I really have no idea how to fix this otherwise. I would be glad if someone comes up with a better solution

prudho:
I won't find a better solution than you. Just left a comment explaining this method and we're good :)

lubber-de:
@prudho uhm, i am unsure/ confused :do you want me to explain why I did the fixed EM width to get this approved ? Well, because that's the only solution I found... And because the calendars input is not at the same node level than usual field inputs, so width :auto is not going to work because the browser cannot figure out that that the calendars input should behave equally to the other Form inputs.... Don't know what to tell more... Or did i misunderstand? 😉

prudho:
I'm myself unable to tell you what I wanted to know there... Anyway, this is the best solution actually, so LGTM ;)

PR #1316, opened 2020-02-08
fix(#1315): allow optional compression

### LINKED ISSUE #1315: [Gulp] Optional compression
The default build configuration builds both compressed and non-compressed files with no way to disable on or the other.

An option could be added into ``semantic.json`` to define whether compressed files should be generated or not.

jimmy-lt:
Add a build option to only generate compressed CSS and Javascript on demand.
The default is kept on the previous behavior of building both flavours.

## Closes
#1315

### DIFF
--- a/tasks/build/css.js
+++ b/tasks/build/css.js
@@ -90,7 +90,7 @@ function build(src, type, compress, config, opts) {
  */
 function pack(type, compress) {
   const output       = type === 'docs' ? docsConfig.paths.output : config.paths.output;
-  const ignoredGlobs = type === 'rtl' ? globs.ignoredRTL + '.rtl.css' : globs.ignored + '.css';
+  const ignoredGlobs = type === 'rtl' ? globs.ignoredRTL + (compress ? '.min.rtl.css' : '!(*.min).rtl.css') : globs.ignored + (compress ? '.min.css' : '!(*.min).css');
 
   let concatenatedCSS;
   if (type === 'rtl') {
@@ -98,11 +98,10 @@ function pack(type, compress) {
   } else {
     concatenatedCSS = compress ? filenames.concatenatedMinifiedCSS : filenames.concatenatedCSS;
   }
-
-  return gulp.src(output.uncompressed + '/**/' + globs.components + ignoredGlobs)
+  return gulp.src((compress ? output.compressed : output.uncompressed) + '/**/' + globs.components + ignoredGlobs)
     .pipe(plumber())
     .pipe(dedupe())
-    .pipe(replace(assets.uncompressed, assets.packaged))
+    .pipe(replace(compress ? assets.compressed : assets.uncompressed, assets.packaged))
     .pipe(concatCSS(concatenatedCSS, settings.concatCSS))
     .pipe(gulpif(config.hasPermissions, chmod(config.parsedPermissions)))
     .pipe(gulpif(compress, minifyCSS(settings.concatMinify)))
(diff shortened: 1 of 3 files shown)


lubber-de (dismissed):
LGTM

tasks/build/css.js
```diff
@@ -139,10 +139,16 @@ function buildCSS(src, type, config, opts, callback) {
   const packCompressed       = () => pack(type, true);
   packCompressed.displayName = 'Packing compressed CSS';
 
-  gulp.parallel(
-    gulp.series(buildUncompressed, packUncompressed),
-    gulp.series(buildCompressed, packCompressed)
-  )(callback);
+  if (config.compressed === undefined || config.compressed === 'both') {
```
exoego: `compressed`  won't be `null` ?
If it will, `null` should be handled in if.

lubber-de: Just `!config.compressed` should fetch undefined,null and false, doesnt it?

spack971: > `compressed` won't be `null` ?
> If it will, `null` should be handled in if.

It can only if the user set the config value to ``null``. By the way, I've replicated the same structure as the [`rtl`](https://github.com/fomantic/Fomantic-UI/blob/a49c82330af6715f4de6332d40ed80f2fb070dbc/tasks/build/css.js#L159) config option where ``null`` is not taken into account.

> Just `!config.compressed` should fetch undefined,null and false, doesnt it?

``false`` should only produce uncompressed files. ``null`` and ``undefined`` can be bound to the original behavior of building both kind of files.

lubber-de (dismissed):
LGTM

tasks/build/css.js
exoego: I think it is common practice that handle `null` and `undefined` in same manner  in terms of configuration.
Unless intended for some reasons, `rtl` should be fixed in separate PR, and this PR should not repeat the error (not taking `null` into account)  in `rtl`.

tasks/build/javascript.js
```diff
@@ -46,19 +46,16 @@ const
...
  */
-function build(src, type, config) {
+function build(src, type, compress, config) {
   return gulp.src(src)
     .pipe(plumber())
     .pipe(flatten())
     .pipe(replace(comments.license.in, comments.license.out))
-    .pipe(gulpif(config.hasPermissions, chmod(config.parsedPermissions)))
-    .pipe(gulp.dest(config.paths.output.uncompressed))
-    .pipe(print(log.created))
```
lubber-de: Shouldn't this be kept? It might be needed to do the uglify and compress.. The destination path is now only changed after the possible compression, when this is removed

Otherwise there will always be the same path for compressed/uncompressed . (Line 58)
This is probably always the case anyway, but to follow the possible different paths we should keep it backward compatible....Or am i missunderstanding the logic here ? @ColinFrick Any advise?

ColinFrick: This should be fine.

ColinFrick (changes requested):
Currently generating the packaged, compressed JavaScript / CSS files (semantic.min.js and semantic.min.css) are generated from the non compressed files, so when `compressed` is true no packaged files are generated anymore.

spack971:
@ColinFrick this is fixed in this latest update although I'm not a big fan of the adopted solution. I'll check whether [gulp-filter](https://www.npmjs.com/package/gulp-filter) can provide with better help on this.

lubber-de:
@spack971 Have you been successful investigating if gulp-filter provides a better soluttion or should we re-review your changes now?
Otherwise: Your PR has merge conflicts, please resolve (merging/rebasing on the current develop branch should do it)

spack971:
@lubber-de I've updated the branch. gulp-filter can be an option but I think it requires another pull request as it can impact other areas of the configuration.

lubber-de:
> @lubber-de  ... but I think it requires another pull request as it can impact other areas of the configuration.

So, should we review **this** PR now or do you plan to create another PR based on gulp-filter which should then be used/reviewed **instead** of this PR ?

spack971:
You can review this one now.

lubber-de:
Closing, as source branch was deleted and PR is therefore unmergeable. Also there was no more investigation about gulp-filter as the original author wasnt happy with the workaround

PR #1323, opened 2020-02-12
Replace all rule tokens in prompt message 

Joe-Soap:
## Scope 
Affects all prompt text values that use replacement tokens

## Description
Within the prompt property method of the form module, I updated the replace calls, where it replaces contextual tokens relating to the rules definition,  to use a regular expression of the same respective values in order to have it replace all tokens present in the prompt text. 

The tokens used are the following: 
* identifier
* name
* ruleValue

If one uses a singular token multiple times in the prompt, for example,

`{ 
    type: 'levelCheck[2]', 
    prompt: 'Level {ruleValue} not detected. Please ensure that {name} is set to a level at least equal to {ruleValue}.' 
}`

 the first token of {ruleValue} is successfully replaced, but the subsequent {ruleValue} isn't. This change merely applies a regex with the "global" flag set in order to replace multiple tokens of the same name found in the prompt text. 

### Broken
https://jsfiddle.net/08361wud/3/

#### Fixed
https://jsfiddle.net/1wyx2vt7/

#### Broken
image

#### Fixed
image

The above example illustrates a use case and the problem.


src/definitions/behaviors/form.js
```diff
@@ -545,18 +545,18 @@ $.fn.form = function(parameters) {
               name
             ;
             if(requiresValue) {
-              prompt = prompt.replace('{value}', $field.val());
+              prompt = prompt.replace(/\{value\}/g $field.val());
```
lubber-de: Missing comma, should read
```javascript
prompt = prompt.replace(/\{value\}/g, $field.val());
```

Joe-Soap: fixed. Thank you

lubber-de:
And, as @ColinFrick already pointed out: Name your PR Title properly instead of "Develop". Thank you 🙂

Joe-Soap:
> All changes have to be made in the source files, please do not commit the dist files.
> Additionally please change the title to something meaningful

Done

Joe-Soap:
> And, as @ColinFrick already pointed out: Name your PR Title properly instead of "Develop". Thank you 🙂

Done

Joe-Soap:
> Please, still, remove the dist files. You may have to reset your develop branch, apply your patch to the single form.js in the behavior folder _only_ and then commit and do a force push

I reverted the changes to the dist folder. Do I really need to reset?

src/definitions/behaviors/form.js
```diff
@@ -545,18 +545,18 @@ $.fn.form = function(parameters) {
...
+              prompt = prompt.replace(/\{value\}/g, $field.val());
             }
             if(requiresName) {
               $label = $field.closest(selector.group).find('label').eq(0);
               name = ($label.length == 1)
                 ? $label.text()
                 : $field.prop('placeholder') || settings.text.unspecifiedField
               ;
-              prompt = prompt.replace('{name}', name);
+              prompt = prompt.replace(/\{name\}/g', name);
```
lubber-de: Remove the single quote, which breaks the script. It should read:
```javascript
prompt = prompt.replace(/\{name\}/g, name);
```

Joe-Soap: Fixed. Apologies. This is my first contribution. Silly mistake. Nerves, I think.

lubber-de: Just test your code before submitted a PR 😉 don't do "dry" PRs, because you think it will work

Joe-Soap: Understood. Agreed. That's one of the underlying reasons I was apologizing

Joe-Soap:
> LGTM
> I adjusted the PR description and added another testcase where the fix is applied and proof screenshots

Thanks for your help lubbe

Joe-Soap:
@ColinFrick After the recent changes, is there anything else for me to change? The merge will only be one file...

PR #1329, opened 2020-02-17
[Calendar] Fire onchange only when date selection is complete

prudho:
## Description
This PR allow to fire `onChange` only when a date selection if complete. For a datetime for example, onChange is called when a date is selected, then when a hour is selected, and again when minutes are selected... which is IMHO is a complete nonsense. If I apply a method when a date is changed, i'd prefer trigger this change only when i have finished to select my date...

Note 1: This may be labelled as a breaking change.

Note 2: This PR also include a small "refactor" of variables declarations I planned to publish later. As you know my git skills are legendary so revert this is really a PITA for me. As it's only minimal style change, do you mind if we let this PR as it ?

## Testcase
Select a date, then a time and then a hour.

Before: [JSFiddle](https://jsfiddle.net/cxb4ogzn/)
After: [JSFiddle](https://jsfiddle.net/cxb4ogzn/1/)

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -815,7 +815,7 @@ $.fn.calendar = function(parameters) {
             module.set.mode(newMode);
             if (mode === 'hour' || (mode === 'day' && module.get.date())) {
               //the user has chosen enough to consider a valid date/time has been chosen
-              module.set.date(date);
+              module.set.date(date, true, false);
             } else {
               module.set.focusDate(date);
             }
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -202,33 +202,38 @@ $.fn.calendar = function(parameters) {
           calendar: function () {
             var i, r, c, p, row, cell, pageGrid;
 
-            var mode = module.get.mode();
-            var today = new Date();
-            var date = module.get.date();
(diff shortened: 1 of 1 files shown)


lubber-de:
@fomantic/maintainers Should we merge this into 2.8.4 now or postpone to 2.9 because of the possible breaking change consideration prudho mentioned? Your thoughts?

prudho:
I'm pretty mixed about releasing it with 2.8.4, since I have another incoming PR which will be a breaking change, and I think both of them will need heavy tests to be sure there's no hidden bug (date setting logic is pretty much rewritten).

lubber-de:
So then postpone this for 2.9.  Especially when you will provide another PR later on :)

lubber-de:
> I'm pretty mixed about releasing it with 2.8.4, since I have another incoming PR which will be a breaking change, and I think both of them will need heavy tests to be sure there's no hidden bug (date setting logic is pretty much rewritten).

@prudho Do you still work on that mentioned PR?

prudho:
@lubber-de no. As you've seen it i'm not really active during this strange days. I think this PR can be merged if you want to.

lubber-de:
Then i tag this for 2.8.5, because it's really more a bugfix than a feature, IMHO

PR #1334, opened 2020-02-19
[Table] Invented a helper class to solve the styling logic when rowspan is used in tables

lubber-de:
## Description
The styling is sometimes broken in the `table`  element, when `rowspan` used. This confused later table rows which then usually have less columns.

As there is no way to look backward via CSS to find out if a previous column in a row was rowspanned and the table component is not a JS module, i invented a little helper class to solve this issue via pure html/css.

One just has to create a `<td>` node , where it is usually not necessary, and give it the new `rowspanned`  class. This will just hide it from the screen but all existing table classes are working again, because the amount of td remains the same just like if there was no rowspan used at all.

So, instead of

```html
<table class="ui definition table">
		<tr>
			<td rowspan="2">Category</td>
			<td>Value 1</td>
		</tr>
		<tr>
			<td>Message for previous row</td>
		</tr>
</table>
```
you do

```html
<table class="ui definition table">
		<tr>
			<td rowspan="2">Category</td>
			<td>Value 1</td>
		</tr>
		<tr>
			<td class="rowspanned"></td>
			<td>Message for previous row</td>
		</tr>
</table>
```

### Broken
https://jsfiddle.net/uoswazex/

### Fixed 
https://jsfiddle.net/7u2qodh9/1/

### Before
image

### After
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/3571

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -59,6 +59,12 @@
   transition: @transition;
 }
 
+/* Rowspan helper class */
+.ui.table th.rowspanned,
+.ui.table td.rowspanned {
+  display:none;
+}
+
 /* Headers */
 .ui.table > thead {
   box-shadow: @headerBoxShadow;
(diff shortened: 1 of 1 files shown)


lubber-de:
docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/198

PR #1346, opened 2020-02-25
[Form Validation] fix "disabled" detection

### LINKED ISSUE #1345: [Form Validation] should not validate fields inside `fieldset[disabled] `
Any fields inside `fieldset[disabled]` should be disabled: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attr-disabled

Currently, the code only checks field props whether it's disabled:
```js
            var isDisabled = true;
            $.each($field, function(){
                if(!$(this).prop('disabled')) {
                  isDisabled = false;
                  return false;
                }
            });
```

It should check `:disabled` with jQuery (IE and old Edge cannot handle native `:disabled` selector properly): https://api.jquery.com/disabled-selector/

Test Case:
https://codepen.io/jamesjav/pen/ZEGOMWO

`baz` is under `fieldset[disabled]` so it should not prevent submitting the form

jameschenjav:
Currently, the code to detect a field is `disabled` cannot handle fields under `fieldset[disabled]`

## Description
Properly detect `disabled` with jQuery `:disabled` selector

## Testcase
https://codepen.io/jamesjav/pen/ZEGOMWO

Click "Validate" button, "baz" field should not have error since it won't be in form data.

> edited: added a test case from @lubber-de for radio buttons with first one disabled

## Closes
#1345

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1128,7 +1128,7 @@ $.fn.form = function(parameters) {
                 $elGroup   = $(el).closest($group),
                 isCheckbox = ($el.filter(selector.checkbox).length > 0),
                 isRequired = $el.prop('required') || $elGroup.hasClass(className.required) || $elGroup.parent().hasClass(className.required),
-                isDisabled = $el.prop('disabled') || $elGroup.hasClass(className.disabled) || $elGroup.parent().hasClass(className.disabled),
+                isDisabled = $el.is(':disabled') || $elGroup.hasClass(className.disabled) || $elGroup.parent().hasClass(className.disabled),
                 validation = module.get.validation($el),
                 hasEmptyRule = validation
                   ? $.grep(validation.rules, function(rule) { return rule.type == "empty" }) !== 0
@@ -1207,13 +1207,7 @@ $.fn.form = function(parameters) {
               module.debug('Using field name as identifier', identifier);
               field.identifier = identifier;
             }
-            var isDisabled = true;
-            $.each($field, function(){
-                if(!$(this).prop('disabled')) {
-                  isDisabled = false;
-                  return false;
-                }
-            });
+            var isDisabled = !$field.filter(':not(:disabled)').length;
             if(isDisabled) {
               module.debug('Field is disabled. Skipping', identifier);
             }


VanquishedWombat:
Question - what is the use-case for when I would want a disabled field, where I cannot enter data, to be validated in form validation? Since I cannot enter data into the disabled field then I cannot fix the failed validation situation. I may be missing something but IMHO this would not be great UX ? I'm probably missing something.

jameschenjav:
Hi @VanquishedWombat 
For example, payment needs to be made, and a few options are available. So each of the methods is assigned to a `<fieldset>` that contains several fields with different rules. Depends on the selection, simply toggle `disabled` and show/hide the fieldsets.

prudho:
`fieldset` isn't handled by fomantic actually. But you can also surround grouped fields with class `disabled fields` and it should work.

jameschenjav:
> `fieldset` isn't handled by fomantic actually. But you can also surround grouped fields with class `disabled fields` and it should work.

Hi @prudho 
not sure following you, but i changed my test case to `fieldset.disabled.fields[disabled]`

src/definitions/behaviors/form.js
```diff
@@ -1207,13 +1207,7 @@ $.fn.form = function(parameters) {
...
               field.identifier = identifier;
             }
-            var isDisabled = true;
-            $.each($field, function(){
-                if(!$(this).prop('disabled')) {
-                  isDisabled = false;
-                  return false;
-                }
-            });
+            var isDisabled = $field.is(':disabled');
```
prudho: This line doesn't work properly when a field has multiple input. Example:
```html
<fieldset class="disabled grouped fields" disabled>
  <label>Bananaaaaa</label>
  <div class="field">
    <div class="ui checkbox">
      <input type="checkbox" name="fruit" tabindex="0" class="hidden">
      <label>Apples</label>
    </div>
  </div>
  <div class="field">
    <div class="ui checkbox checkbox">
      <input type="checkbox" name="fruit" tabindex="0" class="hidden">
      <label>Oranges</label>
    </div>
  </div>
  <div class="field">
    <div class="ui checkbox checkbox">
      <input type="checkbox" name="fruit" tabindex="0" class="hidden">
      <label>Pears</label>
    </div>
  </div>
  <div class="field">
    <div class="ui checkbox checkbox">
      <input type="checkbox" name="fruit" tabindex="0" class="hidden">
      <label>Grapefruit</label>
    </div>
  </div>
</fieldset>
```

lubber-de: Yes,  we added that field loop because otherwise radio groups were not valiadated if the first one was disabled. https://github.com/fomantic/Fomantic-UI/pull/372
So there was a reason why each field has to be checked separately

jameschenjav: @prudho i still don't understand what's your "work" means. all fields inside `fieldset[disabled]` are disabled [by design](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attr-disabled). jquery [`:disabled` selector](https://api.jquery.com/disabled-selector/) works properly [even in IE](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled#Browser_compatibility), while `prop('disabled')` cannot handle this case.

@lubber-de updated PR and added your test case in my demo

prudho: @jameschenjav sorry i was mistaken there :/ I supposed that validation was testing an eventual `disabled` class for field groups, which isn't the case here.

src/definitions/behaviors/form.js
```diff
@@ -1207,13 +1207,7 @@ $.fn.form = function(parameters) {
...
               field.identifier = identifier;
             }
-            var isDisabled = true;
-            $.each($field, function(){
-                if(!$(this).prop('disabled')) {
-                  isDisabled = false;
-                  return false;
-                }
-            });
+            var isDisabled = !$field.filter(':not(:disabled):first').length;
```
lubber-de: The `:first` selector does not seem to be necessary . It is also [deprecated as of Jquery 3.4](https://api.jquery.com/first-selector/).

jameschenjav: ok i will remove it. thought it's faster

lubber-de:
@all-contributors please add @jameschenjav for code

PR #1353, opened 2020-03-05
Update visibility.js

uurcm16:
$context.scrollLeft must be replaced with $context.scrollLeft() otherwise it returns NaN.

### DIFF
--- a/src/definitions/behaviors/visibility.js
+++ b/src/definitions/behaviors/visibility.js
@@ -902,7 +902,7 @@ $.fn.visibility = function(parameters) {
               element.offset.top += $context.scrollTop() - $context.offset().top;
             }
             if(module.is.horizontallyScrollableContext()) {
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/visibility.js
```diff
@@ -902,7 +902,7 @@ $.fn.visibility = function(parameters) {
               element.offset.top += $context.scrollTop() - $context.offset().top;
             }
             if(module.is.horizontallyScrollableContext()) {
-              element.offset.left += $context.scrollLeft - $context.offset().left;
+              element.offset.left += $context.scrollLeft() - $context.offset().left;
```
exoego: Note) [$elem.scrollLeft is a function](https://api.jquery.com/scrollLeft/), so it requies `()` when invoked.

lubber-de:
@all-contributors please add @ucyildirim for code

PR #1366, opened 2020-03-14
[Calendar] new option autoAdjustDateEdges avoid invalid mindate/maxdate combinations

### LINKED ISSUE #1364: setting a Calendar minDate that is later than the currently set maxDate fails silently
I've come across some unexpected behaviour using the Calendar module:
Given I have a calendar widget with minDate and maxDate set,
when I set a new minDate that is later than the current maxDate
I would like to 
- either get an exception saying the minDate must not be later than the current maxDate
- or the new minDate to be accepted and the maxDate to be unset (i.e. set to null).

What I experience instead is that the setting of the new minDate silently fails.
[Have a look at this fiddle to see what I mean.](https://jsfiddle.net/7qgtjzvx/)

lubber-de:
## Description
This PR adds a new setting `autoAdjustDateEdges` (default `false` to stay backward compatible)

When enabled and a call to `set.minDate` or `set.maxDate` would result in an invalid range (min > max or max < min) it will automatically set the opposite edge to the same value, so the range stays valid.

The current (and still default) behavior already fetches invalid values but only shows this in the log when debug and verbose is enabled, so fails silently. As this is rather an error than a verbose information i also changed the log level to error in those cases.

## Testcase
https://jsfiddle.net/k386q42s/1/

## Closes
#1364

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -745,20 +745,26 @@ $.fn.calendar = function(parameters) {
           minDate: function (date) {
             date = module.helper.sanitiseDate(date);
             if (settings.maxDate !== null && settings.maxDate <= date) {
-              module.verbose('Unable to set minDate variable bigger that maxDate variable', date, settings.maxDate);
-            } else {
-              module.setting('minDate', date);
-              module.set.dataKeyValue(metadata.minDate, date);
+              if (!settings.autoAdjustDateEdges) {
+                module.error('Unable to set minDate variable bigger than maxDate variable', date, settings.maxDate);
+                return;
+              }
+              module.set.maxDate(date);
             }
+            module.setting('minDate', date);
+            module.set.dataKeyValue(metadata.minDate, date);
           },
           maxDate: function (date) {
             date = module.helper.sanitiseDate(date);
             if (settings.minDate !== null && settings.minDate >= date) {
-              module.verbose('Unable to set maxDate variable lower that minDate variable', date, settings.minDate);
-            } else {
-              module.setting('maxDate', date);
-              module.set.dataKeyValue(metadata.maxDate, date);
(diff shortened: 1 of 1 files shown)


src/definitions/modules/calendar.js
```diff
@@ -745,20 +745,26 @@ $.fn.calendar = function(parameters) {
...
             if (settings.maxDate !== null && settings.maxDate <= date) {
-              module.verbose('Unable to set minDate variable bigger that maxDate variable', date, settings.maxDate);
-            } else {
-              module.setting('minDate', date);
-              module.set.dataKeyValue(metadata.minDate, date);
+              if (!settings.autoAdjustDateEdges) {
+                module.error('Unable to set minDate variable bigger than maxDate variable', date, settings.maxDate);
+                return;
+              }
+              module.set.maxDate(date);
```
exoego: My understanding of new behavior: If the new option is enabled, minDate and maxDate both use `date`.
What use cases can be considered for such behavior ?

lubber-de: The same date is only set for both in case the mindate would otherwise be set after the maxdate (or the maxdate before an existing mindate) which is an invalid range. To keep the range valid, the date is set to the same date.  
This is for cases where, for example, the min/max date is set by a user interaction and has to keep a valid range. 

The only other possibility I see is to set the other date to null instead of the same date.

Better approach then?

exoego: Ah, I see.

exoego (dismissed):
LGTM

lubber-de:
@fomantic/maintainers  Before reviewing, please take a look at the conversion in the original issue https://github.com/fomantic/Fomantic-UI/issues/1364#issuecomment-599206893

ColinFrick (dismissed):

PR #1379, opened 2020-03-23
fix(search) Fix padding it description when search inside modal

byte916:
Modal module adds padding to ".description" block when window size less 769px.
Search blocks can contain ".description" inside. When search block is inside modal, then ".description" inside search breaks down on screen less 769px.

[Example](https://jsfiddle.net/xrgs3eyu/) (reduce the size of the browser window)
image

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -256,7 +256,7 @@
   .ui.modal .image.content {
     flex-direction: column;
   }
-  .ui.modal .content > .image {
+  .ui.modal > .content > .image {
     display: block;
     max-width: 100%;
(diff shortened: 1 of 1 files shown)


lubber-de:
Please adjust the jsfiddle (its still the default fiddle without any code)

byte916:
> Please adjust the jsfiddle (it's still the default fiddle without any code)

Sorry, I fixed it

byte916:
@lubber-de maybe you’re right, I don’t know exactly all the places where the description in modal can be used

lubber-de:
@all-contributors please add @byte916 for code

PR #1385, opened 2020-03-30
[Build] Allow individual components to be defined in config

hugopeek:
## Description

Individual components will have their .css/.js file generated under /dist/components, but are not included in semantic.css and semantic.js.

This is useful for including these components in your project only if you need them, while keeping the file size of the main .css/.js file as small as possible.

## Testcase

To make it work, create a new array in your semantic.json:
```
{
    "components": [
        ...
    ],
    "individuals": [
        form,
        modal,
        step
    ]
}
```

Don't forget to remove these items from the regular "components" array.

This change should be fully backwards compatible. It only changes the src variable if an "individuals" array has been defined in the config.

### DIFF
--- a/tasks/build/css.js
+++ b/tasks/build/css.js
@@ -127,6 +127,13 @@ function buildCSS(src, type, config, opts, callback) {
     src      = config.paths.source.definitions + '/**/' + config.globs.components + '.less';
   }
 
+  if (globs.individuals !== undefined) {
+    const individuals = config.globs.individuals.replace('{','');
+    const components = config.globs.components.replace('}',',').concat(individuals);
+
+    src = config.paths.source.definitions + '/**/' + components + '.less';
+  }
+
   const buildUncompressed       = () => build(src, type, false, config, opts);
   buildUncompressed.displayName = 'Building uncompressed CSS';
 
--- a/tasks/build/javascript.js
+++ b/tasks/build/javascript.js
@@ -99,6 +99,13 @@ function buildJS(src, type, config, callback) {
     src      = config.paths.source.definitions + '/**/' + config.globs.components + (config.globs.ignored || '') + '.js';
   }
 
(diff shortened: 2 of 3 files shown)


exoego:
Thanks for opening this feature request.
I was thinking this idea too.

My use-case is for `emoji` and `icon` modules: those 2 modules are quite huge, so I want to use them as separate CSS/JS file only when needed.

lubber-de:
@hugopeek Would you like to adjust the Docs accordingly here: https://fomantic-ui.com/introduction/build-tools.html#semanticjson 🙂

hugopeek:
@lubber-de Sure, [done](https://github.com/fomantic/Fomantic-UI-Docs/pull/203) :)

@exoego Yeah exactly, I was going back and forth with adding them to semantic.json, building, then removing them again and then building again to keep the file size down.. And now semantic.css hit 1MB in a project, so time to act!

It's not a silver bullet solution btw. Some components depend on each other, so you can't just take them out of the main files without disabling some functionality. But that's for another day ;)

lubber-de:
> @lubber-de Sure, [done](https://github.com/fomantic/Fomantic-UI-Docs/pull/203) :)

Thanks 😎 Please see my comment https://github.com/fomantic/Fomantic-UI-Docs/pull/203#issuecomment-606073378

hugopeek:
@lubber-de @exoego Thanks for merging guys.

I found a small glitch however: the gulp watch task is no longer updating only the files that changed.. It goes through the whole list now.

I'm looking at how to correct this now. Shall I make a new PR with the fix later?

exoego:
@hugopeek 
Don't worry, this feature is not published yet in stable version.

> Shall I make a new PR with the fix later?

Ofcourse !!
I am very excited to use this feature.

hugopeek:
PR #1408 fixes the issues with gulp watch..

PR #1400, opened 2020-04-06
Remove max-width definition from comments

Sorien:
it makes problems when you have to override it to fill parent element, you can use `max-width: unset` but it's not supported in IE/Edge


lubber-de:
Please rebase your branch to develop to reduce the amount of commits. We are not merging PRs into master. Thanks

Sorien:
@lubber-de done, sorry for this

PR #1416, opened 2020-04-13
feat(segment): border to top instead of bottom for vertical variant

hugopeek:
## Description

This makes it easier to remove the border if needed, because you're able to target the second segment with + in CSS.

For example, in case you have 2 adjacent vertical segments with a custom class of 'white', you want to remove the border and top padding to make them appear as 1. This currently requires a few overrides, but with a top border this would become:

```
.ui.vertical.segment.white + .white {
    border-top: none;
    padding-top: 0;
}
```

If you look at the Table component, you'll see that table rows are using this too:
```
/* Table Row */
.ui.table > tr > td,
.ui.table > tbody > tr > td {
  border-top: @rowBorder;
}
.ui.table > tr:first-child > td,
.ui.table > tbody > tr:first-child > td {
  border-top: none;
}
```

## Testcase

https://jsfiddle.net/drjuto9f/


lubber-de:
We have to adjust the docs example for this , because the change to `first-child` won't work in the current example anymore, as the first vertical segment is not the first child of the parent container.
A simple wrapping to another div tag will do it, so i'll prepare a docs pr for that

lubber-de:
Docs fixed by https://github.com/fomantic/Fomantic-UI-Docs/pull/208

lubber-de:
@fomantic/maintainers 
I wonder if we should consider this a breaking change and postpone this to 2.9?

exoego:
This might be considered as a small breaking change, if user override styles as @hugopeek explained.

Issue #1437, opened 2020-04-25
[modal] When two modals with allowMultiple: true are opened, clicking the dimmer on one will close both regardless of the closable setting on the first


Merlin04:
# Bug Report
I have two modals. Both of them have `allowMultiple: true`, and the first has `closable: false`. When I open the first one and then the second one, clicking on the dimmer closes both modals, not just the second one.

## Steps to reproduce

1. Create two modals with `allowMultiple: true`, with the first one having `closable` set to false
2. Open the first one, then the second one
3. Click on the dimmer

## Expected result

Only the second modal closes, with the first remaining open

## Actual result

Both modals close, even though I set `closable` on the first modal to false

## Testcase
https://jsfiddle.net/15bd4x20/

## Version

2.8.4

Merlin04:
From looking at the source code this appears to be intended behavior, there should be an option to disable it.

PR #1438, opened 2020-04-25
[modal] Add option to disable clicking the dimmer calling hideAll when allowMultiple = true

Merlin04:
## Description

Add a new option, multipleHideAll, which when false makes clicking the dimmer when multiple modals have allowMultiple = true hide only the modal at the front. When true (the default) the previous behavior is used, i.e. hiding all of the open modals. 

## Testcase
I couldn't figure out how to create a testcase on JSFiddle but I tested it locally and it worked. 

## Closes

#1437

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -313,7 +313,7 @@ $.fn.modal = function(parameters) {
             ;
             if(!isInModal && isInDOM && module.is.active() && $module.hasClass(className.front) ) {
               module.debug('Dimmer clicked, hiding all modals');
-              if(settings.allowMultiple) {
+              if(settings.allowMultiple && settings.multipleHideAll) {
                 if(!module.hideAll()) {
                   return;
                 }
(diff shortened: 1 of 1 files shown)


lubber-de:
🤔 I am unsure, if your proposed solution is the way we should fix this. As you already pointed out (and as it is implemented for single modals already) the hideAll Method should check each of the other modals if its closable before hiding it via behaviors. (and should stop immediatly if one of them is not closable)

Merlin04:
That sounds like a better way to do it. I'm trying to implement it but it seems like `hideAll` iterates over the modals from the one in the back, which would prevent this from working. Is there any reason it is done that way?

lubber-de:
Mmh, the current code reverses the found jquery modal elements to make sure it starts from the very top... Doesn't this work?

Merlin04:
It seems that it goes through them from the one furthest back, so I wouldn't be able to do the checking starting from the top.

PR #1440, opened 2020-04-28
feat(button): split tertiarylinepadding variable

hugopeek:
I wanted to set a horizontal padding of 0 for tertiary buttons, so I changed tertiaryLinePadding to '0.5em 0'. But it resulted in a build error, because this variable is also used for calculating tertiary button margin.

I think it's better so split this variable into vertical and horizontal values and then use the vertical value for calculating the margins.

This also aligns better with other button variables (verticalPadding and compactVerticalPadding).


src/themes/default/elements/button.variables
```diff
@@ -253,7 +253,8 @@
 /*-------------
    Tertiary
 -------------*/
-@tertiaryLinePadding: 0.5em;
+@tertiaryVerticalPadding: 0.5em;
+@tertiaryHorizontalPadding: 0;
```
lubber-de, line 257: The default value for horizontal padding should also be 0.5em to keep the same overall padding for the default theme as before
otherwise:
before: `padding: 0.5em;`
after: `padding: 0.5em 0;`

hugopeek: @lubber-de Ah yes, my apologies. I just pasted my own preference there. Adjusted now to 0.5em.

hugopeek:
@lubber-de No problem to wait for 2.9.x. Thank you for maintaining FUI!

ko2in (approved):

lubber-de:
@all-contributors please add @hugopeek for code

PR #1467, opened 2020-05-14
[Build] Support gulp calls to find semantic.json from outside fomantic folder

### LINKED ISSUE #1466: Improved support for lerna / yarn workspaces
# Feature Request
We are using fomantic-ui 2.8.4 in a lerna/yarn environment. The build process in a subfolder fails because it is not able to find the semantic.json configuration and falls back to some default - which in our case is of no help since our semantic.json is customized.

The reason is that in the **user.js task** **require-dot-file** ist called with only the single parameter "semantic.json". In this case require-dot-file will start looking up for the indicated file starting at its very own location in the lerna global node_modules which is wrong - in a single file structure with a single project in a folder this happens to be correct.

There is an ugly workaround using the nohoist feature of lerna/yarn workspaces. I think a far more better solution would be to add **process.cwd()** as a second parameter to the require-dot-file call in the user.js task. In fact this may be a more consistent and defensice programming style than rely on some starting folder randomly choosen by some external (to fomantic-ui) code.

I can't estimate the importance of this issue and but since I found no other report using lerna/yarn workspaces may be rare. But especially in combination of an application development in parallel to some library the concept is quite helpful. For example in our case we have a couple of applications using a common theming which has been outsourced to a separate npm module.

lubber-de:
## Description
In some situations (for example in yarn workspaces), mostly within custom build chains, the users semantic.json cannot be found because the requireDotFile starts from within the fomantic folder.
Instead , the second parameter should be given as `process.cwd()` to support starting the search in a different folder

Thanks to @JMS-1 for the solution, which i just adopted by this PR

## Closes
#1466

### DIFF
--- a/tasks/config/user.js
+++ b/tasks/config/user.js
@@ -28,7 +28,7 @@ var
 
 try {
   // looks for config file across all parent directories
-  userConfig = requireDotFile('semantic.json');
+  userConfig = requireDotFile('semantic.json', process.cwd());
 }
 catch(error) {
   if(error.code === 'MODULE_NOT_FOUND') {
--- a/tasks/install.js
+++ b/tasks/install.js
@@ -57,7 +57,7 @@ var
 module.exports = function (callback) {
 
   var
-    currentConfig = requireDotFile('semantic.json'),
+    currentConfig = requireDotFile('semantic.json', process.cwd()),
     manager       = install.getPackageManager(),
     rootQuestions = questions.root,
     installFolder = false,


JMS-1:
Great, thanks!

JMS-1:
Sorry, me again: I updated to 2.8.5 in my lerna environment and now it nearly works as expected. BUT it seems as if the version indicator can not be read from the semantic.json: in the generated output the version number in the first line (formerly 2.8.4) is now empty. Surprisingly the reminder of the semantic.json esp. all paths work as configured.

lubber-de:
🤔 , the line you mentioned is supposed to only be only for creating releases and not to be used for general usage/build of your local instance.
However, also reagarding the missing version attribute you mentioned, please create new issues for those, so we can track them more easily

JMS-1:
Ok, as soon as I get some time. But FYI: I added the , process.pwd() and the version number reappeared after the build.

lubber-de:
Ok, i'll prepare a PR for this 😉

lubber-de:
Should be fixed now by #1497

PR #1473, opened 2020-05-19
fix(dropdown): the nested dropdown menus do not behave correctly

### LINKED ISSUE #1369: Rigth -> left dropdown menu
The submenu on the right is not droping down on the left side. Am I doing something wrong? 

https://jsfiddle.net/7rpb6zsu/1/

ko2in:
## Description
When the dropdown is used as simple dropdown without JavaScript, it doesn't support more than two nested level of sub-menus. It always exposes only two levels of sub -menus.

And despite the sub-menu declares the value '**left**' for the class name, it always appears rightward. It should respect the value '**left**' from the class name and should be appeared leftward.

When the dropdown menu is invoked with JavaScript, the last dropdown in the right menu doesn't behave properly and the sub-menus inside the dropdown overlaps each others. They should appear in the correct location like other dropdown menus.

## Testcase
Bug version: https://jsfiddle.net/w3nejghr/
Fix version: https://jsfiddle.net/k31zf0h2/

## Closes
#1369

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -318,10 +318,9 @@
 }
 
 /* Automatically float dropdown menu right on last menu item */
-.ui.menu .right.menu .dropdown:last-child .menu,
-.ui.menu .right.dropdown.item .menu,
-.ui.buttons > .ui.dropdown:last-child .menu {
-  left: auto;
+.ui.menu .right.menu .dropdown:last-child .menu:not(.left),
+.ui.menu .right.dropdown.item .menu:not(.left),
+.ui.buttons > .ui.dropdown:last-child .menu:not(.left) {
   right: 0;
 }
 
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.less
```diff
@@ -1097,7 +1096,7 @@ select.ui.dropdown {
 
   .ui.dropdown > .left.menu .menu,
   .ui.dropdown .menu .left.menu {
-    left: auto;
+    left: auto !important;
```
exoego, line 1099: Can we remove `!important` and use repeating selector (e.g. `.left.left.left.menu`) to give more specificity ?
`!important` makes it harder to customise theme.

lubber-de: I thought the same first, but i suspect the increased specificity for the other settings (especially `right: 100%`) within that selector will probably cause trouble in other situations then. margin+border are already !important too so to be clean we should then try to remove them as well..

lubber-de: Did some quick check and unfortunately the proposed `.left.left.left` does not work and would need adjustments elsewhere too.
Lets face this in a separate PR if needed

PR #1479, opened 2020-05-22
[Popup] Make pre-existing popup works properly regardless of jQuery versions

### LINKED ISSUE #1476: [Popup] Wrong positioning with jQuery 3.3.0+
# Bug Report

Positionig of popup in table seems fine with jQuery 3.2.1 (2017-3-20),
but broken with jQuery 3.3.0 (2018-1-19) and later includling latest 3.5.1 (2020-5-4).

So FUI users may affect if they are updating jQuery for security/bugfix/new features.
Maybe [some changes in jQuery 3.3.0](https://blog.jquery.com/2018/01/19/jquery-3-3-0-a-fragrant-bouquet-of-deprecations-and-is-that-a-new-feature/) affects popup module?🤔 

## Steps to reproduce

1. Open https://jsfiddle.net/xfgm5au0/

## Expected result

With jQuery 3.2.1: Correct position ✔️ (each popup shows right after the trigger icon)
https://jsfiddle.net/xfgm5au0/2/

image

## Actual result

With jQuery 3.3.1: Wrong position ❌ (every popup shows topside of table)
https://jsfiddle.net/xfgm5au0/
image

The wrong positioning seems reproducible with jQuery 3.3.0 and latest 3.5.1.

## Version
2.8.5

exoego:
## Closes

#1476


src/definitions/modules/popup.js
```diff
@@ -510,9 +510,14 @@ $.fn.popup = function(parameters) {
               $popupOffsetParent = module.get.offsetParent($popup),
               targetElement      = $target[0],
               isWindow           = ($boundary[0] == window),
-              targetPosition     = (settings.inline || (settings.popup && settings.movePopup))
+              usePosition        = settings.inline || (settings.popup && settings.movePopup),
+              targetOffset       = $target.offset(),
+              targetPosition     = usePosition
                 ? $target.position()
-                : $target.offset(),
+                : targetOffset,
```
lubber-de, line 517: `$target.position()` will never be used anymore later on, even if `usePosition` is true . The later code always refers to targetOffset later.
So we may use either targetPosition or targetOffset

I suggest to reduce it to either
```javascript
// remove the extra targetOffset declaration
//targetOffset       = $target.offset(),
targetPosition = $target.offset();
....

top     : usePosition ? Math.max(targetPosition.top - offsetParentOffset.top, 0) : targetPosition.top,
left    : usePosition ? Math.max(targetPosition.left - offsetParentOffset.left, 0) : targetPosition.left,

```

or

```javascript
// remove targetPosition at all and always work with targetOffset
targetOffset       = $target.offset(),
//targetPosition = $target.offset();
....

top     : usePosition ? Math.max(targetOffset.top - offsetParentOffset.top, 0) : targetOffset.top,
left    : usePosition ? Math.max(targetOffset.left - offsetParentOffset.left, 0) : targetOffset.left,

```

lubber-de:
@exoego  While you think about my comment, please rebase to develop 😉

src/definitions/modules/popup.js
exoego: Thanks !!
I chosed the 2nd suggestion.
Because `targetPosion = $target.offset()` seems weird: it stores `offset` instead of `position`...

exoego:
I also rebased the base branch to develop.

src/definitions/modules/popup.js
exoego: I also simplified (and optimized) code, in a same fashion for other variables.

PR #1491, opened 2020-05-31
fix(dimmer): page dimmer in firefox results in misaligned modals

### LINKED ISSUE #1490: [Modal] when dragging inside an input in long modals (modal with browser scrollbar), modal scrolls up in some browsers
# Bug Report
when modal is long and has a scrollbar, if it is in the scrolled down position (.scrollTop() not zero), then if you drag the mouse in an input, the modal scrolls up only in 'Firefox Browser 76.0.1' .
if the testing modal isn't long enough in your browser to have a scrollbar, then just duplicate some Fomantic-UI rows to the HTML (before the input)

## Expected result
i expect that the long modal doesn't scroll up when dragging inside an input in all browsers. not in just some. the bug occurs in Firefox Browser version 76.0.1

## Actual result
the long modal scrolls up when dragging inside an input (when the modal is scrolled down)

## Testcase
  How to create an example:
   1. Open the following JSFiddle - https://jsfiddle.net/hzk7vwa4/7/
   2. the minimum necessary HTML,CSS and JavaScript is already added in the example .If the testing modal isn't long enough in your browser to have a scrollbar, then just duplicate some Fomantic-UI rows to the HTML (before the input)
   3. Do the necessary changes if you want
   4. Click "Save" at the top
   5. Copy the URL of your fiddle and link it here

## Screenshot
Untitled

one browser version that this bug occurs is in the picture below
photo_2020-05-31_20-22-05

## Version
2.8.4

lubber-de:
## Description
Firefox messes around with (large) modals in certain situations. For example when trying to drag inside an input inside a modal, it scrolls to the top of the modal. Even worse on Mac where the whole modals overlays other div containers...

Reason for this is the `perspective` setting for the page dimmer. It seems this was introduced some years back to also fix some firefox issues, so i don't dare to remove this setting completely to avoid possible other, currently unknown, issues in probably other browsers as well.

## Testcase
- Try to drag in the below input inside of the modal

### Broken (in Firefox only)
- The modal scrolls to the top (on macos it also overlaps the upper CSS container of jsfiddle (tried using browerstack)
https://jsfiddle.net/hzk7vwa4/7/

### Fixed
- Nothing happens even in Firefox now
https://jsfiddle.net/hy2fzkca/

## Closes
#1490

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -151,6 +151,9 @@
     transform-style: @transformStyle;
     perspective: @perspective;
     transform-origin: center center;
+    &.modals {
+      -moz-perspective: none;
+    }
   }
 }
 


src/definitions/modules/dimmer.less
```diff
@@ -151,6 +151,9 @@
     transform-style: @transformStyle;
     perspective: @perspective;
     transform-origin: center center;
+    &.modals {
+      -moz-perspective: none;
```
exoego, line 155: If `@perspective` had been added to resolve firefox-specific issue, why not settings `@perspective` to `none` for all browsers ?

lubber-de: That's what I also thought at first, but that might break existing code, because I am not 100% sure this setting was really only implemented because of Firefox issues.
I believe the perspective setting was done for performance reasons too whenever other transforms within the page dimmer occur, like transitions. To be in the safe side (at least within 2.8.x) I decided for that extra setting for modals only. Removing the perspective setting at all, as I stated in the issue text, might lead into unknow visual/performance breaking code.
When looking at the git history, the perspective setting was needed because transition animations were not working in Firefox at all too, so removing it entirely is risky for the 2.8 branch.
We could either postpone this pr to 2.9 and remove the setting entirely, but this would most probably need lots of testing, because, as said, withing a page dimmer (which can also be used without modals!) anything can exists.

exoego: Thanks for detail.
Thats sounds fair reason for me!

PR #1492, opened 2020-06-01
[Calendar ]Fix timezone issue when time is not given and fix json dates for Firefox

### LINKED ISSUE #1462: [calendar] Parsing date in format yyyy-mm-ddT00:00:00 in Firefox
# Bug Report
On Firefox (I'm currently using the latest version 76.0.1), parsing a date such as 2018-08-13T00:00:00 results in August 13, 2000.
## Steps to reproduce
See #935 
1. Using Firefox, supply the date in the format 2018-08-13T00:00:00 using set date 

## Expected result
August 13, 2018

## Actual result
August 13, 2000

## Testcase
https://jsfiddle.net/scblack/qzLvkxm9/1/

## Additional Information
In the date parser, assuming text is '2018-08-13T00:00:00'.

It changes to lowercase with the following line:
```javascript
text = ('' + text).trim().toLowerCase();
```
After this line, text is '2018-08-13t00:00:00', which is then used to attempt to create a date:
```javascript
var textDate = new Date(text);
```
After this, textDate is an invalid date. Firefox doesn't seem to handle a lowercase 't'. when parsing the date.

## Version
2.8.4

lubber-de:
## Description
When using JSON dates (YYYY-MM-DD**T**HH:II:SS) Firefox parses those dates wrong, because the 'T' gets lowercased and Firefox does not support that
Additionally when living in areas with a large timezone difference Javascript does not always return the correct Date when only "2020-05-11" is given, but the current timezone already is beyond the current day resulting in 2020-05-12 instead. 
To fix this i added a fixed midnight time in case dateonly pickers are used with format YYYY-MM-DD, so the resulting JS Date will always return the exact date that was provided to the calendar module.

## Testcase
https://jsfiddle.net/8b3rta4u/

## Closes
#1462

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1342,16 +1342,20 @@ $.fn.calendar.settings = {
       if (!text) {
         return null;
       }
-      text = ('' + text).trim().toLowerCase();
+      text = String(text).trim();
       if (text.length === 0) {
         return null;
       }
+      if(text.match(/^[0-9]{4}[\/\-\.][0-9]{2}[\/\-\.][0-9]{2}$/)){
+        text += ' 00:00:00';
+      }
       // Reverse date and month in some cases
-      text = settings.monthFirst ? text : text.replace(/[\/\-\.]/g,'/').replace(/([0-9]+)\/([0-9]+)/,'$2/$1');
+      text = settings.monthFirst || !text.match(/^[0-9]{2}[\/\-\.]/) ? text : text.replace(/[\/\-\.]/g,'/').replace(/([0-9]+)\/([0-9]+)/,'$2/$1');
       var textDate = new Date(text);
       if(!isNaN(textDate.getDate())) {
         return textDate;
       }
+      text = text.toLowerCase();
 
       var i, j, k;
       var minute = -1, hour = -1, day = -1, month = -1, year = -1;


src/definitions/modules/calendar.js
```diff
@@ -1342,16 +1342,20 @@ $.fn.calendar.settings = {
...
+        text += ' 00:00:00';
+      }
       // Reverse date and month in some cases
-      text = settings.monthFirst ? text : text.replace(/[\/\-\.]/g,'/').replace(/([0-9]+)\/([0-9]+)/,'$2/$1');
+      text = settings.monthFirst || !text.match(/^[0-9]{2}[\/\-\.]/) ? text : text.replace(/[\/\-\.]/g,'/').replace(/([0-9]+)\/([0-9]+)/,'$2/$1');
       var textDate = new Date(text);
       if(!isNaN(textDate.getDate())) {
         return textDate;
       }
+      text = text.toLowerCase();
```
exoego, line 1358: I am asking because I am not familiwar with calendar module.
What is a reason that  `toLowerCase()` is required?

lubber-de: In further parsing, calendar also checks for month names and 12h clock definitions (am/pm) against existing array strings. To support sensitive spelling, the text is set to lowercase before.

exoego: Ah, I see!

bterkuile:
Adding the `' 00:00:00'` seems to fix the timezone issue. Confirmed for latest chromium and Firefox on Ubuntu :+1:

PR #1497, opened 2020-06-02
fix(build): missing extra parameter for current directory in release task leads into missing build number

### LINKED ISSUE #1498: [BUG] In a lerna environment the version from semantic.json is not copied into the generated artifacts
# Bug Report
I'm using fomantic-ui in a lerna environment. The root folder contains three folders for libs, servers and apps. In libs only a single subfolder [let's call it themes] contains a project using fomantic-ui. The lerna package.json makes sure that fomantic-ui is localized to this project using "nohoist": ["**/fomantic-ui"] to avoid messing with the update scripts - there is no semantic.json in the root folder which seems to bedazzle the script which runs the semantic.json initialisation. If inside the libs/themes folder the build script is called the version from  semnatic.json is not copied to the generated artifacts as semantic.css.

## Steps to reproduce
[report.zip](https://github.com/fomantic/Fomantic-UI/files/4716758/report.zip)

0. OS Ubuntu 2004
1. make sure node.js 12.17.0 and yarn 1.22.4
2. unpack the zip
3. in the lerna root folder (lerna.json) call yarn
4. in the libs/themes folder call yarn build
5. check src/semantic/dist/semantic.css line 2: version is missing

## Expected result
The version number from semantic.json schould be included in all generated artefacts.

## Actual result
The version number is missing.

## Testcase
n/a, may be fixed with https://github.com/fomantic/Fomantic-UI/pull/1497

## Screenshot (if possible)
n/a

## Version
2.8.5

lubber-de:
## Description
By #1467 we added the current directory location. The release task was still missing that extra parameter. According to @JMS-1 this lead into a missing version number in semantic.json when the release is build locally from a different path

## Closes
#1498

### DIFF
--- a/tasks/config/project/release.js
+++ b/tasks/config/project/release.js
@@ -15,7 +15,7 @@ var
 *******************************/
 
 try {
-  config = requireDotFile('semantic.json');
+  config = requireDotFile('semantic.json', process.cwd());
 }
 catch(error) {}
 


JMS-1:
I added the bug description: https://github.com/fomantic/Fomantic-UI/issues/1498

PR #1505, opened 2020-06-07
fix(dropdown): fix some bugs introduce by the previous commit (#1473)

### LINKED ISSUE #1502: [dropdown] simple dropdown menu items positioned incorrectly
# Bug Report
As of v2.8.5, simple dropdown menus will now go offscreen if the parent dropdown label is shorter than the child dropdown labels. This issue does not occur for Javascript dropdown menus.

PR that caused the issue: #1473 

## Steps to reproduce
1. Create a menu with a simple dropdown
2. Set the dropdown header / parent label to something short, such as an icon
3. Put a child item in the menu with a long label
4. The child item will not fit on the screen as it goes to the right instead of the left.

## Expected result
Simple dropdown menu items should be positioned correctly to fit on the screen

## Actual result
Simple dropdown menu items don't fit on the screen currently.

## Testcase
Bug version (v2.8.5): [https://jsfiddle.net/18jogrs9/](https://jsfiddle.net/18jogrs9/)
Working version (v2.8.4): [https://jsfiddle.net/t4ymose7/](https://jsfiddle.net/t4ymose7/)

## Screenshot
v2.8.5
image

v.2.8.4
image

## Version
2.8.5

ko2in:
## Description
This PR fixes the first level of simple dropdown inside the right menu was cropped and out of the screen when the parent menu item has short text, and the dropdown doesn't specify the class name `left`.

To avoid this, all dropdown inside the right menu will be forced to appear leftward by default. The dropdown can also appear leftward or rightward by having `left` or `right` class name for it's nested dropdown menus.

This PR also fix the dropdown inside the buttons which are positioned incorrectly.

### Broken
https://jsfiddle.net/18jogrs9/
https://jsfiddle.net/usfh3tpw/1/

### Fixed
https://jsfiddle.net/ymo0f9st/

## Closes
#1502
#1514

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -318,9 +318,10 @@
 }
 
 /* Automatically float dropdown menu right on last menu item */
-.ui.menu .right.menu .dropdown:last-child .menu:not(.left),
-.ui.menu .right.dropdown.item .menu:not(.left),
-.ui.buttons > .ui.dropdown:last-child .menu:not(.left) {
+.ui.menu .right.menu .dropdown:last-child > .menu:not(.left),
+.ui.menu .right.dropdown.item > .menu:not(.left),
+.ui.buttons > .ui.dropdown:last-child > .menu:not(.left) {
+  left: auto;
   right: 0;
 }
 
@@ -1096,7 +1097,7 @@ select.ui.dropdown {
 
   .ui.dropdown > .left.menu .menu,
   .ui.dropdown .menu .left.menu {
-    left: auto !important;
+    left: auto;
     right: 100%;
     margin: @leftSubMenuMargin !important;
     border-radius: @leftSubMenuBorderRadius !important;
@@ -1324,6 +1325,13 @@ select.ui.dropdown {
     top: 0 !important;
     left: 100%;
     opacity: 1;
+  }
+  .ui.simple.dropdown > .menu > .item:active > .left.menu,
+  .ui.simple.dropdown .menu .item:hover > .left.menu, 
+  .right.menu .ui.simple.dropdown > .menu > .item:active > .menu:not(.right), 
+  .right.menu .ui.simple.dropdown > .menu .item:hover > .menu:not(.right) {
+    left: auto;
+    right: 100%;
   }
   & when (@variationDropdownDisabled) {
     .ui.simple.disabled.dropdown:hover .menu {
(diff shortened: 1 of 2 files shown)


ko2in:
I updated the left and right margin of sub menu `.ui.dropdown .menu .menu` within one rule, since the sub menu are absolute positioned element, and it's safe to declare the negative margin for both left and right together for leftward or rightward direction without affecting on the neighbor element.

Previously in dropdown.variables: `@subMenuMargin: 0 0 0 @subMenuDistanceAway;`. And now: `@subMenuMargin: 0 @subMenuDistanceAway;`.

I don't think it's necessary to specify individually for leftward and rightward.

And I also would like to suggest removing the declaration `important` from the following from `menu.less`.
```
/* Left Floated */
.ui.menu:not(.vertical) .left.item,
.ui.menu:not(.vertical) .left.menu {
  display: flex;
  margin-right: auto !important;  
}
/* Right Floated */
.ui.menu:not(.vertical) .right.item,
.ui.menu:not(.vertical) .right.menu {
  display: flex;
  margin-left: auto !important;
}
```

This especially effects on the simple dropdown in the right menu, that need to specify to appear rightward direction with the class name `right menu`, and the negative margin for the left side is overruled by the above declarations, and which gives the slightly different position for the rightward dropdown in the right menu.

## With `!important`
Menu-1

## Without `!important`
Menu-2

I'm not sure how much it can effect and break the backward compatibility. By testing with this fiddle (https://jsfiddle.net/ymo0f9st/1/) doesn't break anything and works as expected.

ko2in:
@lubber-de Thanks for the details. Then, I'll test with the other modules around and it doesn't break anything, I'll send a PR for this.

hakeem86:
The previous version did not fix the left submenu on iOS, worked fine on other os/browsers.

Is this fix tested on iOS?

lubber-de:
> Is this fix tested on iOS?
The above mentioned fixed jsfiddle runs fine as expected on iPad Pro, iOS 13, Safari

ceisele-r:
@ko2in is your fix going to be released, soon? The menu out of screen bug prevents us from updating to the latest version.

lubber-de:
@ceisele-r v 2.8.6 (which included this fix) is supposed to be released tomorrow :)
The current master branch was already updated last night

ceisele-r:
Great, thanks @lubber-de

lubber-de:
@ceisele-r 2.8.6 has been released just now 🎉  We hear your voices 😄

ceisele-r:
@lubber-de thank you very much!

PR #1507, opened 2020-06-07
[Label] Basic Labels had different height and adjusted (attached) image labels 

### LINKED ISSUE #1430: Height difference b/w label and basic label
# Bug Report
The heght of the `<a class="ui label">Testing</a>` is `26px` whereas `<a class="ui basic label">Basic</a>` has a height of `28px` 

## Expected result
Both labels should have the same height. Since the size is mentioned using a different property.
Is it intentional? Shouldn't be they have the same height?

## Testcase
https://codesandbox.io/s/quiet-surf-w7upu?file=/index.html

## Screenshot (if possible)
https://imgur.com/ZYS6Wt1

## Version
2.8.4

lubber-de:
## Description
- `basic labels` have a border which increases the height in comparison to usual labels.
When inline together the line height gets increased
- Because of that, a `basic image label` was not properly supported
- All attached labels were not really attached to the boundary of a segment (as every other `attached` component is doing)  . That was especially visible when a `basic attached label` was used

### Before
Watch the height and width difference on basic labels as well as the misaligned image labels and unattached borders in attached labels
https://jsfiddle.net/lubber/wt0heq5k/60/

### After
https://jsfiddle.net/lubber/wt0heq5k/99/

### Before
Watch the height and width difference on basic labels as well as the misaligned image labels and unattached borders in attached labels
image
image
image

### After
image
image
image
image

## Closes
#1430

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -174,7 +174,7 @@ a.ui.label {
 .ui.segment:not(.basic) > .ui.bottom.attached.label {
   margin-bottom: @attachedOffset;
 }
-.ui.segment:not(.basic) > .ui.left.attached.label {
+.ui.segment:not(.basic) > .ui.attached.label:not(.right) {
   margin-left: @attachedOffset;
 }
 .ui.segment:not(.basic) > .ui.right.attached.label {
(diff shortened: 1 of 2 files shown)


prudho:
Probably a little bit off topic, but the last image seems to be rounded at position `bottom left`, which seem wrong.

lubber-de:
>  but the last image seems to be rounded at position `bottom left`, which seem wrong.

@prudho  You've got Eagle Eyes 😆 
Fixed by https://github.com/fomantic/Fomantic-UI/pull/1507/commits/1922bf7197ea87e472fe6a6f96a54d4562215e8a now. Changed screenshot and updated jsfiddle accordingly
image
@exoego Please review again

PR #1508, opened 2020-06-08
fix(calendar): change to listen to click event by default

### LINKED ISSUE #1506: [Calendar] Calendar is not opened again when clicking on input
# Bug Report
When the calendar input already has focus and the calendar isnt shown, clicking on the input again does not open the calendar. You have to un-focus the input and click it again to have the calendar shown.
I think clicking on the input should open the calendar, think of users accidently closing the calendar and trying to reopen it by clicking on the input field.
Or think of the calendar input as some date jumper in a calendar. You could use it several times in a row without focussing anything else than the input.

## Steps to reproduce
1. Go to https://fomantic-ui.com/modules/calendar.html#/definition
2. Click the calendar input. The calendar opens.
3. Click the calendar input again / select a date. The calendar closes.
4. Click the calendar input again. **The calendar does not open again**.

## Expected result
1. Go to https://fomantic-ui.com/modules/calendar.html#/definition
2. Click the calendar input. The calendar opens.
3. Click the calendar input again / select a date. The calendar closes.
4. Click the calendar input again. **The calendar opens again**.

## Version
2.8.5

Browser: Chrome

Lutonite:
## Description

As I said in #1506, to reflect the docs and with the changes from #1439, the calendar module should always listen to the click event listener by default.

## Testcase

[Ly1ng3u0/4/](https://jsfiddle.net/Ly1ng3u0/4/)

## Closes

#1506


147loch:
Goes with #1509 to close #1150.
Corresponding changes to the docs made in fomantic/Fomantic-UI-Docs#214.

src/definitions/modules/calendar.js
```diff
@@ -116,7 +116,7 @@ $.fn.calendar = function(parameters) {
               module.set.maxDate($module.data(metadata.maxDate));
             }
             module.setting('type', module.get.type());
-            module.setting('on', settings.on || ($input.length ? 'focus' : 'click'));
+            module.setting('on', settings.on || 'click');
```
exoego, line 119: If "click" is "always" expected, I think "click" should be added regardless of settings, something like:


```suggestion
            if (settings.on) module.setting('on', settings.on);
            if (settings.on.indexOf('click') >= 0) module.setting('on', 'click'); // avoid duplicate click
```

lubber-de: Sorry, maybe it's too late for me today 😴 , but I don't get the point of your suggested change..
```javascript
if (settings.on) module.setting('on', settings.on);
```
is totally senseless IMHO, because it's equal to `settings.on = settings.on` 😉 

```javascript
if (settings.on.indexOf('click') >= 0) module.setting('on', 'click');
```
What does this do/prevent? the only valid options are 'click', 'hover' and 'focus' anyway.

147loch: `settings.on` is not a required parameter for calendars. However, the module needs `module.setting('on')` to be set in order to know what event it should listen to.

I think that, with this change, if the user doesn't include `on: (...)` to his calendar definition then the calendar will have no `module.setting('on')` which will end up breaking. (The docs themselves don't use the `on` option, so IMHO doing this would end up breaking the docs even more.

With my commit, the user could have his on option set to `null`, `undefined` or `false` and it would set it to the default: `click`. That's how I envisioned it, `click` should not be expected, just the default.

exoego: Oh, sorry, please forget my above comment !!

I reazlized that I misunderstood what `settings.on` is.
I just looked the surface of changes from mobile GitHub app... but now I browsed code on IDE and docs.

lubber-de:
@all-contributors please add @Lutonite for code

PR #1531, opened 2020-06-22
[Calendar] Avoid unnecessary redraw which slows down the module

lubber-de:
## Description
Whenever a calendar popup is shown the calendar gets redrawn several times even if nothing was changed!

## Testcase
Open the console, i adjusted the 2.8.6 calendar as well as the fixed one to show a console.log whenever the calendar gets redrawn.
Click into the date field, but do not select anything. Instead click outside of the field, so the calendar closes again

### Broken
The calendar gets redrawn all the time. Even 2 times when no date was selected!
https://jsfiddle.net/lubber/gzsxr8t1/2/

### Fixed
The calendar only gets redrawn when something changed (date value, mode view (day/time))
https://jsfiddle.net/lubber/gzsxr8t1/3/

### Broken
calendarrefreshbroken

### Fixed
calendarrefreshfixed

## Closes
https://github.com/mdehoog/Semantic-UI-Calendar/issues/130

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -157,7 +157,7 @@ $.fn.calendar = function(parameters) {
             var onShow = function () {
               //reset the focus date onShow
               module.set.focusDate(module.get.date());
-              module.set.mode(settings.startMode);
+              module.set.mode(module.get.validatedMode(settings.startMode));
               return settings.onShow.apply($container, arguments);
             };
             var on = module.setting('on');
@@ -197,6 +197,7 @@ $.fn.calendar = function(parameters) {
               date = parser.date($input.val(), settings);
             }
             module.set.date(date, settings.formatInput, false);
+            module.set.mode(module.get.mode(), false);
           }
         },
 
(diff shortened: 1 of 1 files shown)


src/definitions/modules/calendar.js
```diff
@@ -157,7 +157,7 @@ $.fn.calendar = function(parameters) {
             var onShow = function () {
               //reset the focus date onShow
               module.set.focusDate(module.get.date());
-              module.set.mode(settings.startMode);
+              module.set.mode(module.get.validatedMode(settings.startMode));
```
exoego, line 160: Maybe stupid question.
Why not `module.set.mode(modue.get.mode())` here ?
`module.get.mode()` attempts to get `$module.data(metadata.mode)` in addition to `settings.startMode`.

lubber-de: This is needed, because at `onShow` we have to make sure, that always the startingMode will be used. If we would use `module.get.mode()` here and a value at $module.data(metadata.mode) is indeed found _and_ you left the previous opening of the calendar by cancelling it in a different  (for example year selection) mode, then the calendar will popup with that last remembered mode, which is wrong.
I adjusted a jsfiddle to demonstrate this here https://jsfiddle.net/lubber/gzsxr8t1/5/

### Wrong Behavior
calendarkeepstartmode_bad

### Correct behavior
calendarkeepstartmode_good

exoego: Thanks for detailed explanation 👍 
That helps understanding current calendar module implementation.

PR #1537, opened 2020-06-25
[Popup] left/right center tooltips were misaligned on Firefox (posix)

### LINKED ISSUE #442: [Popup] extra tooltop appears underneath on CSS only
# Bug Report

## Steps to reproduce
1. Create a new tooltip (CSS only) https://fomantic-ui.com/modules/popup.html#tooltip
2. Hover mouse over in FireFox and wait 1-2 seconds

## Expected result
Tooltip should show as normal

## Actual result
Tooltip shows and then it appears an extra tooltip opens underneath it after a few seconds

## Testcase
https://jsfiddle.net/9850tpo4/

## Screenshot
image

## Version
2.7.1

lubber-de:
## Description
`left/right centered` tooltips got misaligned border after a second.
Thanks to @ko2in, [he figured out ](https://github.com/fomantic/Fomantic-UI/issues/442#issuecomment-649751669)it's obviously a bug in firefox on posix systems (linux/mac) when a scaling factor in transform is used (obviously in combination with translateY because all other tooltip positions work and also use the scale transform)
It does not happen at all on Windows

## Testcase
Hover over the below labels 'left center' and 'right center' using firefox on linux or mac. 

### Broken
The tooltip slightly moves to the bottom and the tooltips lower border gets 'ghosted' after 1-2 seconds
https://jsfiddle.net/bg7j2Lzo

### Fixed
The text still moves slightly, but the border is not ghosting anymore
https://jsfiddle.net/c3qorphu/

## Screenshot
|Broken|Fixed|
|-|-|
|ffpopup_broken|ffpopup_fixed|

## Closes
#442

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -189,6 +189,7 @@
   [data-tooltip][data-position="left center"]:hover:after,
   [data-tooltip][data-position="right center"]:hover:after {
     transform: translateY(-50%) scale(1) !important;
+    -moz-transform: translateY(-50%) scale(1.0001) !important;
   }
   [data-tooltip][data-position="top left"]:after,
   [data-tooltip][data-position="top right"]:after,


lubber-de:
Wait... @ColinFrick already prepared a much better fix using `-moz-transform-style: preserve-3d;` in his repo but never prepared a PR for this...
https://github.com/ColinFrick/Fomantic-UI/commit/38f46ce20a46854e2a500060b05ac7eb7fd3aee8

This does not even do the text moving anymore...
@ColinFrick Why did you hesitate to provide a PR?
It seems to work much better than the scaling hack approach. See
https://jsfiddle.net/c3qorphu/1/

ColinFrick:
@lubber-de as far as I remember (and commented on that specific commit) I was not able to fix all cases.
It was a long time ago, so I can't say what didn't work...

lubber-de:
@ColinFrick Then, i would like to adopt your fix instead to this PR if you don't mind (or would like to create a PR on your own)
@ko2in Can you confirm,  all your testcases will also work by applying
```
-moz-transform-style: preserve-3d;
```
instead of the `scale(1.0001)` change?

What i love about Colins solutions is the fact that not even the text moves, so its a cleaner fix for me

ColinFrick:
@lubber-de sure, go ahead

lubber-de:
I'll leave the PR as it is, because the preserve-3d solution was not working in @ko2in other fiddles and even after reloading the last fiddle with colins fix, it was not working again anymore (maybe thats what he also recognized back then (?))
So, sorry for the inconvenience and please review the PR as it is

ko2in:
As far as I tested, `-moz-transform-style: preserve-3d;` doesn't work yet.

PR #1547, opened 2020-06-30
[Slider] operation sometimes returns not rounded value

### LINKED ISSUE #1545: [slider] Rounded value is not always rounded
# Bug Report
Despite the code I've spotted in the slider.js codebase to avoid a nasty JS rounding bug, it looks like there are still some cases where this happens.

## Steps to reproduce

Move a slider several times to get the value bug. It roughly happens 1 time out of 5 in my app.

## Expected result

A value rounded to the closest step (ex: 2.1)

## Actual result

(ex: 2.1000000001)

## Testcase

https://jsfiddle.net/dfj7urbz/

## Screenshot

Capture d’écran 2020-06-30 à 09 03 57

## Version

2.7.8
2.8.6

TheJltres:
## Description
I changed the return result, beacause of this https://floating-point-gui.de/
variables values where rounded when they arrive thanks to rounded difference, but the result of difference and module.get.min whas not rounded

## Testcase
Issue jsfiddle: https://jsfiddle.net/dfj7urbz/

## Closes
#1545

### DIFF
--- a/dist/components/slider.js
+++ b/dist/components/slider.js
@@ -789,7 +789,6 @@ $.fn.slider = function(parameters) {
             }
             // Use precision to avoid ugly Javascript floating point rounding issues
             // (like 35 * .01 = 0.35000000000000003)
-            difference = Math.round(difference * precision) / precision;
             module.verbose('Cutting off additional decimal places');
             return Math.round((difference + module.get.min()) * precision) / precision;
           },
--- a/dist/components/slider.js
+++ b/dist/components/slider.js
@@ -791,7 +791,7 @@ $.fn.slider = function(parameters) {
             // (like 35 * .01 = 0.35000000000000003)
             difference = Math.round(difference * precision) / precision;
             module.verbose('Cutting off additional decimal places');
-            return difference + module.get.min();
+            return Math.round((difference + module.get.min()) * precision) / precision;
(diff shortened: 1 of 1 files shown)


dist/components/slider.js
```diff
@@ -791,7 +791,7 @@ $.fn.slider = function(parameters) {
             // (like 35 * .01 = 0.35000000000000003)
             difference = Math.round(difference * precision) / precision;
```
ColinFrick, line 792: Could this line be removed, when the rounding is done later now?

TheJltres: I dont know what happen if difference is not rounded on return, I will take a look right now

TheJltres: Yes, result seems correct even when JS is doing strange things.

lubber-de:
Damn... i totally missed, that you did the changes in the `dist` folder... 🙄 

could please recreate the same PR and do the changes in the `src/definitions/modules` folder instead?

TheJltres:
I had to create other PR, sorry.
Is branch correct now?

lubber-de:
> I had to create other PR, sorry.
> Is branch correct now?

All fine now in #1548. Thank you!  We should have recognized this earlier

PR #1190, opened 2019-11-20
[Flag] use emoji flags with (auto) sizing 

lubber-de:
## Description
This PR replaces the sprite flag backgrounds by the new emoji flags keeping existing code bases backward compatible (you still use `<i class="uk flag"></i>` as before)

Additionally
- almost all countries are available now (famfamfam sprite misses a lot of flags)
- flags can now be sized
- they are autosized by current font-size if  no size class is given
- the old sprite approach was kept into an additional theme `famfamfam` which one can easily switch back by changing the flag theme in theme.config

## Screenshot
image
image
image

## Closes
#1172 
#660

PR #1549, opened 2020-07-01
[Flag] Added catalan flag

TheJltres:
## Description
Added Catalan flag.
In theory there's other flag, but I can't see it.

## Testcase
https://jsfiddle.net/TheJltres/9bxmc7rq/1/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6926

Is the correct folder? :roll_eyes:

### DIFF
--- a/src/themes/default/elements/flag.overrides
+++ b/src/themes/default/elements/flag.overrides
@@ -146,6 +146,10 @@ i.flag.bz:before,
 i.flag.belize:before {
   background-position: 0 -884px;
 }
+i.flag.es.ca:before,
+i.flag.catalonia:before {
+  background-position: 0 -936px
+}
 i.flag.ca:before,
 i.flag.canada:before {
   background-position: 0 -910px;


ColinFrick:
Thanks for the PR. This looks fine to me, but we'll replace the current sprite-based flag with a new system in 2.9 #1190
So I'm unsure if we'll merge it now @fomantic/maintainers

lubber-de:
@TheJltres  Although your PR is technically valid, there does not exist either an official country code nor an existing unicode for the catalonian flag. That said it's not officially supported by any emoji provider.

I don't want to have politial discussion here, but if we merge this now, the flag won't appear in 2.9 anymore until it gets an offical country code or is at least supported by the official unicode list for country flags or subdivision flags, so that tweemoji, joypixels and others are supporting it.
https://unicode.org/emoji/charts/full-emoji-list.html#country-flag

TheJltres:
It's okey, as you saw now if you want you can add like I answered in semantic issue

lubber-de:
Closing, because of the above mentioned reasons we will not merge this. @TheJltres also agreed to this decision via discord
If somebody still needs this vor the 2.8.x branch, just add the 3 LOC from this PR to your own code

PR #1562, opened 2020-07-10
feat(icon): make fontface import and icons optional separately

### LINKED ISSUE #1560: Can @font-face about font-family: 'Icons' be made optional too?
I use webfontloader Javascript for loading font: https://github.com/typekit/webfontloader . In this case @font-face about icons is not needed to be included within the resulting css. There are variables like @importOutlineIcons and @importBrandIcons, it seems that they can be used to exclude @font-face definitions, but there is no variable such as @importIcons that could do the same for 'Icons' font. So, the original source needs to be touched, it seems to me inevitable, which is not good.

Could you add an option for removing @font-face about 'Icons' font?

lubber-de:
## Description
This PR allows to create some icon groups on demand by the usual switches in `variation.variables` and make the font-face loading optional and independent of the icon declarations themselves.
It is now possible to move the font-face loading out of fomantic ui in favor of loading the fonts separately via a 3rd party lib such as  https://github.com/typekit/webfontloader 

To keep consistency, the creation of brand and outline icons are now also separated into creating the icons and loading the fontface.

Because of this logic the usage of the variable `importOutlineIcons` and `importBrancIcons` (and the new `importIcons`) now only applies for the font-face loading and creation of icons is dependent on the new `variationIconBrand` and `variationIconOutline` switches

It's a breaking change, but it's a more consistent logic.

Related PR for the icon creation script is https://github.com/fomantic/create-fomantic-icons/pull/108

## Closes
#1560


src/themes/default/elements/icon.variables
```diff
@@ -7,6 +7,7 @@
 ---------------*/
 
 /* Solid Icons */
+@importIcons: true;
```
exoego, line 10: Forgot to add `@imporBrandIcons` ?

lubber-de: No, `@importBrandIcons` is already declared here
https://github.com/fomantic/Fomantic-UI/blob/2580bc67fd3e6865ee4dc99c83308c525ef559f7/src/themes/default/elements/icon.variables#L36

exoego: I see

Issue #1558, opened 2020-07-08
[Search] Search field with non-rounded borders


smartm0use:
[Search module](https://fomantic-ui.com/modules/search.html) page has a lot of example and all of them show search field with rounded borders.

I see that if I include search module within a form with `ui form` class then I get a [squared field](https://fomantic-ui.com/elements/input.html).

So, there is a way to have squared search field in case of use out of a form?

lubber-de:
Sure, just override the border-radius
See https://jsfiddle.net/lubber/3dowLgsx/2/
![image](https://user-images.githubusercontent.com/18379884/86934605-30413600-c13c-11ea-85cd-1b3f7e5eb306.png)

smartm0use:
What about improving Search module by adding a specific class to have it?

lubber-de:
🤔 , unsure, because
- you can already define this according to your needs by changing the variable value at https://github.com/fomantic/Fomantic-UI/blob/0269a8213ac2ff438965c6a82503e7561d2024ae/src/themes/default/modules/search.variables#L13
- providing a `ui squared search` seems to be inconsistent. The search module is individual on its own. Having the same style like other input fields is already covered when the search field is used inside a form (as you found out already)
@fomantic/maintainers  What do you think?

ko2in:
Yes. I agreed that it's unnecessary to have specific class name. Since the search field commonly have different style with goggle icon, loading icon etc, and I don't think it should mix with other input field styles. It should have individual customization on it's own.

smartm0use:
>  Having the same style like other input fields is already covered when the search field is used inside a form (as you found out already)

In that case how to have api search within a search field inside a form?
I tried but it doesn't work: https://jsfiddle.net/d7a6zv5h


> Yes. I agreed that it's unnecessary to have specific class name. Since the search field commonly have different style with goggle icon, loading icon etc, and I don't think it should mix with other input field styles. It should have individual customization on it's own.

In my case I have a search field among other form fields. It is a form for personal data and users can search for cities in the search field...

ko2in:
@smartm0use You need to wrap your search module within `ui search` container and the input field need the class name `prompt`. Here is fiddle (https://jsfiddle.net/4L2khmyu/).

Please read [documentation](https://fomantic-ui.com/modules/search.html) for complete details and usages.

smartm0use:
@ko2in Your code doesn't work. It works because you add a CSS rule to have it squared.

[...]

PR #1576, opened 2020-07-14
feat(search): squared borders search

smartm0use:
## Description
In F-UI we have search as standalone box and as field within a form. In the first case search has rounded borders, in the second case it has squared borders. This change makes possibile to have standalone search box with squared borders, using a dedicated class.

## Testcase
Here is a JSFiddle: https://jsfiddle.net/tnfdpv5e

## Screenshot (if possible)
image

## Closes

#1558

### DIFF
--- a/src/definitions/modules/search.less
+++ b/src/definitions/modules/search.less
@@ -393,7 +393,7 @@
 & when (@variationSearchScrolling),
   (@variationSearchShort),
   (@variationSearchLong) {
-  
+
   /*-------------------
        Scrolling
   --------------------*/
@@ -526,7 +526,7 @@
 
 & when (@variationSearchSquared) {
   .ui.squared.search .prompt {
-    border-radius: .28571429rem;
+    border-radius: @promptBorderRadiusSquared;
   }
 }
 
--- a/src/themes/default/globals/variation.variables
+++ b/src/themes/default/globals/variation.variables
@@ -473,6 +473,7 @@
 @variationSearchCategory: true;
 @variationSearchLoading: true;
 @variationSearchAligned: true;
+@variationSearchSquared: true;
 @variationSearchFluid: true;
 @variationSearchShort: true;
 @variationSearchLong: true;
--- a/src/themes/default/modules/search.variables
+++ b/src/themes/default/modules/search.variables
@@ -11,6 +11,7 @@
 @promptPadding: (@promptVerticalPadding + ((1em - @promptLineHeight) / 2)) @promptHorizontalPadding;
 @promptBorder: 1px solid @borderColor;
 @promptBorderRadius: @circularRadius;
+@promptBorderRadiusSquared: @absoluteBorderRadius;
 @promptColor: @textColor;
 @promptTransition:
   background-color @defaultDuration @defaultEasing,


src/definitions/modules/search.less
```diff
@@ -520,6 +520,16 @@
   }
 }
 
+/*--------------
+    Squared
+---------------*/
+
+& when (@variationSearchSquared) {
+  .ui.squared.search .prompt {
+    border-radius: .28571429rem;
```
lubber-de, line 529: - Please exchange the hardcoded value by a variable `@promptBorderRadiusSquared`
- In the file `search.variables` create an entry `@promptBorderRadiusSquared: @absoluteBorderRadius;`

smartm0use: > Please create a proper PR description (yes, your code is obvious, but we usually don't guess functionality just by viewing the code)

Ehm... how to add it now? General speaking, how to edit a PR here, including file changes?

> At least a jsfiddle to prove your change works should be provided and always helpful.
> Friendlier for reviewers 

Do you mean [this](https://jsfiddle.net/lubber/3dowLgsx/2)?

smartm0use:
> Please create a proper PR description (yes, your code is obvious, but we usually don't guess functionality just by viewing the code)

Ehm... how to add it now? General speaking, how to edit a PR here, including file changes?

> At least a jsfiddle to prove your change works should be provided and always helpful.
> Friendlier for reviewers 

Do you mean [this](https://jsfiddle.net/lubber/3dowLgsx/2)?

src/definitions/modules/search.less
smartm0use: I saw a lot of occurences of `.28571429rem` in the source code of the project, so I thought that it could be right.

lubber-de: where did you see those raw values? Raw values are only inside of generated CSS files, but should never appear inside of .less files

lubber-de:
> > Please create a proper PR description (yes, your code is obvious, but we usually don't guess functionality just by viewing the code)
> 
> Ehm... how to add it now? General speaking, how to edit a PR here, including file changes?

Just edit your description 😉 
editdesc

For additional changes, just push new commits to the branch you created your PR in

> 
> > At least a jsfiddle to prove your change works should be provided and always helpful.
> > Friendlier for reviewers
> 
> Do you mean [this](https://jsfiddle.net/lubber/3dowLgsx/2)?

Yes, for example. Whenever you create a PR there is a prefilled template shown where you should add your info for "Description", "Testcase", "Screenshot", "Closes". This really helps reviewers.

src/definitions/modules/search.less
smartm0use: You are right, sorry!

smartm0use:
Ok, I have updated PR description and changes. Thank you.

ko2in:
PR looks good to me. But, as I've already agreed to @lubber-de , it's confusing to have such name.

lubber-de:
After looking through the whole source, I think we should name this `square` search instead of `squared` out of consistency.

We already have `square ad`, `header square icon`,  `placeholder square image`,  and `square embed` but we never got any `squared`
image

@fomantic/maintainers Opinion?

smartm0use:
I agree.

lubber-de:
Again the question about if we want to merge this and **still** the questionable name `square` 

The field itself is not square, it's rectangle to be exact, every other "square" names i mentioned above really display the according items as a square, so for the search component this PR just wants to have its border more edgy, but the whole component isn't square,,,
@fomantic/maintainers I really would like to have an opinion from everybody

smartm0use:
So the class name could be "edgy" :)

Anyway, before making a new commit I will wait for other opinions.

exoego:
IMHO, variation classes should be added when it make sense to use those when designing single app/web site.
It make sense to have `square image` and `rectangular image` for different contexts of single app, for example.

However, regarding search module, I think it is not a good idea to have 2 different styles, rounded and squared (edgy), within same app.
Such inconsistent design confuses end-users.
So, if FUI user want to reduce border-radius, I suggest to use customizing theme or use input inside form.

(This is my personal design principle.
I also understand that having variations helps FUI users so they have less customization)

ko2in:
In FUI v3, I would like to propose to control such UI styles with specific variant class names, example in this case, the search field border radius should be changeable by using a variant like `rounded-sm` , `rounded-lg` etc.

Some UI libraries like Tailwind CSS and Twitter Bootstrap are good examples for such implementation.

ColinFrick:
I concur. If someone needs their search fields to be like the input fields they can set the border radius like proposed in the pull request.

lubber-de:
I am afraid, but, your suggestion does not fit into the fui naming logic and (most of) the other maintainers also think the same. It's a design requirement to have a different border radius, so it has to be done as a LESS variables change in a different theme rather than implementing a new fui variant name as [explained above](https://github.com/fomantic/Fomantic-UI/pull/1576#issuecomment-660019854)

We left this PR open for a long time, but for housekeeping i am closing this now, as we are not going to merge it anyway.

PR #1578, opened 2020-07-15
[Popup] Remove text-shadow explicitly on tooltip to avoid inheritance from parent

### LINKED ISSUE #1577: [Popup] data-tooltip inherits parent's font styling unintendedly  ?
# Bug Report

It seems that tooltip for button inherits some font styling (`text-shadow` at least).
This may cause accidental cosmetic issue, if `text-shadow`'s color fits fot button, but not for tooltip.

## Steps to reproduce

1. Create a `.ui.button` with `data-tooltip` 
2. Hover the button

## Expected result
Tooltip (popup) should show text in the style for popup.
Font styling (e.g. `text-shadow`) should not be inherited from parent element.

## Actual result

Tooltip inherits `text-shadow` of normal button, `.ui.button` without color variation.

## Testcase
https://jsfiddle.net/Lf1qouje/2/

Note) This testcase intendedly override button's text-shadow with red.

## Screenshot (if possible)
image

## Version

2.8.6

exoego:
## Description
To avoid accidental inheritance of styling from parent element, `text-shado: none` is explicitly set for `[data-tooltip]` popup.

## Testcase
https://jsfiddle.net/kpbo7qjg/

## Screenshot (if possible)
image

## Closes
#1577

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -108,6 +108,7 @@
     position: absolute;
     text-transform: none;
     text-align: left;
+    text-shadow: none;
     white-space: nowrap;
 
     font-size: @tooltipFontSize;


src/definitions/modules/popup.less
```diff
@@ -91,6 +91,7 @@
     position: absolute;
     content: '';
     font-size: @medium;
+    text-shadow: none;
```
lubber-de, line 94: Should we put this into a variable and also add this to the usual `.ui.popup` definition? Just in case somebody wants to add a text-shadow to popup and tooltip in a theme

ko2in: And this should also be inside `:after` pseudo where the content is set. `:before` would not apply the text-shadow because it has no content and it's only to create arrow.

exoego: Oops. 
Moved to `:after` pseuod selector.

IMHO, adding new variable for customizing `text-shadow` is unnecessary, since text-shadow in popup/tooltip are less readable.
Even without new variable, users could override text-shadow if they actually needs text-shadow.

PR #1584, opened 2020-07-16
[Segment] Horizontal segments' rounded corner

ko2in:
## Description
The horizontal segments should have the rounded corner, but the rounded corners seem to be covered by the background of the segment inside of it.

The first and last segment inside the horizontal segments now adjust the border-radius to appear the rounded corner from the horizontal segments.

## Testcase
Before: https://jsfiddle.net/fax0kqd4/
After: https://jsfiddle.net/1ho0egx3/

## Screenshot
Before:
Before

After:
After


ko2in:
Sorry. I've selected wrong branch. I'll pull new PR base on develop.

PR #1596, opened 2020-07-18
[Calendar] Adjust cell tooltip positions according to popup position to avoid it being hidden

lubber-de:
## Description
The default tooltip position of cell tooltips is `left center`. This leads into situations where the tooltip is cut to the left.
On first instantiation of a calendar popup the position of tooltips is not correctly fetched because the popup module hasn't adjusted its pointer position classes yet. Because of this ,the direction of possible cell tooltips is wrongly calculated. 
If the same calendar is opened a second time, this does not occur anymore, because the popup module has now adjusted its position.

This PR now adjusts any existing cell tooltip positions inside the calendar whenever the popup get's shown (the popup module always (re-)calculates its position then). So this not only fixes it on first instantiation but also in case the calendar position changes while scrolling , so the popup will possibly to displayed in another direction.

As this is not a module setting but a decision of the module itself, this is a backward compatible change which just fixes the behavior.
Btw. it wasn't happening before, because the calendar popup was called twice (thus correcting it's position), but that was now fixed by #1439

### Broken
https://jsfiddle.net/lubber/afer3s7b/1/

### Fixed
https://jsfiddle.net/lubber/afer3s7b/7/

## Screenshot
calendarTooltips

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -141,12 +141,16 @@ $.fn.calendar = function(parameters) {
               $container = $('<div/>').addClass(className.popup)[domPositionFunction]($activatorParent);
             }
             $container.addClass(className.calendar);
-            var onVisible = settings.onVisible;
+            var onVisible = function () {
+              module.refreshTooltips();
+              return settings.onVisible.apply($container, arguments);
+            };
             var onHidden = settings.onHidden;
             if (!$input.length) {
               //no input, $container has to handle focus/blur
               $container.attr('tabindex', '0');
               onVisible = function () {
+                module.refreshTooltips();
                 module.focus();
                 return settings.onVisible.apply($container, arguments);
               };
@@ -436,6 +440,16 @@ $.fn.calendar = function(parameters) {
 
         refresh: function () {
           module.create.calendar();
+        },
+
+        refreshTooltips: function() {
+          var popupPos = $container.hasClass('left') ? 'left' : 'right';
+          var popupPosOpposite = popupPos === 'left' ? 'right' : 'left';
+          $container.find('td[data-position*="'+popupPos+'"]').each(function () {
+            var cell = $(this);
+            var tooltipPosition = cell.attr('data-position');
+            cell.attr('data-position',tooltipPosition.replace(popupPos,popupPosOpposite));
+           });
         },
 
         bind: {
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -182,6 +182,7 @@ $.fn.calendar = function(parameters) {
             if ($activator.length && !settings.inline) {
               return;
             }
+            settings.inline = true;
             $container = $('<div/>').addClass(className.calendar).appendTo($module);
             if (!$input.length) {
               $container.attr('tabindex', '0');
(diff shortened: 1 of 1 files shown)


lubber-de:
@ko2in 🤔 Yes....ok, i think we should rather fix the auto positioning of the popup on first instantiation somehow without it being called twice again....i'll invetigate into this tomorrow

lubber-de:
@ko2in Thanks for the hint, i now changed the PR accordingly. Please review again

ko2in:
The popup on the date 26 of the second calendar is out of screen and not visible for me. And I also notice that the calendar appears too far away from the date field.

Calendar

lubber-de:
@ko2in I cannot reproduce with my latest jsfiddle. Did you use that ? https://jsfiddle.net/lubber/afer3s7b/6/

lubber-de:
mmh, you are using firefox, right? Firefox is getting the new IE6 .... :(
In Chrome it should work... however my PR should be totally browser independent..i'll investigate

ko2in:
Same happens in my Chromium too. I've tested in Brave, Chromium and Firefox. All have the same results. But, the calendar position stays far away in Brave browser only. Chrome and Firefox looks OK.

ko2in:
After refreshing several times, the calendar position looks OK in Brave browser. Only the popup issue persists.

**Firefox:**
Firefox

**Chromium:**
Chromium

**Brave:**
Brave

lubber-de:
This is totally weird....As you can see by the PR's changes i did not touch the calendar positioning at all (and i still cannot reproduce that...)

lubber-de:
ok, found the reason, will investigate how to properly fix it.

lubber-de:
@ko2in Please review again, i think i got it now. 
New jsfiddle https://jsfiddle.net/lubber/afer3s7b/7/

lubber-de:
> The method name `refreshTooltips` may be confusing, better use `relocateTooltips`?

It's an internal method, so i really don't care. If renaming this method is a condition to get this PR approved by you, then i will of course rename it 😆 

Lets see, what the other @fomantic/maintainers say, before i am adjusting the PR again. 😉

src/definitions/modules/calendar.js
```diff
@@ -141,12 +141,16 @@ $.fn.calendar = function(parameters) {
               $container = $('<div/>').addClass(className.popup)[domPositionFunction]($activatorParent);
             }
             $container.addClass(className.calendar);
-            var onVisible = settings.onVisible;
+            var onVisible = function () {
+              module.refreshTooltips();
+              return settings.onVisible.apply($container, arguments);
```
exoego, line 146: Why `function.apply`?
It seems that the passed `$container` (used as `this` of `onVisible` function) is never used inside `onVisible`.

lubber-de: But it **could**, because the `onVisible` method is a dynamic setting which can be adjusted by the user. It's already used this way everywhere else in the code. Also the docs tell, that the context is the active content

image

exoego: Ah, sorry my bad...
I overlooked that it is `settings.onVisible`, not `onVisible` created in this code.

PR #1602, opened 2020-07-24
[Calendar] trigger native change event on date selection

### LINKED ISSUE #1251: onChange event not fired when date picked from datepicker
Expected behaviour:
* When a date is picked from the datepicker in the Calendar component, the "change" event is fired on the updated input.

Actual behaviour:
* When I manually modify the date and navigate away from the field (blur the input), the change event is fired as expected.
* When I choose a date from the Calendar datepicker, change event is not fired neither immediately nor navigating away from the field.
* When I manually modify the date in the input, then open the datepicker, close it without picking anything and then blur the field, change event does not occur.

Tested with 2.8.3.

lubber-de:
## Description
The native `change` event should be triggered whenever a date is selected from the picker and the input field blurs.
Currently it is only triggered when the dates input field has been changed manually and the field blurs.

I adopted the same method for this as in `checkbox` or `dropdown`

### Broken
https://jsfiddle.net/lubber/asdjkwpo/11/

### Fixed
https://jsfiddle.net/lubber/asdjkwpo/11/

## Screenshots
|Broken|Fixed|
|-|-|
|calendarnativechange_wrong|calendarnativechange_right|

## Closes
#1251

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -215,10 +215,10 @@ $.fn.calendar = function(parameters) {
         trigger: {
           change: function() {
             var
-                events       = document.createEvent('HTMLEvents'),
                 inputElement = $input[0]
             ;
             if(inputElement) {
+              var events = document.createEvent('HTMLEvents');
               module.verbose('Triggering native change event');
               events.initEvent('change', true, false);
               inputElement.dispatchEvent(events);
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -211,6 +211,20 @@ $.fn.calendar = function(parameters) {
           }
         },
 
+        trigger: {
+          change: function() {
+            var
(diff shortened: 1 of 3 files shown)


src/definitions/modules/calendar.js
```diff
@@ -211,6 +212,20 @@ $.fn.calendar = function(parameters) {
           }
         },
 
+        trigger: {
+          change: function() {
+            var
+                events       = document.createEvent('HTMLEvents'),
```
exoego, line 218: I think `events` should be created inside `if`, so it wouldn't consume CPU and memory when `inputElement` not found.

lubber-de: I agree, but for consistency i will change this in checkbox/dropdown as well (because the method is the same there)

lubber-de: Done by 4d6363fde4198b67f9c779c2bd7d4913846247f9 🙂

PR #1612, opened 2020-07-28
[Form] dirty state check failed for multiple variant of native select tag

### LINKED ISSUE #930: [Form Validation] onDirty event unpredictable behavior
When we have a multiple select field inside form (with dropdown module or not) onDirty event behavior would be unpredictable.

https://jsfiddle.net/vtbyr8o1
Try to focus textarea first, then re-run and try to select/unselect options.

Thanks!

lubber-de:
## Description
The `fieldDirty` check failed whenever a dropdown, converted from a `select` tag, was part of the form **and** was defined as `multiple`
The comparison of currentValue and initialValue failed for those dropdowns, because the `.val()` method returned arrays instead of strings for select-multiple fields. Dropdowns made out of `div` were working, because those use an input field, which holds a string.
Raw `==` comparisons of two arrays are never true, so that fiel was always considered dirty.

## Testcase
Just click into the textarea to trigger the forms dirty check ...
### Broken
... the dirty check for the form will immediatly fail (considered dirty), even though no value has been changed, and triggers the onDirty event
https://jsfiddle.net/vtbyr8o1

### Fixed
... the dirty check successfully considers the form to be untouched, thus not triggering the onDirty event.
https://jsfiddle.net/lubber/sjnv3wkp/

## Closes
#930

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -366,14 +366,14 @@ $.fn.form = function(parameters) {
             var initialValue = $el.data(metadata.defaultValue);
             // Explicitly check for null/undefined here as value may be `false`, so ($el.data(dataInitialValue) || '') would not work
             if (initialValue == null) { initialValue = ''; }
+            else if(Array.isArray(initialValue)) {
+              initialValue = initialValue.toString();
+            }
             var currentValue = $el.val();
             if (currentValue == null) { currentValue = ''; }
             // multiple select values are returned as arrays which are never equal, so do string conversion first
-            if(Array.isArray(currentValue)) {
+            else if(Array.isArray(currentValue)) {
               currentValue = currentValue.toString();
-            }
-            if(Array.isArray(initialValue)) {
-              initialValue = initialValue.toString();
             }
             // Boolean values can be encoded as "true/false" or "True/False" depending on underlying frameworks so we need a case insensitive comparison
             var boolRegex = /^(true|false)$/i;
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -368,7 +368,13 @@ $.fn.form = function(parameters) {
             if (initialValue == null) { initialValue = ''; }
             var currentValue = $el.val();
             if (currentValue == null) { currentValue = ''; }
-
+            // multiple select values are returned as arrays which are never equal, so do string conversion first
+            if(Array.isArray(currentValue)) {
```
exoego, line 372: I think this `if` branch can be combined with the above `if`, since `currentValue` is not an `Array` if it was `null`.

```suggestion
            else if(Array.isArray(currentValue)) {
```

src/definitions/behaviors/form.js
```diff
@@ -368,7 +368,13 @@ $.fn.form = function(parameters) {
...
+              currentValue = currentValue.toString();
+            }
+            if(Array.isArray(initialValue)) {
```
exoego, line 375: Same as above for this `if` branch.

src/definitions/behaviors/form.js
lubber-de: Done by 57ba588f8ef8ba8c78e5c772f697c009767da86b (also did that for `initialValue`)

src/definitions/behaviors/form.js
lubber-de: Yes, done by 57ba588f8ef8ba8c78e5c772f697c009767da86b

PR #1614, opened 2020-07-29
[Input,Form] Action or labeled inputs in grouped fields were not styled properly with state classes

### LINKED ISSUE #1238: [Form] action or labeled input doesn't have error style in grouped fields
# Bug Report

Action or labeled input doesn't have error style in grouped fields

## Testcase
https://jsfiddle.net/dutrieux/a2po6z54/

## Screenshot (when possible)
image

## Version
v2.8.2

lubber-de:
## Description
`action` or `labeled` input fields don't have state border style for the buttons or labels when inside grouped fields

#### Broken
https://jsfiddle.net/lubber/nudwt41r/4/

#### Fixed
https://jsfiddle.net/lubber/nudwt41r/3/

## Screenshot
|Before|After|
|-|-|
|image|image|

## Closes 
#1238

### DIFF
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -437,26 +437,26 @@
     @state: replace(@key, '@', '');
     @borderColor: @formStates[@@state][borderColor];
 
-    .ui.form > .field.@{state} > .ui.action.input > .ui.button,
-    .ui.form > .field.@{state} > .ui.labeled.input:not([class*="corner labeled"]) > .ui.label,
+    .ui.form .field.@{state} > .ui.action.input > .ui.button,
+    .ui.form .field.@{state} > .ui.labeled.input:not([class*="corner labeled"]) > .ui.label,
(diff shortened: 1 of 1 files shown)


lubber-de:
@ko2in Lets cover your concerns in a separe PR, because the basic label used in the examples to the right are using box-shadow, while the error color is usually using border. Considering such fix also has to make sure normal labels or basic buttons are used this might need a much larger adjustment as currently expected.
The general question would also be: Should the left button get a right border or should the right label has the left border removed? (As said above, the latter would be nearly impossible as long as basic label uses box shadow)

ko2in (approved):
LGTM. Sure, we can look into this with a separate issue later. I was reviewing the CSS source for this input element last week and they've the most complicated CSS logic and they don't look easy to add some adjustments without effecting the other components.

lubber-de:
@fomantic/maintainers Time to review this if not done already?

ColinFrick (approved):

PR #1615, opened 2020-07-29
[Form] isValid should not delete error state

### LINKED ISSUE #923: [Form Validation] error class remove on fields when use "is valid" behavior
# Bug Report

'error' class remove on fields when use "is valid" behavior

## Steps to reproduce
https://jsfiddle.net/02gfdbx9/
1. click on submit button (the error class is set on all fields with a error)
2. click on the button "$('.ui.form').form('is valid');" -> the error class is remove on all fields

## Expected result
the error class on field should not be remove when we use $('.ui.form').form('is valid');

## Actual result
the error class on field are remove when we use $('.ui.form').form('is valid');

## Testcase
https://jsfiddle.net/02gfdbx9/

## Screenshot (when possible)
is_valid
## Version
2.7.7

lubber-de:
## Description
the `is valid` behavior deletes possible existing error states on a field, although it shouldn't.
That's because the field validator always removes the error state by default (before probably adding it again) and did not respect a `showErrors` setting. showErrors however is only meant to show **new** errors. If it is false it should not touch any existing error state class.

I now also added the `showErrors` parameter as an option to the `is valid` behavior in case someone definately wants to show errors on **not already validated** fields.

As the parameter is optional the change is fully backwards compatible.

## Testcase
- Click on "submit" to invalidate the form
- Click on the second button to call `is valid`

#### Broken
- The invalid field loses its error state
https://jsfiddle.net/02gfdbx9/

#### Fixed
- The invalid field stays invalid
https://jsfiddle.net/lubber/1978duwp/4/

## Screenshot
|Before|After|
|-|-|
|isvalid_wrong|isvalid_right|

## Closes
#923

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -338,18 +338,18 @@ $.fn.form = function(parameters) {
           blank: function($field) {
             return String($field.val()).trim() === '';
           },
-          valid: function(field) {
+          valid: function(field, showErrors) {
             var
               allValid = true
             ;
             if(field) {
               module.verbose('Checking if field is valid', field);
-              return module.validate.field(validation[field], field, false);
+              return module.validate.field(validation[field], field, !!showErrors);
             }
             else {
               module.verbose('Checking if form is valid');
               $.each(validation, function(fieldName, field) {
-                if( !module.is.valid(fieldName) ) {
+                if( !module.is.valid(fieldName, showErrors) ) {
                   allValid = false;
                 }
               });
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -338,18 +338,18 @@ $.fn.form = function(parameters) {
           blank: function($field) {
             return String($field.val()).trim() === '';
           },
-          valid: function(field) {
+          valid: function(field, showErrors) {
```
exoego, line 341: Should `showErros` be added to docs ?
https://fomantic-ui.com/behaviors/form.html#/settings

lubber-de: yes, will do this as a docs PR. Tagging as "awating docs" Thanks ;)

lubber-de:
Added optional parameters to docs by https://github.com/fomantic/Fomantic-UI-Docs/pull/223

PR #1622, opened 2020-07-31
[Sidebar] Avoid scrollbar on non body pushable (visible on bottom sidebars)

### LINKED ISSUE #1607: [Sidebar] "pushable" should add overflow-y: hidden?
# Bug Report

Based on https://github.com/Semantic-Org/Semantic-UI-React/issues/3884.

## Steps to reproduce
1. Open a repro Fiddle
2. Mention that scroll exists for `bottom` variation (for `top` everything is okay)

## Expected result
No scroll

## Actual result
Scroll exists

## Testcase
https://jsfiddle.net/dzpc73rb/5/

## Screenshot (if possible)
image

## Version
2.8.6

lubber-de:
## Description
When a `bottom sidebar` was used inside non body pushable context, it was always added to the context divs height, making the vertical scrollbar appear. As the body will still provide a possible scrollbar it's save to remove the inner pushable vertical scrollbar.

Reason for the behavior is the 100% `translate3d()` setting to hide the sidebar which adds its content to the parent div. For `left`/`right` the `overflow-x:hidden` is already set and for `top` it's a negative position of `-100%`, so the change of overflow-y is not affected in the other cases.

#### Broken
https://jsfiddle.net/lubber/vd2ecqgL/9/

#### Fixed
✔️ Ok, same as above
https://jsfiddle.net/lubber/vd2ecqgL/69/

✔️  when the inner segment should be scrollable itself
https://jsfiddle.net/lubber/vd2ecqgL/77/

✔️  when inside grid columns
https://jsfiddle.net/lubber/vd2ecqgL/78/

## Screenshot
|Broken|Fixed|
|-|-|
|bottom_sidebar_broken|bottom_sidebar_fixed|

## Closes
#1607

### DIFF
--- a/src/definitions/modules/sidebar.less
+++ b/src/definitions/modules/sidebar.less
@@ -112,6 +112,7 @@ body.pushable {
 /* Page Context */
 .pushable:not(body) {
   transform: translate3d(0, 0, 0);
+  overflow-y: hidden;
 }
 .pushable:not(body) > .ui.sidebar,
 .pushable:not(body) > .fixed,


src/definitions/modules/sidebar.less
```diff
@@ -112,6 +112,7 @@ body.pushable {
 /* Page Context */
 .pushable:not(body) {
   transform: translate3d(0, 0, 0);
+  overflow-y: hidden;
```
exoego, line 115: I am afraid that vertical scroll bar may not appear on page context, when the page context is very long.
(Perhaps my overthinking ?)

If the height of the preview areas is shrinked on https://jsfiddle.net/lubber/vd2ecqgL/9/,  its content can not be fully-scrollable by vertical scroll bar. 
image

lubber-de: The /9/ link is the broken one ;)
But yes, i think this is intented to use the body scrollbar whenever the content of the pusher (by design the content needs to be inside the pusher, not as direct child of the context
https://jsfiddle.net/lubber/vd2ecqgL/81/

I think, your example screenshot is due to the jsfiddle console toolbar (i always disable that, because its more "real world" to use the native console

The shrinked example is this one and this is how user should use own overflow areas anyway
https://jsfiddle.net/lubber/vd2ecqgL/77/

If i am still not getting your point, please adjust one of my fiddles accordingly 🙂

exoego: > due to the jsfiddle console toolbar

Ah, I see.
After I disabled the console, vertical scrollbar seems fine.

PR #1626, opened 2020-07-31
feat(form): optional support all field errors in inline prompt

### LINKED ISSUE #1374: [Form Validation] `add prompt` method doesn't set error message
The form validation API for 'add prompt' doesn't work as expected. The [documentation](https://fomantic-ui.com/behaviors/form.html#/settings) looks ambiguous which describes two different explanation for 'add prompt' behavior.

## Testcase
Here's the fiddle (https://jsfiddle.net/k1a9w73t/4/). As the fiddle result, the validation methods return the default prompt message regardless of the message which specifies via `form('add prompt')` method. The fiddle uses version 2.7.5.

I tried to test with the latest version. However, the form validation doesn't work at all as the same methods with version 2.7.5. Here's the fiddle that uses version 2.8.4 (https://jsfiddle.net/zfgs590x/).

lubber-de:
## Description
According to the (unclear) docs there are two possibilities to show errors in an inline prompt label. You can supply either a string or an array of strings.

However, this was always ignored. It always uses only the very first field error, regardless how many other errors may exists.

The error message instead (when not using inline errors) is always showing all possible rule errors by default.
This PR now also shows all errors of a field in an inline prompt label just like the error message already does.

That said, the existing `prompt` template now, more logical, is used to created the label **content** (just like the `error` template does for non inline errors), rather than the label node itself (this is now done automatically).

Even if this is a breaking change, i doubt the `prompt` label was ever customised by somebody , because of the overriding logic (which made the prompt template useless.)

The previous behavior totally ignored the prompt templates inner content, because right after that, the html content inside the label was immediatly overridden all the time. (see comments below)

## Testcase
https://jsfiddle.net/lubber/h3zajpfc/17/

## Screenshot 
Using the above testcase, the form field has 2 rules which are by default not fullfilled

`inline: true` now always shows all invalid rules instead of only the first one, but if there still is only one error is does not show a list bullet making the visual behavior stay the same as before.
image

If there are more errors a bullet list will be shown just like the message (non-inline) already did.
image

## Closes
#1374


lubber-de:
> If someone has overridden the `.prompt: function(errors, labelClasses)` this could potentially break.

No, because the content of the created div was always overridden by `.html()` outside of the template immediatly, making a custom template nearly useless, because only the classname.label is set to the node.

```javascript
              if(!promptExists) {
                $prompt = settings.templates.prompt(errors, className.label);
                $prompt
                  .appendTo($fieldGroup)
                ;
              }
//Thats how it was before: No matter what the previous template creates, the content is overridden immediatly
              $prompt
                .html(errors[0])
              ;
```

As i just removed the additional `.html()` inside the prompt-template it doesn't make any difference. It's just important to the template that it creates the node.
Tbh, that was my exact thought why i created another `promptError()` template to keep everything backward compatible.

> Additionally the parameter `errors` could be removed.

Same here: i left it for backward compatibility (if we remove this, then the parameter order is different making this a definite breaking change.

So, if we want to consider this a breaking change at all, then i would rather rewrite this whole logic even more
- reuse the `error` template also for prompts instead of a separate promptError 
- **OR** let the prompt-template only create the content just like the `errrors` template does. More logical to me.
- create the raw div node of the prompt label outside of the template (only the className.label is added anyway

@fomantic/maintainers Opinion ?

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/223

lubber-de:
@exoego @ko2in @prudho Time to review this? You already approved the related docs PR...:wink:

exoego:
As an end-user, I prefer all errors are shown so one can fix errors at once. So I prefer `inline: true` behave like `inline : "all"`.
It maybe breaking changes, but it is better behavior, I guess.
Since FUI allows breaking changes in 2.x, I suggest to include the braking change into next major version (2.9 or later).

lubber-de:
Ok, i postpone this to 2.9 and do some rewrites to only use the existing prompt-template as suggested by me [earlier](https://github.com/fomantic/Fomantic-UI/pull/1626#issuecomment-667943717)

lubber-de:
I changed the code accordingly:

- `promptError` templates removed again
- `prompt` template is now used to create the label content just like the `error` template does (more logical)
- the prompt label node is created by fui itself but the user can still provide the custom label class
- tagged as breaking change and milestoned to 2.9
- issue description adjusted
- changed testcase

@ColinFrick  @exoego Please review again 🙂

PR #1629, opened 2020-08-03
[Dropdown] Use configured 'fields.values' key instead of fixed 'values'

WardLootens:
## Description
When calling `templates.menu()`, the object passed always has a single fixed key `values`.
However, the implementation of `templates.menu()` reads the object with the configurable key `fields.values`.
So if a key other then `values` is configured, this will fail.

```
$("#selectBox").dropdown({
	fields: {
		values: 'valuesX'
	}
});
```

## Testcase
See [JSFiddle](https://jsfiddle.net/wardlootens/04war7mx/3/)

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -795,7 +795,7 @@ $.fn.dropdown = function(parameters) {
                 }
                 module.remove.message();
                 module.setup.menu({
-                  values: values
+                  [fields.values]: values
                 });
 
                 if(values.length===0 && !settings.allowAdditions) {
@@ -992,7 +992,7 @@ $.fn.dropdown = function(parameters) {
               module.clear();
             }
             module.debug('Creating dropdown with specified values', values);
-            module.setup.menu({values: values});
+            module.setup.menu({[fields.values]: values});
             $.each(values, function(index, item) {
               if(item.selected == true) {
                 module.debug('Setting initial selection to', item[fields.value]);
@@ -1958,7 +1958,7 @@ $.fn.dropdown = function(parameters) {
               select = {},
               oldGroup = []
             ;
-            select.values = [];
+            select[fields.values] = [];
             $module
               .find('option')
                 .each(function() {
@@ -1979,14 +1979,14 @@ $.fn.dropdown = function(parameters) {
                   }
                   else {
                     if(group.length !== oldGroup.length || group[0] !== oldGroup[0]) {
-                      select.values.push({
+                      select[fields.values].push({
                         type: 'header',
                         divider: settings.headerDivider,
                         name: group.attr('label') || ''
                       });
                       oldGroup = group;
                     }
-                    select.values.push({
+                    select[fields.values].push({
                       name     : name,
                       value    : value,
                       text     : text,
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -992,7 +992,7 @@ $.fn.dropdown = function(parameters) {
               module.clear();
             }
             module.debug('Creating dropdown with specified values', values);
-            module.setup.menu({values: values});
+            module.setup.menu({[[fields.values]]: values});
```
lubber-de, line 995: Why double `[[` `]]`? Please reduce to only one

WardLootens: Damn, a stupid mistake.
I reverted the commit in the end, because it did not fix all cases.
Added another commit that should fix all call paths.

src/definitions/modules/dropdown.js
```diff
@@ -412,7 +412,9 @@ $.fn.dropdown = function(parameters) {
             module.refresh();
           },
           menu: function(values) {
-            $menu.html( templates.menu(values, fields,settings.preserveHTML,settings.className));
+            const response = values;
+            response[fields.values] = values.values;
```
lubber-de, line 416: Why the separate constant?

What about doing this instead: 
```javascript
if(!values[fields.values] && values.values) values[fields.values] = values.values;
```


TBH, the previous solution was more logical to me (yes, i understand you would have to adjust `get.selectValues` as well at many places)

WardLootens: Ok, I can see the initial solution might have been better.
Added commits to build up on that initial solution.
Also added commit to extract local variable in `get.selectValues`, but this is probably a matter of taste

lubber-de: Thanks, much better approach!

ColinFrick:
Does this work with IE11+? I haven't had the time to check it out in depth, but caniuse says no to computed property names for IE https://caniuse.com/#feat=mdn-javascript_functions_get_computed_property_names

src/definitions/modules/dropdown.js
```diff
@@ -795,7 +795,7 @@ $.fn.dropdown = function(parameters) {
                 }
                 module.remove.message();
                 module.setup.menu({
-                  values: values
+                  [fields.values]: values
```
lubber-de, line 798: As @ColinFrick pointed out, this won't work in IE11 (which we unfortunately have to support (at least it should not break existing code))
You can fix it by using a helper object like the `select` you used below

```javascript
var menuConfig = {};
menuConfig[fields.values] = values;
module.setup.menu(menuConfig);
```

src/definitions/modules/dropdown.js
```diff
@@ -992,7 +992,7 @@ $.fn.dropdown = function(parameters) {
               module.clear();
             }
             module.debug('Creating dropdown with specified values', values);
-            module.setup.menu({values: values});
+            module.setup.menu({[fields.values]: values});
```
lubber-de, line 995: Same here
```javascript
var menuConfig = {};
menuConfig[fields.values] = values;
module.setup.menu(menuConfig);
```

lubber-de:
@WardLootens Do you have time to implement the required little changes mentioned above?

WardLootens:
> @WardLootens Do you have time to implement the required little changes mentioned above?

I was travelling the past week so I did not have time, but I'll do it today or the coming days.

lubber-de:
@all-contributors please add @WardLootens for code

PR #1631, opened 2020-08-03
feat(slider): prevent line break in slider label and support bottom aligned label

### LINKED ISSUE #1625: Diffrent styling for the slider module
# Feature Request
I am using the slider module and it works perfectly and I wonder if there is any way to change the position of the labels being displayed on the slider.

For example, I want something like this:
request

And we have this:
given

I have tried some modifications at one of my demo servers and here is the [link](https://demo.woochoiceplugin.com/hify-store/product-category/eo_diamond_shape_cat/?EO_WBC=1&BEGIN=eo_diamond_shape_cat&STEP=1&FIRST=&SECOND=) to the page.

having

On the page, I have a filter slider named `SYMM.` which has a `Very Good` label, and unfortunately, my CSS skill is not so good that's why it been overlapped by the slider nodes.

So to avoid my confusing CSS codes I like to ask if there is any way to have the slider as I want?

ko2in:
## Description
This PR encourages the slider label to appear in one line without breaking the word into new line especially when the last label contains more than one word and break into multiple line which make the label overlaps the slider track in some situations.

This PR also supports to change the slider label position using `bottom aligned` flag in CSS class name which makes the label to be appeared below the slider track.

## Testcase
**Before:** https://jsfiddle.net/qr1hxs3L/

**After:** https://jsfiddle.net/qr1hxs3L/1/

## Screenshots
| Label Before | Label After  |
|---|---|
| Broken-Slider-Label | Fix-Slider-Label |

#### Top and bottom aligned labels ####
Sliders

## Closes
#1625

### DIFF
--- a/src/definitions/modules/slider.less
+++ b/src/definitions/modules/slider.less
@@ -195,8 +195,14 @@
 
   .ui.labeled.slider > .labels .label {
     display: inline-flex;
+    padding: @labelPadding;
     position: absolute;
     transform: translate(-50%, -100%);
+    white-space: nowrap;
+  }
+
+  .ui.bottom.aligned.labeled.slider > .labels .label {
+    transform: translate(-50%, 100%);
   }
   & when (@variationSliderTicked) {
     .ui.labeled.ticked.slider > .labels .label:after {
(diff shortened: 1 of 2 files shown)


lubber-de:
@ko2in Do you also have time for a `bottom aligned` example PR for the docs repo? 😉

ko2in:
Sure. Will send PR tomorrow.

noinkling:
This feature breaks when combined with other sizes. For example on a big slider there won't be enough spacing between the slider and the labels, and on a small slider there will be too much. You can fix it by adding `bottom: 0;` to the `.ui.bottom.aligned.labeled.slider > .labels .label` block.

ko2in:
@noinkling Please open a separate issue and provide your fiddle which causes the bug by this PR.

PR #1637, opened 2020-08-11
feat(icon): add the old two classes for linegraph and chartline for c…

lubber-de:
## Description
This PR re-adds the two word icon names for `chartline` and `linegraph` as of #1636  to avoid being a breaking change as @prudho mentioned [here](https://github.com/fomantic/create-fomantic-icons/pull/115#issuecomment-671762616) .

The next full font awesome update will happen in 2.9.0 using the fixes from https://github.com/fomantic/create-fomantic-icons/pull/115 Then two word aliases will be completely removed


hammy2899:
This is also the PR which is deprecating the old icon class names

PR #1639, opened 2020-08-11
[Build] Fix issue when "individuals" or "components" has only single element 

### LINKED ISSUE #1638: [Build] Long "components" with "individuals" break build
# Bug Report

## Steps to reproduce

I prepared a repo for repro. 

1. `git clone https://github.com/exoego/fomantic-ui-error-repro.git`
1. `cd fomantic-ui-error-repro/semantic`
3. `gulp build`

## Expected result
Build should run successfully (build and bundle all components, except ones in `"individuals"`.
https://fomantic-ui.com/introduction/build-tools.html#excluding-components-from-main-cssjs-file

## Actual result
If `individials` array has 1 or more components, and `components` array has many, build fails with "ENAMETOOLONG" error like below:

```
[21:23:26] Using gulpfile ~/IdeaProjects/fui-test/semantic/gulpfile.js
[21:23:26] Starting 'build'...
Building Semantic
[21:23:26] Starting 'build-css'...
[21:23:26] Starting 'Building uncompressed CSS'...
[21:23:26] Starting 'Building compressed CSS'...
[21:23:26] 'Building uncompressed CSS' errored after 22 ms
[21:23:26] Error: ENAMETOOLONG: name too long, lstat '/home/vagrant/IdeaProjects/fui-test/semantic/src/definitions/{reset,site,button,container,divider,flag,header,icon,image,input,label,list,loader,rail,reveal,segment,step,breadcrumb,form,grid,menu,message,table,card,comment,feed,item,statistic,accordion,calendar,checkbox,dimmer,dropdown,modal,placeholder,popup,progress,slider,rating,search,shape,sidebar,sticky,tab,text,toast,transition,state,visibility,emoji.less'
[21:23:26] 'build-css' errored after 23 ms
[21:23:26] 'build' errored after 24 ms
```

## Note

If `components` array is relatively small, build runs successfully.
However, build should run regardless of the number of elements in `components`.

## Version

2.8.6

exoego:
## Description

It seems that the current logic (below) to create file blob like `{foo,bar,buz}.less` assumes that `config.globs.individuals` is always like `{a,b,c}`, but it does not when `individuals` has only one single element.
https://github.com/fomantic/Fomantic-UI/blob/74458216193cbf22653784a98f58ce57b0fc0244/tasks/build/css.js#L131-L134

After This PR, `config.globs.components` and `config.globs.individuals` are `{a,b,c}` if those arrays has 1 or more elements, so `build/css.js` and `build/javascript.js` works regardless of the number of elements in `individuals`/`components`

## Testcase

I have no idea how to test build tasks.... 
Suggestion is highly appreciated 🙏 

## Closes
#1638

### DIFF
--- a/tasks/config/project/config.js
+++ b/tasks/config/project/config.js
@@ -128,18 +128,14 @@ module.exports = {
     }
 
     // takes component object and creates file glob matching selected components
-    config.globs.components = (typeof config.components == 'object')
-      ? (config.components.length > 1)
-        ? '{' + config.components.join(',') + '}'
-        : config.components[0]
+    config.globs.components = (Array.isArray(config.components) && config.components.length >= 1)
+      ? '{' + config.components.join(',') + '}'
       : '{' + defaults.components.join(',') + '}'
     ;
 
     // components that should be built, but excluded from main .css/.js files
-    config.globs.individuals = (typeof config.individuals == 'object')
-      ? (config.individuals.length > 1)
-        ? '{' + config.individuals.join(',') + '}'
-        : config.individuals[0]
+    config.globs.individuals = (Array.isArray(config.individuals) && config.individuals.length >= 1)
+      ? '{' + config.individuals.join(',') + '}'
       : undefined
     ;
 


tasks/config/project/config.js
```diff
@@ -128,18 +128,14 @@ module.exports = {
     }
 
     // takes component object and creates file glob matching selected components
-    config.globs.components = (typeof config.components == 'object')
-      ? (config.components.length > 1)
-        ? '{' + config.components.join(',') + '}'
-        : config.components[0]
+    config.globs.components = (Array.isArray(config.components) && config.components.length >= 1)
+      ? '{' + config.components.join(',') + '}'
```
exoego, line 132: Note) One idea came up on my mind when working on this.
If we exclude the names, defined in `individuals`, from `components` here, I think users can define only `individuals` and cn omit `components`.
This makes it easier to use `individuals` when optimizing bundling.

Currently, one must define `components` (see below doc), but I think this is not great when new component is added.

> https://fomantic-ui.com/introduction/build-tools.html#excluding-components-from-main-cssjs-file
> Note: After defining the individuals in semantic.json, you need to remove them from the main components list. This also means that you can't leave the components list empty anymore (which defaults to building all components).

If this is OK, I will open a separate PR for this enhancement.

ko2in: Yes. That would be more friendly. If we've `individuals`, then everything except from the `individuals` should be bundled without needing to describe with `components`.

lubber-de: Remember, the `components` settings defines all the wanted components inside the semantic.json. If we automatically create that list by substracting the `individuals` list form all available components, you are going to force people to always have every other component inside the semantic.css/.js.

### Example
- one wants to only have dimmer, modal, button, site and reset as `components´
- one also wants to have toast as individual

#### Current logic
- semantic.js/.css will contain the components dimmer, modal, button, side, reset
- toast.js/.css is generated separately

#### Your proposed logic
- semantic.js/.css contains **everything but** toast.js/.css
- toast.js/.css is generated separately

So most probably such change is only save/backward compatible, if no `components` value is given.

ko2in: That's right. To having other components in the semantic.css/js, the user would need to add all components into `individuals` except those he want to bundle. But it'll also end up with the unnecessary components get generated by build process.

exoego: Oh, I do not mean removing `components` completely.
In the below matrix, No. 4 is my proposal.

\No. |components | individuals | Current with this bugfix | Proposal
---|---|---|---|--- 
1|`dimmer`, `button` | undefined/empty | `semantics.js/.css`: only `dimmer` and `button` | same as left
2|`dimmer`, `button` | `toast`| `semantics.js/.css`: only `dimmer` and `button`<br>`toast.js/.css` is generated separately | same as left
3|undefined/empty | undefined/empty | `semantics.js/.css`: all | same as left
4|undefined/empty | `toast`| `semantics.js/.css`: all including `toast` 😞<br>`toast.js/.css` is generated separately | `semantics.js/.css`: all minus `toast` 😄 <br>`toast.js/.css` is generated separately

lubber-de: Alright, that's what i also meant by 
> So most probably such change is only save/backward compatible, if no components value is given.

I'll merge this PR here now, so, you can go ahead and prepare another one for the proposed solution above ;)

PR #1647, opened 2020-08-18
[Calendar] Support safari to accept yyyy-mm-dd format using dashes

lubber-de:
## Description
After implementing #1462, a pure given date in `yyyy-mm-dd` format did not work in safari  anymore , because safari does not handle the internal converted date of `yyyy-mm-dd 00:00:00`

This PR now changes the dashes to slashes, so the internal result is  `yyyy/mm/dd 00:00:00`, which is supported in every browser

## Testcase
- Use Safari

### Broken
The given date `2020-08-12` results in `August, 12, 2000`
https://jsfiddle.net/lubber/s82zoc5w/

### Fixed
The given date `2020-08-12` correctly results in `August, 12, 2020` as in every other browser already
https://jsfiddle.net/lubber/s82zoc5w/2/

#### Broken
image

#### Fixed
image

## Closes
https://github.com/fomantic/Fomantic-UI/issues/1462#issuecomment-672559777

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1468,7 +1468,7 @@ $.fn.calendar.settings = {
         return null;
       }
       if(text.match(/^[0-9]{4}[\/\-\.][0-9]{2}[\/\-\.][0-9]{2}$/)){
-        text += ' 00:00:00';
+        text = text.replace(/[\/\-\.]/g,'/') + ' 00:00:00';
       }
       // Reverse date and month in some cases
       text = settings.monthFirst || !text.match(/^[0-9]{2}[\/\-\.]/) ? text : text.replace(/[\/\-\.]/g,'/').replace(/([0-9]+)\/([0-9]+)/,'$2/$1');


exoego (approved):
LGTM

ko2in (commented):
I know that it's only change of regular expression in JS, but the calendar UI looks weird to me.
Calendar

lubber-de:
What looks weird? I don't nee it 🤔

lubber-de:
@ko2in Ah...the class for the calendar was missing... fixed jsfiddle here 
https://jsfiddle.net/lubber/s82zoc5w/2/

ko2in:
> What looks weird? I don't see it

There are spaces around the calendar. Looks like it doesn't happen to you.

lubber-de:
> There are spaces around the calendar. Looks like it doesn't happen to you.

It does 😆 ... see comment above, jsfiddle fixed here https://jsfiddle.net/lubber/s82zoc5w/2/

ko2in (approved):
LGTM.

PR #1651, opened 2020-08-22
[Dropdown] optional show a scrollhint on ios devices (or firefox mobile) to indicate more entries

lubber-de:
## Description
IOS devices do not show a scrollbar when the content is not scrolled. This is a design decision by apple and cannot be adjusted via CSS (for example by  `-webkit-scrollbar`)
A similiar behavior is built into firefox mobile (also on Android)

That said if one opens a dropdown it is not clear, if the shown entries are the only available options or if the user has to scroll down to see more, which is confusing and probably leads to wrong selections (because the user did not know of more choices, because of avoided scrolling)

This PR provides the optional `scrollhint` class to be used for the `menu` to show a short animation hint on the right edge (where the usual scrollbar appears) when the dropdown opens to indicate swiping to see more entries

## Testcase
Use an iOS Device or Firefox mobile (also on Android)

### Normal
https://raw.githack.com/lubber-de/fomantic-ui/jsfiddlefullscreen/jsfiddle/#!lubber/2mcak7xr/129/

### using `scrollhint menu`
https://raw.githack.com/lubber-de/fomantic-ui/jsfiddlefullscreen/jsfiddle/#!lubber/2mcak7xr/190/

jsfiddle: https://jsfiddle.net/lubber/2mcak7xr/190/

## Screenshot
|Normal|using `scrollhint menu`|
|-|-|
|iosnoscrollhint|iosscrollhint|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5197

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -608,6 +608,40 @@ select.ui.dropdown {
     border: none !important;
     box-shadow: none !important;
   }
+  & when (@variationDropdownScrollhint) {
+    @supports (-webkit-overflow-scrolling: touch) {
+      /* CSS specific to iOS devices only  */
+      .ui.selection.dropdown .scrollhint.menu:not(.hidden):before {
+        content: '';
+        z-index: @scrollhintZIndex;
+        display: block;
+        position: absolute;
+        opacity: 0;
+        right: 0;
+        top: 0;
+        height: 100%;
+        border-right: @scrollhintRightBorder;
+        border-left: @scrollhintLeftBorder;
+        animation: scrollhint @scrollhintDuration @scrollhintEasing @scrollhintIteration;
+        border-image: linear-gradient(to bottom, @scrollhintStartColor, @scrollhintEndColor) 1 100%;
+      }
+
+      .ui.inverted.selection.dropdown .scrollhint.menu:not(.hidden):before {
+        border-image: linear-gradient(to bottom, @invertedScrollhintStartColor, @invertedScrollhintEndColor) 1 100%;
+      }
+
+      @keyframes scrollhint {
+        0% {
+          opacity: 1;
+          top: 100%;
+        }
+        100% {
+          opacity: 0;
+          top: 0;
+        }
+      }
+    }
+  }
 }
 
 & when (@variationDropdownSearch) {
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -609,36 +609,45 @@ select.ui.dropdown {
     box-shadow: none !important;
   }
   & when (@variationDropdownScrollhint) {
-    @supports (-webkit-overflow-scrolling: touch) {
-      /* CSS specific to iOS devices only  */
+    /* CSS specific to iOS devices only  */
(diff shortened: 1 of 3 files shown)


exoego (approved):
LGTM

prudho (approved):
Nice ! LGTM 👍

ko2in (commented):
It doesn't work on my FF 79.0.5 on Android yet. According to [MDN](https://developer.mozilla.org/en-US/docs/Archive/Web/CSS/@media/-moz-touch-enabled), `-moz-touch-enabled` is no longer supported since FF 73.

And it also doesn't work on Safari on my iPad (iOS 10.3.3). According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout), it's fully supported on Safari on iOS. My iPad is 4th generation and a bit old.

lubber-de:
@ko2in 🤔 
The last official firefox app in google playstore is 68.11.0 (for me..?), which works (on my android device)
I now installed Firefox beta (which is 80.x) and can confirm, it does not work there anymore 😢 
Any idea what else we can use instead of `-moz-touch-enabled` to only target ff mobile via pure CSS?

Regarding iOS, I tried ios devices via browserstack, but can only access ipad 5th with ios 11, were it works as well
Even iphone 7 using ios 10 was working as well.
I prepared a jsfiddle to also add the previous approach via `-webkit-overflow-scrolling: touch`
Could you please check if this is working on your ipad?

https://jsfiddle.net/lubber/2mcak7xr/169/
https://raw.githack.com/lubber-de/fomantic-ui/jsfiddlefullscreen/jsfiddle/#!lubber/2mcak7xr/169/

lubber-de:
I found a solution 😄
It works in Firefox nightly on android :) (It doesnt firefox beta, but this seems to be a general issue, because i even tried to not use an media query but a force usage of the animation and even this is not working in firefox beta

https://jsfiddle.net/lubber/2mcak7xr/190/
https://raw.githack.com/lubber-de/fomantic-ui/jsfiddlefullscreen/jsfiddle/#!lubber/2mcak7xr/190/

```css
@supports (-webkit-touch-callout: none) or (-webkit-overflow-scrolling: touch) or (-moz-appearance:none) {
  @media (-moz-touch-enabled), (pointer: coarse) {
    .ui.dropdown .scrollhint.menu:not(.hidden):before {
      animation: scrollhint 2s ease 2;
    }
  }
}
```

lubber-de:
@ko2in @prudho @exoego 
I change the code to support latest firefox and centralised logic. Please review again

ko2in (approved):
LGTM :+1:

lubber-de:
@ko2in As i now only have one unique selector to match ios and firefox mobile, I merged the two separate, but equal, selectors into one now, so the whole CSS settings apply to those devices only instead of only the animation trigger
Please review again

ko2in (approved):
Also fine :+1:

lubber-de:
@exoego @prudho Please review again if you find time.

ColinFrick (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/225

PR #1656, opened 2020-08-25
[Dropdown] Multiple search button dropdown misaligned, placeholder wraps and search text ghosting

### LINKED ISSUE #1623: [dropdown] when multiple, placeholder text is wrapped
# Bug Report
When using a dropdown with multiple selection, placeholder text with spaces is wrapped. It's not when removing the multiple part.

## Steps to reproduce

cf. jsfiddle

## Expected result
text should not be wrapped in multiple like the no multiple dropdown

## Actual result
text is wrapped

## Testcase
This is the test case for latest version of fomantic (2.8.6) https://jsfiddle.net/jLn0wkv9/1/
Also we notice some kind of change between the 2.8.4 and 2.8.5 of this kind of input (but still with the wrap text issue) here the jsfiddle with 2.8.4 https://jsfiddle.net/1yxtrqen/1/

## Screenshot (if possible)
2.8.6 (multiple is on right)
image
2.8.4 (multiple is on right)
image

## Version

2.8.6

## Notes
For the change between 2.8.4 and 2.8.6 its a css rule that override the padding of multiple dropdown
2.8.4:
image

2.8.6:
image

lubber-de:
## Description
This PR fixes

- multiple search button dropdown had wrong padding
- Placeholders in multiple search dropdowns were wrapping
- Search in dropdown buttons had ghosting text

## Broken
https://jsfiddle.net/lubber/47r2v1cf/59

## Fixed
https://jsfiddle.net/lubber/47r2v1cf/60

### Broken
multipleghost

## Fixed
multiplenoghost

## Closes
#1623

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -681,7 +681,7 @@ select.ui.dropdown {
     color: @selectionTextUnderlayColor !important;
   }
 
-  .ui.search.dropdown.button. > span.sizer {
+  .ui.search.dropdown.button > span.sizer {
     display: none;
   }
 
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -681,6 +681,9 @@ select.ui.dropdown {
     color: @selectionTextUnderlayColor !important;
   }
 
+  .ui.search.dropdown.button. > span.sizer {
+    display: none;
+  }
 
   /* Search Menu */
   .ui.search.dropdown .menu {
@@ -742,7 +745,7 @@ select.ui.dropdown {
   ---------------*/
 
   /* Multiple Selection */
-  .ui.multiple.dropdown {
+  .ui.ui.multiple.dropdown {
     padding: @multipleSelectionPadding;
   }
   .ui.multiple.dropdown .menu {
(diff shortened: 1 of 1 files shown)


ko2in (commented):
The default value for `min-width` of `.ui.search.dropdown` is empty which is invalid.
Dropdown

https://github.com/fomantic/Fomantic-UI/blob/df26dde8e64674a497f6ac61e5962d9f4a1a5680/src/themes/default/modules/dropdown.variables#L222

I think we should use the same min-width value as `@selectionMinWidth`.
https://github.com/fomantic/Fomantic-UI/blob/df26dde8e64674a497f6ac61e5962d9f4a1a5680/src/themes/default/modules/dropdown.variables#L131

Or I suggest `min-content` as default value. What do you think?

ko2in:
As far as I tested, all produce the awkward results apart from the unit value.
### min-width: auto | fit-content | initial | max-content | min-content | revert | unset
min-width-1

### min-width: inherit
min-width-2

### min-width: moz-available | -webkit-fill-available
min-width-3

I guess it should use the same unit as `@selectionMinWidth` then.

lubber-de:
@ko2in No, we should not change this (and it's not even part of this PR 😉 )

If you set a fixed value for `.ui.search.dropdown` all code which is using this inline will break because it's getting much wider than probably desired

See the last dropdown in this jsfiddle https://jsfiddle.net/lubber/47r2v1cf/67/
### Current
image

### Your proposed change
image

lubber-de:
@ko2in And `min-content` will also reduce the existing selection dropdown width
image

lubber-de:
Lets discuss / maybe adjust this in a separate PR

ko2in (approved):
LGTM. I've already created an [issue](https://github.com/fomantic/Fomantic-UI/issues/1657) for the invalid CSS syntax, where we can discuss for that.

prudho (approved):
LGTM

PR #1661, opened 2020-08-28
[Form] Dont bind events to dropdown search inputs to prevent jquery 3.4 callstack error

### LINKED ISSUE #1313: [dropdown]  jQuery 3.4.1  Maximum call stack size exceeded when calling '.dropdown('  before '.form('
https://jsfiddle.net/hw2dyLur/ 
It does work with the previous jQuery 3.3.1 . Is Fomantic not compatible yet with 3.4.1?

jquery.min.js?v=1581085047540:2 Uncaught RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at V (jquery.min.js?v=1581085047540:2)
    at Y.get (jquery.min.js?v=1581085047540:2)
    at Object.trigger (jquery.min.js?v=1581085047540:2)
    at k.fn.init.triggerHandler (jquery.min.js?v=1581085047540:2)

Seems like the  blur event on a form field triggers a blur handler that triggers the blur again so it spins like this until it reaches the call stack limit.

lubber-de:
## Description
The field selector in forms also fetches dropdowns search fields which are not supposed to be form inputs.
Each fetched fields  is bound to events. especially the blur event.

This leads into the situation that jquery 3.4 runs into a "maximum callstack exceeded" bug, because the blur recursively triggers another blur for the same field again when the dropdown was initialized before the form was initialized. Its not happening when the dropdown is initialized after the form was initialized because the `search dropdown` possible hasn't created  the internally used search field. The event system handling seems to changed slightly between jquery 3.3.1 and 3.4, because it was not happening the error is not happening with older jqery versions. However the field selector was ever since fetching too many fields.

This PR now makes sure unnecessary input fields are not fetched by the form which removes the error.

## Testcase
- Select something from the dropdown field 
- Watch console

### Broken
"Maxium call stack exceeded" 😠 
https://jsfiddle.net/lubber/0avf64sj/9/

### Fixed
Nothing 🥰 
https://jsfiddle.net/lubber/0avf64sj/11/

## Closes
#1313

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1571,7 +1571,7 @@ $.fn.form.settings = {
   selector : {
     checkbox   : 'input[type="checkbox"], input[type="radio"]',
     clear      : '.clear',
-    field      : 'input, textarea, select',
+    field      : 'input:not(.search):not([type="hidden"]), textarea, select',
     group      : '.field',
     input      : 'input',
     message    : '.error.message',
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1571,7 +1571,7 @@ $.fn.form.settings = {
   selector : {
     checkbox   : 'input[type="checkbox"], input[type="radio"]',
     clear      : '.clear',
-    field      : 'input:not(.search):not([type="hidden"]), textarea, select',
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1664, opened 2020-09-03
[Dropdown] "very short" menu height always overriden by normal "short" height because of  wrong specificity order 

lubber-de:
## Description

The `very short`  size variation did not work. It was always overridden by the `short` variation because both share same specificity, thus the declaration order matters.

## Testcase
I used a new "test" and "short test" naming to show the difference 

### Broken
Wrong order: first `very test`, then  `test`
https://jsfiddle.net/lubber/nq4c5eg1/13

### Fixed
Correct order: first `test`, then  `very test`
https://jsfiddle.net/lubber/nq4c5eg1/12

### Broken
image

### Fixed
image

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -448,11 +448,11 @@ select.ui.dropdown {
 
   @media only screen and (max-width : @largestMobileScreen) {
     & when (@variationDropdownShort) {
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1669, opened 2020-09-09
feat(dropdown): Dropdown formatted with label like input labeled

### LINKED ISSUE #1667: Dropdown right label
Hi, i would like to do Dropdown right label like input right label it is possible? or i should manually add styles as in picture ??
Screenshot 2020-09-09 at 11 40 45
Screenshot 2020-09-09 at 11 57 57

ko2in:
## Description
Sometimes, we will need to create a stack of dropdown and the label together as described in this [issue](https://github.com/fomantic/Fomantic-UI/issues/1667).

This PR intends to support the dropdown to be formatted with the label at left or right position same as input element stacks with the [label](https://fomantic-ui.com/elements/input.html#labeled) and the [action](https://fomantic-ui.com/elements/input.html#action) button.

## Note
This implementation is not standalone and 100% semantic because, the class name `dropdown` is already used by dropdown module.

At the moment, it only works when the theme has enabled the labeled input variation `@variationInputLabeled` and the dropdown module needs to be the child of `labeled input` element. If we could find any suitable alias for this module, I'll update this PR and implement as standalone extension without depending on the input element.

## Test Case
https://jsfiddle.net/ko2in/ucsyf3nb/

## Screenshot
DropdownLabeled

## Closes
#1667

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1323,6 +1323,30 @@ select.ui.dropdown {
   }
 }
 
+& when (@variationInputLabeled) {
+  /*--------------------
+          Labeled
+  ---------------------*/
+
+  /* Regular Label on Left */
+  .ui.labeled.input:not([class*="corner labeled"]) .label:first-child + .dropdown {
+    border-top-left-radius: 0;
+    border-bottom-left-radius: 0;
+    border-left-color: transparent;
+  }
+
+  /* Regular Label on Right */
+  .ui[class*="right labeled"].input > .dropdown:not(:last-child) {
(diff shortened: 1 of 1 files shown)


lubber-de (approved):
LGTM 👍

exoego (approved):
Very nice feature:+1:

ColinFrick (approved):

lubber-de:
@ko2in Do you have time to prepare an example PR for the docs repo? 🙂

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/282

PR #1672, opened 2020-09-11
[Dropdown] Allow displayType transition setting to be overriden by dropdown

lubber-de:
## Description
In situations where a dropdown was inside a right menu (which got a flex setting) the transition module was taking that flex displaytype setting for the dropdown menu. This unfortunately results into a horizontal menu list.
To be able to adjust this, i added the `displayType`, used for the transition, setting to the dropdown module as well. If set, this setting could be used just as in transition to correct the display if needed.

Additionally this setting also had to be provided to the hiding transition, otherwise the wrong flex setting would  be reapplied before hiding the dropdown again.

## Testcase
Click the top right menu button

### Broken
The menu is displayed horizontally
https://jsfiddle.net/lubber/wtvfd7pn/25/

### Fixed
The menu is displayed vertical
https://jsfiddle.net/lubber/wtvfd7pn/24/

### Broken
image

### Fixed
image

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3565,7 +3565,7 @@ $.fn.dropdown = function(parameters) {
               module.set.scrollPosition(module.get.selectedItem(), true);
             }
             if( module.is.hidden($currentMenu) || module.is.animating($currentMenu) ) {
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

exoego (approved):

PR #1673, opened 2020-09-12
[Label,Form] Apply the error color not only to the label tag but also with label class

ko2in:
## Description
When the form with error state contain the labels, these labels will be styled as red color, but only they're used as HTML label element.

Now, any element with `label` class will have the same style as label element, and produces more consistent design flow.

## Testcase
**Before:** https://jsfiddle.net/ko2in/sdca1hjt/2/

**After:** https://jsfiddle.net/ko2in/sdca1hjt/4/

## Screenshot
Before | After
----------|---------
LabelStyle-Before | LabelStyle-After

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -495,7 +495,9 @@
     }
 
     .ui.ui.form .fields.@{state} .field label,
+    .ui.ui.form .fields.@{state} .field .ui.label,
     .ui.ui.form .field.@{state} label,
+    .ui.ui.form .field.@{state} .ui.label,
     .ui.ui.form .fields.@{state} .field .input,
     .ui.ui.form .field.@{state} .input {
       color: @c;
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -468,6 +468,7 @@
     @c: @formStates[@@state][color];
     @bg: @formStates[@@state][background];
     @bdc: @formStates[@@state][borderColor];
+    @lbg: @formStates[@@state][labelBackground];
 
     /* On Form */
     .ui.form.@{state} .@{state}.message:not(:empty) {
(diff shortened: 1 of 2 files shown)


lubber-de:
I basically would approve this. I have a suggestion though:

1. If a label has got a color, then it looks ugly because the background color isn't changed and does perhaps not fit to the error text color
2. `action input` is not supported (wasn't before either)

Proposal
https://jsfiddle.net/lubber/q7fprg81/17/

```css
.ui.ui.form .field.error .labeled.button {
  color: #9f3a38;
}

.ui.ui.form .field.error .labeled.button:not(.basic),
.ui.ui.form .fields.error .labeled.input > label:not(.basic),
.ui.ui.form .fields.error .labeled.input > .ui.label:not(.basic),
.ui.ui.form .field.error .labeled.input > label:not(.basic),
.ui.ui.form .field.error .labeled.input > .ui.label:not(.basic) {
  background: #EACBCB;
}
```
## Before
image


## After
image

ko2in:
Thanks for your suggestion. I agree that we should add error background to look better with the text color. Wouldn't it better we add the error background no matter it is basic or has other background color? That would give more consistency which is my primary intention of this PR.

About the action input, I didn't intend to add the background because, I believe the action button shouldn't change the background color at all. Otherwise, it will lose the user's concentration and it doesn't represent the error state of the field, but the label does.

lubber-de:
Ok, agreed 👍 (no action input support and also have basic labels get a background color)
While testing how this would look then, i found out a `corner labeled input` should not get the text color set, so the label color selector should be adjusted to add `:not(.corner)`

See https://jsfiddle.net/lubber/q7fprg81/36/

image

ko2in:
@lubber-de I updated my PR. Please review again. I also found one minor issue that need to fix. But I should send as a separate PR.

lubber-de (approved):
LGTM

lubber-de:
@fomantic/maintainers Time to review this?

ColinFrick (approved):

PR #1681, opened 2020-09-20
fix(dropdown) - forceSelection trigger event with null addition

### LINKED ISSUE #1679: [Dropdown] JavaScript error when adding to dropdwon while hideAdditions = false
# Bug Report
when allowAdditions is true and  hideAdditions is  false , Javascript error occurs when clicking the new added label

## Steps to reproduce
1. create a new dropdown with options `{ allowAdditions: true, hideAdditions: false}`
2. add a new option
3. click the new added option
4. see the error 

## Expected result
shouldn't get an error

## Actual result
 Uncaught TypeError: Cannot read property 'trim' of undefined

## Testcase
http://jsfiddle.net/kamalkemo/bwhsrvdf/

## Version
2.8.7

TheJltres:
## Description
Hi!
This fix is very simple,  when `allowAdditions` is handled by click event, `forceSelection` triggers event click again, but with `allowAdditions` removed can't find the item.
As far as I know I can't know if the event calling this method is the original click or `forceSelection` trigger, so what I did is: if there's no element and `allowAdditions` and `forceSelection` is set to true return `null`.

## Testcase
- Add a non existing value to the dropdown
- Watch console

## Broken
A JS error occurs `Cannot read property 'trim' of undefined`
http://jsfiddle.net/kamalkemo/bwhsrvdf/

## Fixed
No error anymore
http://jsfiddle.net/lubber/sqpc9d1y/

## Closes
#1679

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1942,15 +1942,16 @@ $.fn.dropdown = function(parameters) {
                 $choice.find(selector.menu).remove();
                 $choice.find(selector.menuIcon).remove();
               }
-              else if(settings.allowAdditions && settings.forceSelection) {
-                return;
+              if($choice.data(metadata.text) !== undefined) {
+                return $choice.data(metadata.text);
               }
-              return ($choice.data(metadata.text) !== undefined)
-                ? $choice.data(metadata.text)
-                : (preserveHTML)
-                  ? $choice.html().trim()
-                  : $choice.text().trim()
-              ;
+              else {
+                if($choice.length == 0 && settings.allowAdditions && settings.forceSelection){
+                  return;
+                }
+
+                return (preserveHTML) ? $choice.html().trim() : $choice.text().trim();
+              }
             }
           },
           choiceValue: function($choice, choiceText) {
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1942,6 +1942,9 @@ $.fn.dropdown = function(parameters) {
                 $choice.find(selector.menu).remove();
                 $choice.find(selector.menuIcon).remove();
               }
+              else if(settings.allowAdditions && settings.forceSelection) {
+                return;
+              }
               return ($choice.data(metadata.text) !== undefined)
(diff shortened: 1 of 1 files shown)


kamalkemo:
This doesn't work as expected

value is removed when clicked

TheJltres:
Okey, sorry, wrong commit 🤦, that doesn't make sense.

TheJltres:
As you can see now it handles well the click and no error is trowed, thanks for your patience

TheJltres:
Now that it works I want to get a hint from maintainers, there is another way to prevent this, and is inside force selection, if 'allowAdditions' is enabled and there is no item, don't trigger click event.

I would like to discuss this for a better fix.

src/definitions/modules/dropdown.js
```diff
@@ -1942,12 +1942,16 @@ $.fn.dropdown = function(parameters) {
                 $choice.find(selector.menu).remove();
                 $choice.find(selector.menuIcon).remove();
               }
-              return ($choice.data(metadata.text) !== undefined)
```
ko2in, line 1945: Only two minor changes here would fix the same instead of 10 new lines.
```
return ($choice.data(metadata.text) !== undefined)
  ? $choice.data(metadata.text)
  : (preserveHTML)
    ? $choice.html() && $choice.html().trim()
    : $choice.text() && $choice.text().trim()
;
```

TheJltres: Okey, no problem!

ko2in (approved):
LGTM. :+1:

lubber-de (approved):
LGTM. I added a testcase jsfiddle to your PR description. Please try to create a testcase yourself whenever providing a PR. Makes it easier for maintainers to review and increases chances to get things merged quicker 😉

TheJltres:
> LGTM. I added a testcase jsfiddle to your PR description. Please try to create a testcase yourself whenever providing a PR. Makes it easier for maintainers to review and increases chances to get things merged quicker 😉

I didn't know how to, now that i saw your jsfiddle i will do it next time. Thanks!

PR #1683, opened 2020-09-24
[Dropdown] Add `preventChangeTrigger` parameter to the `set selected` behavior

prudho:
## Description
This PR simply add the ability to prevent the `onChange` event trigerring when the `set selected` behavior is used. Until now, the only way to not trigger the `onChange` was to use the `set value` behavior, which is not IMHO the right stuff to call when you want to programmatically select a value.


lubber-de (approved):
LGTM

prudho:
Doc added by fomantic/Fomantic-UI-Docs#233.

ko2in (approved):
LGTM.

PR #1686, opened 2020-09-24
[Slider] - With step 0, divides by 0

TheJltres:
## Description

When step is 0, this division returns `Infinity`, so later on function `gapRatio` while loop is infinite.
Now is checking if step is 0 to change for 1 else return step.

## Testcase
* Add an slider
* Resize window
* Window is freezed

## Broken
https://jsfiddle.net/mthrcf5n/3/

--> Simply move JSFiddle's grid divider or resize your browser window

## Fixed

It should have a normal behavior now
https://jsfiddle.net/TheJltres/p4zte923/3/

## Closes
#1684

Issue #1684, opened 2020-09-24
[Slider] browser freezes upon window resize with step-less slider


mhthies:
# Bug Report
Resizing the browser window with a step-less slider (initialized with `step: 0`) on the page, freezes the overall browser tab.

## Steps to reproduce
1. Create a simple page with the slider module loaded and a slider initialized without steps (`<div class="ui slider"></div>` and `$(...).slider({ step: 0 });`)
2. Open the page
3. Resize the browser window

## Expected result
Nothing happens.

## Actual result
Browser tab hangs up. I assume, it enters some endless JavaScript loop or JavaScript <-> layout recalculation loop.

Tested with Chromium 85.0.4183.102 and Firefox 80.0.1.

## Testcase
https://jsfiddle.net/mthrcf5n/3/

--> Simply move JSFiddle's grid divider or resize your browser window

## Version
2.8.7

TheJltres:
I think the bug is caused by dividing number/0 so it returns Infinity number, this hapens on `numLabels`.

https://github.com/fomantic/Fomantic-UI/blob/develop/src/definitions/modules/slider.js#L580

ko2in:
@TheJltres You're right. The `numLabels` returns infinity when step is disabled and has value '0', and causes endless loop on resize event.

TheJltres:
I'm preparing a PR, 2 minutes 😄

PR #1687, opened 2020-09-25
[Slider] No minimum step for label generator

### LINKED ISSUE #1685: [Slider] labeled step-less slider
# Feature Request

I'm using the Slider module to create controls for setting a value in a contuinous range (0-100%). Now, I'd like to add labels to these step-less sliders, forming a scale, to give the user more feeling for the right knob position.

The result could look just like a slider *with* steps, but still have the knob freely movable to any position between the labels to represent any real number, just like a slider with `step: 0`:

## Screenshot
(I used a `ui ticked labeled slider` for this example image, with a custom `interpretLabel` function for generating the percentage labels.)

labeled-stepless-slider

TheJltres:
## Description
Here was set a minimum step of 0, i think for the same reason of #1684, if you remove this step it handles `atoLabel` for all kind of steps.

**This fix needs #1686**

PD. Sorry @ko2in I could not do it with fewer line changes 🤣 

## Testcase

https://jsfiddle.net/TheJltres/z43vub60/1/

## Closes
#1685

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -589,7 +589,7 @@ $.fn.slider = function(parameters) {
 
             switch (settings.labelType) {
               case settings.labelTypes.number:
-                return Math.round(((value * module.get.step()) + module.get.min()) * precision ) / precision;
+                return Math.round(((value * (module.get.step() === 0 ? 1 : module.get.step())) + module.get.min()) * precision ) / precision;
               case settings.labelTypes.letter:
                 return alphabet[(value) % 26];
               default:
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -205,28 +205,26 @@ $.fn.slider = function(parameters) {
             });
           },
           autoLabel: function() {
-            if(module.get.step() != 0) {
-              $labels = $module.find('.labels');
-              if($labels.length != 0) {
-                $labels.empty();
-              }
-              else {
-                $labels = $module.append('<ul class="auto labels"></ul>').find('.labels');
-              }
-              for(var i = 0, len = module.get.numLabels(); i <= len; i++) {
-                var
-                  labelText = module.get.label(i),
-                  $label = (labelText !== "") 
-                    ? !(i % module.get.gapRatio())
-                      ? $('<li class="label">' + labelText + '</li>') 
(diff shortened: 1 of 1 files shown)


lubber-de (changes requested):
This only works, if one definately uses `interpretLabels`
Without that (which is probably the default for most users) lots of "0" ticks are shown
https://jsfiddle.net/lubber/owturj5e/1/
image
In order to fix this, the `get.label` method should also use a fallback of 1 for the step if the step is set to 0

```javascript
      case settings.labelTypes.number:
       return Math.round(((value * (module.get.step() === 0 ? 1 : module.get.step())) + module.get.min()) * precision ) / precision;
```

TheJltres:
Thanks lubber, commit done!

src/definitions/modules/slider.js
```diff
@@ -205,28 +205,26 @@ $.fn.slider = function(parameters) {
...
+                $label = (labelText !== "")
+                  ? !(i % module.get.gapRatio())
+                    ? $('<li class="label">' + labelText + '</li>')
+                    : $('<li class="halftick label"></li>')
+                  : null,
+                ratio  = i / len
+              ;
+              if($label) {
+                module.update.labelPosition(ratio, $label);
+                $labels.append($label);
```
ko2in, line 227: The `for` loop shouldn't execute inside `else` block since it needs to wrap inside the `labels` container?

TheJltres: Else block is only creating `auto labels` ul, if block is only asuming you have already created, no?
So if it goes to if statement it will only empty all labels and no one will be added, we could add auto class to `.labels` element.

TheJltres: Also when `autoLabel` is called, and if `labels` with no `auto` class is found, it goes to `customLabel`.
So in `autoLabel` if `labels` is found it should already have `auto` class.

ko2in: You're right. I misread the `$labels.empty()` as it remove the `ul` element and thus the for loop would be failed in some situations.

lubber-de (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
@all-contributors please add @TheJltres for code

PR #1693, opened 2020-10-03
[slider]: onChange is triggered without moving slider

### LINKED ISSUE #1382: Range slider get triggered even user doesn't move it
I'm using Fomantic 2.8.3 and I noticed that the `onChange` callback is triggered even if the slider is not moved.
Both `onChange` and `onMove` are triggered just clicking on the slider, without any values change.

Here is a Plunker: http://jsfiddle.net/5nmhgvdp
As you can see if you just click on the slider, without change value, it get always triggered.

I would like to have it triggered only when users move the slider and only when values are different from previous move.

MrL1605:
##  Description

Store previous value to skip onChange trigger if no change is found in value.

I've assumed here that `onChange` is to behave differently than `onMove`, let me know if this assumption is incorrect and there are changes needed.
 
Added a new variable `previousValue` and compare it with new value before triggering `onChange` and `onMove`. 
If new changes are found then trigger both or else trigger only `onMove`.

## Testcases
Fiddle: https://jsfiddle.net/MrL1605/5vp4u38g/16/
- Testcase 1:
  Sliding the slider and going back to same position should trigger `onMove` on every position and should not trigger `onChange`.
- Testcase 2:
  Simply clicking on standard slider tab should not trigger `onChange`, but `onMove` should be triggered. (standard and range)
- Testcase 3:
  If a standard slider is already at `x` position, calling the behaviour to set value `x` should trigger `onMove` and not `onChange`.
- Testcase 4:
  If a range slider has positions set to `a1` and `b1`, calling the behaviour to set range as `a2` to `b2`, 
  - if the abs difference `a1` and `b1` is different from abs difference of `a2` and `b2`, then this should trigger both `onMove` and `onChange`.
  - if the abs difference `a1` and `b1` is same as abs difference of `a2` and `b2`, then this should trigger only `onMove` and not `onChange`.

## Screenshot (if possible)
No CSS changes

## Closes
Closes #1382

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -88,6 +88,7 @@ $.fn.slider = function(parameters) {
         precision,
         isTouch,
         gapRatio = 1,
+        previousValue,
 
         initialPosition,
         initialLoad,
@@ -331,7 +332,13 @@ $.fn.slider = function(parameters) {
               } else {
                 $currThumb = module.determine.closestThumb(newPos);
               }
+              if (previousValue === undefined) {
+                previousValue = module.get.currentThumbValue();
+              }
+            } else if (previousValue === undefined) {
+                previousValue = module.get.value();
             }
+
             if(!module.is.disabled()) {
               module.bind.slidingEvents();
             }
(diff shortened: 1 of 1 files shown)


lubber-de (approved):
LGTM 👍

prudho (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
@all-contributors please add @MrL1605 for code

PR #1695, opened 2020-10-05
[Form Validation] Set optional field behavior

### LINKED ISSUE #1500: [Form Validation] Add optional fields programmatically
Hey there dear FUI Team,

right now my forms are all dynamically created (in combination with vue.js) and displayed in one modal dialog.

Depending on the content the user can choose from, the form is reconstructed (fields, labels, prompts, validation rules).

I can use `$el.form( "add rule", str_identifier, arr_rules);` to add rules to a specific field.
How can I set a field to optional like (https://fomantic-ui.com/behaviors/form.html#optional-fields) ?

Like `$el.form( "set optional", str_identfier, bool_value);` to un/set an optional field for validation process.

I tried in the js at Line 1653 (in the "module.set" Object:
<pre>optional: function(identifier, boolValue) {
  $.each(validation, function(fieldName, field) {
    if (identifier == fieldName || identifier == field.identifier) {
      field.optional = boolValue;
    }
  });
  module.verbose((boolValue ? 'S' : 'Uns') + 'etting field as optional', identifier);
}</pre>

If the first validation passes (no validation errors found), all is doing well.
If I insert an error (validation error occures and show error messages), the fields are no longer flagged as optional.

Is there a reset to all fields after one validation interation?

Keep on rocking,
Matthias

Giandrop:
## Description
Adds the possibility to use `$('.ui.form').form('set optional', fieldName, bool);` to set or unset a field as optional for validation.

## Testcase
https://jsfiddle.net/giandrop/gbwunp7j/14/

## Closes
#1500

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1165,6 +1165,17 @@ $.fn.form = function(parameters) {
                 }
               }
             });
+          },
+          optional: function(identifier, bool) {
+            bool = (bool !== undefined)
+              ? bool
+              : true
+            ;
+            $.each(validation, function(fieldName, field) {
+              if (identifier == fieldName || identifier == field.identifier) {
+                field.optional = bool;
+              }
+            });
           }
         },
 
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1167,10 +1167,7 @@ $.fn.form = function(parameters) {
             });
           },
           optional: function(identifier, bool) {
-            bool = (bool !== undefined)
-              ? bool
-              : true
-            ;
+            bool = (bool !== false);
             $.each(validation, function(fieldName, field) {
               if (identifier == fieldName || identifier == field.identifier) {
                 field.optional = bool;


lubber-de (approved):
LGTM

ko2in (dismissed):
LGTM.

prudho (approved):
LGTM

src/definitions/behaviors/form.js
```diff
@@ -1165,6 +1165,17 @@ $.fn.form = function(parameters) {
                 }
               }
             });
+          },
+          optional: function(identifier, bool) {
+            bool = (bool !== undefined)
```
exoego, line 1170: Is it intended to use `null` ?

Giandrop: @exoego 
I thought it makes more sense to set it as true by default in case you omit that argument: `$('.ui.form').form('set optional', fieldName)`
Should I change this?

lubber-de: i second @Giandrop. "set optional" already sounds to enable it by default ( -> `true`). So i think he should not change anything here. @exoego Or did i miss your point here?

exoego: `null` is false-y value, so it is weird for me `null` is handled as `true`.

exoego: I wonder what are the use cases users want to set string, int or null to `field.optional`.

lubber-de: Ah, i see. So should he change it to
`bool = (bool !== false);` then?

ko2in: I think it's better to check whether the argument is boolean if it expects only boolean.

lubber-de: @ko2in So, you agree to my suggestion? Or do you have another suggestion what/how to change?
@exoego  Your final decision about any required change?

ko2in: @lubber-de I agree your suggestion.

exoego: `!== false` looks good to me

lubber-de: @Giandrop Could you please change this line accordingly ? 😉

lubber-de (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
@Giandrop Thanks. I saw you are working on the related changes for the docs :+1: But it seems haven't created a PR for in https://github.com/fomantic/Fomantic-UI-Docs/pulls yet

lubber-de:
@all-contributors please add @Giandrop for code

PR #1704, opened 2020-10-11
[Progress] Support right aligned progress bar

lubber-de:
## Description
This PR allows to display the progress bar aligned to the right growing to the left by using `ui right aligned progress`

## Testcase
https://jsfiddle.net/lubber/zrx39u1b/15/

## Screenshot
rightprogress

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/7017

### DIFF
--- a/src/definitions/modules/progress.less
+++ b/src/definitions/modules/progress.less
@@ -24,7 +24,7 @@
 
 .ui.progress {
   position: relative;
-  display: block;
+  display: flex;
   max-width: 100%;
   border: @border;
   margin: @margin;
@@ -89,7 +89,6 @@
 
 & when (@variationProgressRightAligned) {
   .ui.right.aligned.progress {
-    display: flex;
     justify-content: flex-end;
     & .bar > .progress {
       left: @progressRightAlignedLeft;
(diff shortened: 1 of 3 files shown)


src/definitions/modules/progress.less
```diff
@@ -87,6 +87,17 @@
...
 
+& when (@variationProgressRightAligned) {
+  .ui.right.aligned.progress {
+    display: flex;
+    justify-content: flex-end;
+    & .bar > .progress {
+      left: @progressRightAlignedLeft;
+      right: @progressRightAlignedRight;
+    }
+  }
```
ko2in, line 98: I think it's better to specify `display: flex` to `ui.progress` by default instead of changing for `right.aligned`. Basically, it should be the same display type and the direction can be styled with `justify-content: flex-end` as right aligned and `justify-content: flex-start` which will be default and can be ignored.

lubber-de: Very good idea 👍  I adjusted it accordingly by 198168aa0eaf4857d53b4554038242f18591442b
Please review again

ko2in (approved):
LGTM.

exoego (approved):
LGTM

ColinFrick (approved):

PR #1710, opened 2020-10-15
[Header] Icon content was not aligned

### LINKED ISSUE #1708: "center aligned icon header" miss aligned using "content"
# Bug Report
From v2.7.5 to v2.8.7 something got broken with "center aligned icon header" that miss aligned the "content" class.

## Steps to reproduce
1. Please see the JSFiddle
2. Removing the &lt;div class="content"&gt; work around the issue
3. Re-adding CSS .ui.icon.header .content { display: block; } works too

Seems like the CSS order changed from 2.7 to 2.8

## Expected result
Center aligned content (using a "content" class)

## Actual result
Left aligned content

## Testcase
https://jsfiddle.net/sLnaf9yg/

## Version
2.8.7

lubber-de:
## Description
The specificity has changed for icons by #1593 . This had an impact when using an `aligned icon header` to not correctly be aligned anymore because of the wrong display setting.
This PR adjusts the specificity again

## Testcase
https://jsfiddle.net/lubber/j294u6mx/7/

### Broken
image

### Fixed
image

## Closes
#1708

### DIFF
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -234,7 +234,7 @@
   .ui.icon.header .corner.icon {
     font-size: @cornerIconHeaderSize;
   }
-  .ui.icon.header .content {
+  .ui.ui.icon.header .content {
     display: block;
     padding: 0;
   }
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -111,7 +111,7 @@
 }
 
 /* After Icon */
-.ui.header > i.icon + .content {
+.ui.header:not(.centered):not(.aligned) > i.icon + .content {
   padding-left: @iconMargin;
   display: table-cell;
   vertical-align: @contentIconAlignment;
@@ -234,7 +234,7 @@
   .ui.icon.header .corner.icon {
     font-size: @cornerIconHeaderSize;
   }
-  .ui.ui.icon.header .content {
+  .ui.icon.header .content {
     display: block;
     padding: 0;
   }


exoego (approved):
LGTM

prudho (approved):
LGTM

ko2in (commented):
Is it possible to remove `display: table-cell` from `.ui.header > i.icon + .content`? I've tested with this fiddle (https://jsfiddle.net/ko2in/jh62gstx/) and it doesn't break anything.

FUI

I usually prefer to avoid unnecessary overrides as much as possible.

lubber-de:
@ko2in The override is necessary, because the `block` is the previous behavior even in SUI
When removing the `table-cell` it will break non aligned icon headers 
See
breakingtablecell

ko2in:
@lubber-de I see. Then I think it's better to omit for center header instead of overrides like:
```
.ui.header:not(.center) > i.icon + .content {
  padding-left: .75rem;
  display: table-cell;
  vertical-align: middle;
}
```
It looks more semantic to me. And it wouldn't have unnecessary left padding too for center icon header. I've tested with this [page](https://fomantic-ui.com/elements/header.html) and it works well.

lubber-de:
@ko2in Agreed. Nice catch 👍 
However, i had to change your suggestion to `:not(.centered):not(.aligned)` because otherwise `right aligned` as well as `centered` would not work anymore (there exists an alias `centered` for `center aligned`, that's why).
I adjusted the PR accordingly and reverted the specificity change, please review again @prudho  @exoego

exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1713, opened 2020-10-16
[Form] Option errorfocus to auto focus the first error field on form validation

### LINKED ISSUE #1659: Focus to first input field after form validation.
# Feature Request
I am working on an update of one of my online shops and it's UI is based on Fomantic.
At the one-page checkout process are several input fields (name, address, phone number etc.)
The form validation works perfectly.
But as i am using the onFailure event to show a toast, i also wanted it to jump to the field where the validation fails.
It would be great to have the example on the project's documentation IMHO.

- Why should we add this feature/enhancement
As i could not find any onFailture example i had to search the web for it. It would be great to have it inside the documentation.

- What problem(s) does it solve
To find out how the form validations work and can like in my case substitute the inline HTML error code with a toast, alert, console.log or similar.

- Complexity
Easy Peasy.

- How we should implement it
Putting the example in the following page as example:
https://fomantic-ui.com/behaviors/form.html

I have found one example of doing it on the Semantic Github pages. But as we all know SUI is not updated "that" "much" :)
BTW: the code on said page did not work.
https://github.com/Semantic-Org/Semantic-UI/issues/3059

So after searching the web for 5 minutes i found a asian page that had an example that worked out for me.
http://blog.starbug1.com/archives/1886

The code on that page is this:
`onFailure: function(erros, field) {
    var errorFields = $('.field.error input,.field.error select', this);
    setTimeout(function(){
      errorFields.first().focus();
    }, 100);
    return false;
  },`

lubber-de:
## Description
This PR adds a new option `errorFocus` (default `false` to stay backward compatible).
When set to `true`, the first found error field will be focussed ready to immediatly change the values.
When set to a string, it will be used as a dom selector to be able to focus anything else (most probably the error message box)

Until now the user always had to click into the related field before

### Before
https://jsfiddle.net/lubber/r91wax23/1/

### After
https://jsfiddle.net/lubber/r91wax23/2/

## Screenshots
|Before|After|
|-|-|
|errorfocusbad|errorfocusgood|

## Closes
#1659

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1209,6 +1209,9 @@ $.fn.form = function(parameters) {
               if(event && $module.data('moduleApi') !== undefined) {
                 event.stopImmediatePropagation();
               }
+              if(settings.errorFocus) {
+                $group.filter('.' + className.error).first().find(selector.field).focus();
+              }
               if(ignoreCallbacks !== true) {
                 return settings.onFailure.call(element, formErrors, values);
               }
@@ -1514,6 +1517,7 @@ $.fn.form.settings = {
 
   autoCheckRequired : false,
   preventLeaving    : false,
+  errorFocus        : false,
   dateHandling      : 'date', // 'date', 'input', 'formatter'
 
   onValid           : function() {},
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1210,7 +1210,22 @@ $.fn.form = function(parameters) {
                 event.stopImmediatePropagation();
               }
               if(settings.errorFocus) {
-                $group.filter('.' + className.error).first().find(selector.field).focus();
+                var focusElement, hasTabIndex = true;
+                if (typeof settings.errorFocus === 'string') {
+                  focusElement = $(settings.errorFocus);
+                  hasTabIndex = focusElement.is('[tabindex]');
+                  // to be able to focus/scroll into non input elements we need a tabindex
+                  if (!hasTabIndex) {
+                    focusElement.attr('tabindex',-1);
+                  }
+                } else {
+                  focusElement = $group.filter('.' + className.error).first().find(selector.field);
+                }
+                focusElement.focus();
+                // only remove tabindex if it was dynamically created above
+                if (!hasTabIndex){
+                  focusElement.removeAttr('tabindex');
+                }
               }
               if(ignoreCallbacks !== true) {
                 return settings.onFailure.call(element, formErrors, values);


prudho (dismissed):
LGTM

If I can make another suggestion, it would be cool to allow focus on message panel too, if it's enabled.

lubber-de:
> If I can make another suggestion, it would be cool to allow focus on message panel too, if it's enabled.

Message panel?  You mean the error message box below the form? This isn't an input field/textarea.... Could you please explain a bit more in detail?

prudho:
Yup, the message box which can be placed on top or bottom of the form. In case of very long form, panel is sometimes not visible on form validation, it would be cool to jump on it in this case.

ko2in (dismissed):
LGTM. Nice feature. Yes, I think it would also be nice if it could scroll to the message box in case the message box is out of visible area like @prudho said.

lubber-de:
Well, ok..but wouldn't this mean it's going to scroll the message box to the top of the screen? If the form itself it located above that message box the user would have to scroll up again manually anyway.

So, do you suggest to somehow focus either the first error field _or_ the message panel? 
So for example if `errorFocus` is a string, it's supposed to be the dom selector to be used instead of the first found error field? This way the user can decide on his/her own what to be focussed in case of an invalid form.

ko2in:
What if `errorFocus` also scroll the page to the top of the form? So, the error message would also be visible and the field is focused. I think we can ignore for the message at the bottom since it stays near the submit button and it would always be easy to catch.

lubber-de:
> What if `errorFocus` also scroll the page to the top of the form? 
When focussing into a specific field it automatically scrolls accordingly. So i would add my previous suggestion: If errorFocus is true is behaves as already implemented (focussing first error field), if it's a string, it's used as a dom selector, so you can focus (implies scrolling) whatever element you like

lubber-de:
I adjusted the PR and description accordingly. Please review again

ko2in:
It doesn't seem to work with class element. `errorFocus: '#element'` works, but `errorFocus: '.element'` doesn't.

lubber-de:
> It doesn't seem to work with class element. `errorFocus: '#element'` works, but `errorFocus: '.element'` doesn't.

🤔 Cannot reproduce, works fine when i use `errorFocus: '.ui.error.message' `
See https://jsfiddle.net/lubber/r91wax23/4/

of course the dom selector has to be unique, otherwise `focus()` won't work

ko2in (approved):
LGTM. It works. There was a typo in my class name and I haven't notice this.

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/236

exoego (approved):
LGTM

PR #1724, opened 2020-10-23
[Slider] Remove old dist files for renamed slider (previously "range") 

### LINKED ISSUE #1723: Range Slider does not fill range between buttons
In dist/components/range.css in line 40 the width: 0; is not commented out as in dist/semantic.css. This causes the range slider to not fill the range between the two buttons as the width is fixed to 0.

lubber-de:
## Decsription
The `dist` folder still had the `range.*` files, although this component was renamed to  `slider` in 2.7.0, thus those files are not generated anymore when fomantic is built

## Closes
#1723


exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1736, opened 2020-11-01
[Icon] Labels in icon groups where shown italic

lubber-de:
## Description
When making use of a `floating label` inside an icon group, the label text is shown italic (because icon groups use the `i` tag

## Testcase
https://jsfiddle.net/lubber/786vh0Lt/3/

## Screenshot
image


exoego (approved):
Nice catch !!

ColinFrick (approved):

PR #1743, opened 2020-11-03
fix(chore): use url encoded semicolons so gulp-concat doesnt break

lubber-de:
## Description
This PR tries to solve the issue that `gulp-concat-css` doesn't find the google path because of the semicolons
This issue does only occur when building inside a node_modules folder. using `npx gulp build` inside a forked git repo folder is /was still working while having semicolons inside the google path

## Closes (hopefully)
#1738


ko2in (approved):
LGTM. Nice catch :+1:

ColinFrick (approved):

PR #1744, opened 2020-11-03
fix(checkbox): remove the change trigger for enable and disable behavior

prudho:
As discussed in #1728, triggering the `onChange` callback in the `enable` and `disable` behaviors is useless so I've just removed it.

Other question is: should we consider this as a breaking change and wait 2.9 to release it ?


lubber-de:
If some existing projects relies on the onchange event when enabling/disabling  a checkbox this should be considered a breaking change, IMHO, yes.
Let' vote 😄

lubber-de (approved):
LGTM

ko2in (approved):
LGTM. I also think it should be a breaking change.

exoego (approved):
LGTM

> https://github.com/fomantic/Fomantic-UI/discussions/1728#discussioncomment-113240
> `onchange` should IMHO only be called when the check state changes. Even, if you disable a checked checkbox, it will still be checked (thus: did not change)

This sounds reasonable to me.

PR #1746, opened 2020-11-03
[Calendar] Dont draw complete adjacent rows

lubber-de:
## Description
Whenever the first of a month is shown within the first 5 columns of a calendar, then, according to the months amount of days, a full adjacent row is added to the calendar, which is unnecessary.

This PR skips the last row on rendertime if it would consists adjacent days only, making sure the calendar concentrates as much on the current month

#### Current
https://jsfiddle.net/lubber/0m192y5b/1/

#### Enhanced
https://jsfiddle.net/lubber/0m192y5b/2/

## Screenshot

|Current|Enhanced|
|-|-|
|calendarfixedrows|calendardynamicrows|

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -358,8 +358,9 @@ $.fn.calendar = function(parameters) {
 
               var tbody = $('<tbody/>').appendTo(table);
               i = isYear ? Math.ceil(year / 10) * 10 - 9 : isDay ? 1 - firstMonthDayColumn : 0;
+              rowsLoop:
               for (r = 0; r < rows; r++) {
-                row = $('<tr/>').appendTo(tbody);
+                row = $('<tr/>');
                 if(isDay && settings.showWeekNumbers){
                     cell = $('<th/>').appendTo(row);
                     cell.text(module.get.weekOfYear(year,month,i+1-settings.firstDayOfWeek));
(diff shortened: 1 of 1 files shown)


src/definitions/modules/calendar.js
```diff
@@ -358,8 +358,9 @@ $.fn.calendar = function(parameters) {
 
               var tbody = $('<tbody/>').appendTo(table);
               i = isYear ? Math.ceil(year / 10) * 10 - 9 : isDay ? 1 - firstMonthDayColumn : 0;
+              rowsLoop:
```
exoego, line 361: Perhaps `var` missing ?
If no `var`, the variable `rowsLoop` is accidentally exposed globally.

exoego: Aha, this is not variable, but a label for loop !!

exoego (approved):
LGTM

prudho:
There is a `constantHeight` setting to adjust this behavior (default: `true`).

lubber-de:
> There is a `constantHeight` setting to adjust this behavior (default: `true`).

OMG..... I totally missed that setting and it works just as expected!
So i am closing this PR and suggest to change the default value to `false` for 2.9 instead. 

Thanks for pointing that out @prudho , you already did a good job 2 years ago ! 👍

PR #1751, opened 2020-11-04
[Slider] support control over change trigger when setting a value or range

### LINKED ISSUE #1750: [Slider] Add fireChange parameter to slider setting behaviors
# Feature Request
The **Calendar**'s `set date` behavior has the `fireChange` (default false) which toggles whether or not setting the date triggers its `onSelect` function. This is amazingly helpful:

Screen Shot 2020-11-04 at 2 05 47 PM

It would be wonderful if the **Slider** had this option as well, for both `set value` and `set rangeValue`. Currently, it has no such option, and the slider's `onMove` function fires when `set value` is called:

Screen Shot 2020-11-04 at 2 05 55 PM

I see an undocumented `setting.fireOnInit` in the js, but this defaults to false, so it must do something else?

lubber-de:
## Description
This PR adds a new option to the `set value` or `set rangeValue` behavior to optionally avoid the `onChange` callback

The change is backward compatible, because when the new parameter is omitted, it behaves like before (triggering the callback)

`.slider('set value',10,false)`

## Before
The onChange callback is triggered immediatly
`.slider('set value',10)`
https://jsfiddle.net/lubber/pwfd81rL/2/

## After
The onChange callback is not triggered
`.slider('set value',10,false)`
https://jsfiddle.net/lubber/pwfd81rL/3/

## Closes
#1750

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -861,11 +861,12 @@ $.fn.slider = function(parameters) {
         },
 
         set: {
-          value: function(newValue) {
+          value: function(newValue, fireChange) {
+            fireChange = fireChange !== false;
             var toReset = previousValue === undefined;
             previousValue = previousValue === undefined ? module.get.value() : previousValue;
             module.update.value(newValue, function(value, thumbVal, secondThumbVal) {
-              if (!initialLoad || settings.fireOnInit){
+              if ((!initialLoad || settings.fireOnInit) && fireChange){
                 if (newValue !== previousValue) {
                   settings.onChange.call(element, value, thumbVal, secondThumbVal);
                 }
@@ -876,7 +877,8 @@ $.fn.slider = function(parameters) {
               }
             });
           },
-          rangeValue: function(first, second) {
+          rangeValue: function(first, second, fireChange) {
+            fireChange = fireChange !== false;
             if(module.is.range()) {
               var
                 min = module.get.min(),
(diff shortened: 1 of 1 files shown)


src/definitions/modules/slider.js
```diff
@@ -861,11 +861,12 @@ $.fn.slider = function(parameters) {
...
 
         set: {
-          value: function(newValue) {
+          value: function(newValue, fireChange) {
+            fireChange = fireChange !== false;
             var toReset = previousValue === undefined;
             previousValue = previousValue === undefined ? module.get.value() : previousValue;
             module.update.value(newValue, function(value, thumbVal, secondThumbVal) {
-              if (!initialLoad || settings.fireOnInit){
+              if ((!initialLoad || settings.fireOnInit) && fireChange){
```
exoego, line 869: > The change is backward compatible, because when the new parameter is omitted, it behaves like before (triggering the callback)

Hmm, it seems breaking change for me.
IIUC, if `fireChange` is not supplied, it will be `undefined`, which is `false`-y.
So this `if` block are skipped, therefore `onChange` callback are not invoked.

lubber-de: @exoego fireChange is fetched at the start of the function by (same logic is inside calendar)
```javascript
            fireChange = fireChange !== false;
```

So, if the `fireChange` is not given (undefined) it's converted to `true` automatically to preserve the current logic
image

exoego: Ah, I see.

exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1755, opened 2020-11-06
[Icons] allow corner icons and icon groups to be bordered

### LINKED ISSUE #1753: [Icons] Bordered icon together with a corner icon
# Bug Report
I want to use a cornered icon, inside a bordered frame (bordered icon). I want to put an edit pen in the corner of a book, with a frame around it.

## Steps to reproduce
```html
<i class="bordered icons">
  <i class="book icon"></i>
  <i class="corner pen icon"></i>
</i>
```

## Expected result
How a bordered icon and a corner icon works, but together at the same time.

## Actual result
Corner icon as intended, without a border

## Testcase
https://jsfiddle.net/oz9hkube/5/

## Version
Am using Fomantic-UI 2.8.4, tested with 2.8.7 in the jsfiddle linked above with the same results.

Yrlish:
## Description
I added functionality to make an icon group bordered.

## Testcase
```html
<p>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="corner pen icon"></i>
    </i>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="user icon"></i>
    </i>
    <i class="huge bordered icons">
        <i class="big red dont icon"></i>
        <i class="black user icon"></i>
    </i>
</p>
<p>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="corner pen top left icon"></i>
    </i>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="corner pen top right icon"></i>
    </i>
    <br>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="corner pen bottom left icon"></i>
    </i>
    <i class="huge bordered icons">
        <i class="book icon"></i>
        <i class="corner pen bottom right icon"></i>
    </i>
</p>
```

## Screenshot (if possible)
image

## Closes
#1753

### DIFF
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -273,7 +273,9 @@ each(@colors, {
       color: @l;
     }
     i.inverted.bordered.@{color}.icon.icon.icon.icon.icon,
-    i.inverted.circular.@{color}.icon.icon.icon.icon.icon {
+    i.inverted.circular.@{color}.icon.icon.icon.icon.icon,
+    i.inverted.bordered.@{color}.icons,
+    i.inverted.circular.@{color}.icons {
       background-color: @c;
       color: @white;
     }
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -372,4 +372,49 @@ i.icons {
   }
 }
 
+& when ((@variationIconGroups) or (@variationIconCorner)) and (@variationIconBordered) {
+  i.bordered.icons {
+    width: @borderedSize;
+    height: @borderedSize;
+    box-shadow: @borderedShadow;
+  }
+  i.bordered.icons i.icon:first-child {
+    position: absolute;
+    transform: translateX(-50%) translateY(-50%);
+  }
+
+  & when (@variationIconCorner) {
+    i.bordered.icons .corner.icon {
+      top: auto;
+      left: auto;
+      right: .8em;
+      bottom: .8em;
+    }
+    i.bordered.icons .icon.corner[class*="top right"] {
+      top: .8em;
+      left: auto;
+      right: .8em;
+      bottom: auto;
+    }
+    i.bordered.icons .icon.corner[class*="top left"] {
+      top: .8em;
+      left: .8em;
+      right: auto;
+      bottom: auto;
+    }
+    i.bordered.icons .icon.corner[class*="bottom left"] {
+      top: auto;
+      left: .8em;
+      right: auto;
+      bottom: .8em;
+    }
+    i.bordered.icons .icon.corner[class*="bottom right"] {
+      top: auto;
+      left: auto;
+      right: .8em;
+      bottom: .8em;
+    }
+  }
+}
+
 .loadUIOverrides();
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -372,47 +372,78 @@ i.icons {
   }
 }
 
-& when ((@variationIconGroups) or (@variationIconCorner)) and (@variationIconBordered) {
-  i.bordered.icons {
+& when ((@variationIconGroups) or (@variationIconCorner))
+   and ((@variationIconBordered) or (@variationIconCircular)) {
+  /*************************************************
+     Bordered/circular with corner or group icons
+  *************************************************/
+  i.bordered.icons,
+  i.circular.icons {
     width: @borderedSize;
     height: @borderedSize;
     box-shadow: @borderedShadow;
+    vertical-align: middle;
   }
-  i.bordered.icons i.icon:first-child {
+  i.circular.icons {
+    border-radius: 500em !important;
+  }
+  i.bordered.icons i.icon:first-child,
+  i.circular.icons i.icon:first-child {
     position: absolute;
     transform: translateX(-50%) translateY(-50%);
   }
 
+  & when (@variationIconInverted) {
+    /* Inverted Icon */
+    i.circular.inverted.icons,
+    i.bordered.inverted.icons {
+      border: none;
+      box-shadow: none;
+    }
+
+    i.inverted.bordered.icons,
+    i.inverted.circular.icons {
+      background-color: @black;
+      color: @white;
+    }
+  }
+
   & when (@variationIconCorner) {
-    i.bordered.icons .corner.icon {
+    /* Corner Icon */
+    i.bordered.icons .corner.icon,
+    i.circular.icons .corner.icon {
       top: auto;
       left: auto;
-      right: .8em;
-      bottom: .8em;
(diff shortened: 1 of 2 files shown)


src/definitions/elements/icon.less
```diff
@@ -372,4 +372,49 @@ i.icons {
...
+      left: .8em;
+      right: auto;
+      bottom: .8em;
+    }
+    i.bordered.icons .icon.corner[class*="bottom right"] {
+      top: auto;
+      left: auto;
+      right: .8em;
+      bottom: .8em;
+    }
```
ko2in, line 416: Are these position unit (.8em) fit for the other icon size? In your test case, you only use for huge icons. So, I'm not sure if you forgot to test with the other icon sizes.

Can you please also provide JS fiddle? So, we can test your changes easily.

Yrlish: I'm not sure how do to the jsfiddle correctly in this case, so I copied in the output from the complied less code into it.
https://jsfiddle.net/fpb9qLu1/

`.8em` works fine in all sizes.

lubber-de (changes requested):
### Thanks for your time enhancing Fomantic-UI 👍 

I have some change requests, i am going to explain now:
To reproduce everything i am mentioning below, please see https://jsfiddle.net/lubber/8gLrbvtz/52/

#### Please also support `circular icons` as well
It's just adding the `circular` variant to all your existing css selectors for `bordered` and adding a border-radius to `circular icons`) (see my jsfiddle above)
It exists for non icon groups, so we should make this consistent (and it's only a small change to your existing code anyway 🙂 )

image

#### Please add `vertical-align;middle;` to `i.bordered.icons` (and `circular` as mentioned above) because otherwise they wont appear inline correctly anymore

##### Your proposal
image

##### My suggestion
image

#### Please move your corner offset value into a variable like `@borderedCornerOffset`
Please also use a default value of `1.15em` instead of `0.8em` Why? Because the corner icons should look the same when inside a bordered group as well as inside a non-bordered group

Usual icon groups have their corner icons positioned exactly in the corner. The behavior in the default theme should be the same for consistency

##### Non bordered corner icons
image

##### Your proposed bordered corner icons
image

##### Equal Look bordered corner icons
image

I understand why you have chosen the corner icon to be placed  slightly outside of the icon corner itself (because it looks better while there is more space to the border itself), but we should keep the same look as for usual corner icons. By providing a separate variable, one can adjust this according to their needs (in your case 0.8em)

#### Please support the `inverted` variant as for non grouping borders/circular icon
For a bordered (and circular) icon the inverted variant is supported, so we should also support this for grouping icons as well.
Everything is already in place inside the LESS file, we just have to add the .icons selector to the existing css for single icon
Don't forget to support the colors inside the each loop as well then (not covered in my jsfiddle above)

image

Yrlish:
@lubber-de Great feedback! I have used your jsfiddle and applied the requested changes. 
Your argument for `1.15em` makes sense but you also understood my reasoning for it. 👌 

Hopefully I have not missed something. 🙂

lubber-de:
@Yrlish Almost perfect. Only thing missing is the color loop adjustment in 
https://github.com/fomantic/Fomantic-UI/blob/ce85cf35b64d80c51482047581059c2c31fcea58/src/definitions/elements/icon.less#L275-L280

The selector should be enhanced by

```css
i.inverted.bordered.@{color}.icons,
i.inverted.circular.@{color}.icons,
```

Yrlish:
@lubber-de there we go! This one is a bit more annoying to get into a jsfiddle tho.

lubber-de:
@Yrlish Awesome! Last suggestion: I think, you can remove the `!important` from the border radius ,  try in your jsfiddle, it works fine without (yes, the single icon also got that , but i also think its not necessary there as well)

Yrlish:
@Yrlish Yes, that is true. I just copied it over from the single icon stuff. I tested without and it works. I also verified, there is nothing it is overriding anyways, when I checked with the Inspect Element devtool with my circular icons tests.

lubber-de (dismissed):
LGTM 👍

src/definitions/elements/icon.less
```diff
@@ -372,4 +374,80 @@ i.icons {
...
+    transform: translateX(-50%) translateY(-50%);
+  }
+
+  & when (@variationIconInverted) {
+    /* Inverted Icon */
+    i.circular.inverted.icons,
+    i.bordered.inverted.icons {
+      border: none;
+      box-shadow: none;
+    }
```
lubber-de, line 404: Let's combine those. The selector is the same

Yrlish: Oh, lol. Didnt even realize.

lubber-de (approved):
LGTM

src/definitions/elements/icon.less
```diff
@@ -372,4 +374,76 @@ i.icons {
...
+      bottom: @borderedGroupCornerOffset;
+    }
+    i.bordered.icons .icon.corner[class*="bottom right"],
+    i.circular.icons .icon.corner[class*="bottom right"] {
+      top: auto;
+      left: auto;
+      right: @borderedGroupCornerOffset;
+      bottom: @borderedGroupCornerOffset;
+    }
+  }
```
ko2in, line 446: I think we don't need this, it's the same as line 411 to 417 which is default.

Yrlish: This is copied from the existing corner implementation. "bottom right" and default has the same values.

lubber-de: 🤔 Yeah, but @ko2in is technically right. We could even remove the `bottom right` variant from the usual corner specification.
@fomantic/admins Should we clean this up here or keep the "bottom right" variant nevertheless?

ko2in: If we need to keep `bottom right` variant, we should merge the selectors instead of copy the same definitions.

lubber-de: @ko2in Yes, agreed

exoego (approved):
LGTM 🥇

src/definitions/elements/icon.less
Yrlish: So does this mean I should combine the selectors the default one and the bottom right one?

lubber-de: yes, thats the easiest compromise so save CSS declarations and still have the bottom right selector in place.

Yrlish: Alright, I will do this later. 🙂

Yrlish: Done

lubber-de (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
@all-contributors please add @Yrlish for code

PR #1758, opened 2020-11-10
[Table] Allow pointer events for the first header cell in definition table

ko2in:
## Description
The first header cell of definition table disables the pointer event which causes the problem when the cell contains the content that are needed the pointer event such as dropdown, input field or the content from title attribute is unable to be seen.

This PR changes the behavior of pointer event to become available if there is content inside the cell.

## Testcase
Put the mouse cursor over the header `Foo title` and `subtitle` in the first cell.

**Before (Tooltip disappear):** https://jsfiddle.net/ko2in/furswynh/2/
**Now (Tooltip appears):** https://jsfiddle.net/ko2in/furswynh/3/

## Closes
#1756

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -334,6 +334,10 @@
     -moz-transform: scale(1);
   }
 
+  .ui.definition.table > thead:not(.full-width) > tr > th:first-child:not(:empty) {
+    pointer-events: auto;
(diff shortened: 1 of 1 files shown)


lubber-de (approved):
LGTM

exoego (approved):
LGTM

PR #1759, opened 2020-11-11
[Card] Fix the top border thickness of horizontal card

ko2in:
## Description
The content element of the card has top border and the card also has box shadow which makes the card border thicker when 
the card representing the content horizontally.

This PR removes the top border from the card content if it's inside the horizontal card.

## Testcase
**Before:** https://jsfiddle.net/ko2in/9fyxu08L/

**After:** https://jsfiddle.net/ko2in/82ykuofc/

## Screenshot
**Before:**
Card_Fomantic_UI_Before

**After:**
Card_Fomantic_UI_After

## Closes
None


lubber-de:
@ko2in Please rebase on current develop branch...and you pushed all dist files as well... 😨

PR #1761, opened 2020-11-12
[Card]: Inverted header links were not styled properly

lubber-de:
## Description
When header link was inside an inverted cards group (not inside a single card!), it was not styled properly.
This PR now adds the same logic as for single card making link header visible again

#### Before
https://jsfiddle.net/lubber/t8asvzdk/12/
#### After
https://jsfiddle.net/lubber/t8asvzdk/13/
## Screenshot
|Before|After|
|-|-|
|image|image|


ko2in (approved):
LGTM.

exoego (approved):
LGTM 👍

PR #1764, opened 2020-11-13
[Segment] Prevent user selection on disabled and loading segments

lubber-de:
## Description
In `disabled` or `loading` segments, selection of text via mouse or even clicking on links was still possible
I also removed the unnecessary text-shadow setting

## Testcase
Try to select the header inside the two segments with your mouse or click in the link, it should not work
Remove the custom CSS part to see the issue
https://jsfiddle.net/lubber/dz30etxj/26/

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -491,6 +491,8 @@
   .ui.disabled.segment {
     opacity: @disabledOpacity;
     color: @disabledTextColor;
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM
I think this is reasonable, since other modules (e.g. checkbox) also prevent selection on disabled state.

ko2in (approved):
LGTM.

PR #1770, opened 2020-11-15
[Dropdown] multiselect values encoding, label cannot be removed, quotes missing in select dropdown

### LINKED ISSUE #1373: HTML entity encoding behavior in multiselect dropdown
I’m experiencing an issue in Fomantic UI multiselect dropdowns version 2.8.4.

When I choose an item from the multiselect dropdown that contains characters such as & (ampersand) or ‘ (single quote), the hidden input value is encoded into a HTML character entity. However, if I select two items which both contain the special characters, in the hidden input value, the first item is encoded as HTML character entity (eg: &amp;) and the second is unencoded. If I continue to select the third item without the special characters, both items are encoded into HTML entities.

Here is my codepen: https://codepen.io/moistyburger/pen/zYGjQBK

However, if I use semantic-ui version 2.4.1, it does not have such behavior.

lubber-de:
## Description
When a multiple dropdown (non select) was used and selected data has characters like `&` or `'` set, there were stored entity encoded (returning into false positives when comparing later to their original values). Even worse this was only done for all previous selected values, the current to be added value was kept raw already. dropdowns using select tags do not store the selected data in an input field, so the issue does not happen there.

Original data should be kept, as it is already the case for dropdown made out of select tags. Also SUI does not do this.

Additionally this PR fixes a situation when a value has double quotes which led to a js error when trying to remove that label .
Double quotes in select menus are now also kept encoded instead of being removed to have the same behavior for select/non select dropdowns

> Double quotes are always encoded (when text needs to be kept) or removed (everywhere else where they dont make sense like classnames) for security reasons, because all internal templates use them for HTML generation

## Testcase
- Try to select the entries with a & or '
- Watch the hidden inputs content after every selection in the elements console
- Try to remove the `Item 3" inches` label in the first dropdown

### Wrong
- While selecting each previous value gets encoded making it different to the original value
- Trying to remove the `Item 3" inches` label results in a console syntax error 
- The second dropdown, made out of a select, does not show the `"` inside the Item 3 label anymore
https://jsfiddle.net/lubber/fL56791d/13/

### Correct
- Identical selected values
-  `Item 3" inches` label can be removed
- Item 3 shows the `"`
https://jsfiddle.net/lubber/fL56791d/15/

## Screenshots
To see the difference look at the elements console for the hidden input tag holding the selected data. Each time a selected value is added the previous values are encoded.
The second dropdown is made out of a selected tag which works fine already

### Wrong
image
image

### Correct
image
image

## Closes
#1373

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3179,8 +3179,9 @@ $.fn.dropdown = function(parameters) {
           },
           label: function(value, shouldAnimate) {
             var
+              escapedValue  = module.escape.value(value),
               $labels       = $module.find(selector.label),
-              $removedLabel = $labels.filter('[data-' + metadata.value + '="' + module.escape.string(settings.ignoreCase ? value.toLowerCase() : value) +'"]')
+              $removedLabel = $labels.filter('[data-' + metadata.value + '="' + module.escape.string(settings.ignoreCase ? escapedValue.toLowerCase() : escapedValue) +'"]')
             ;
             module.verbose('Removing label', $removedLabel);
             $removedLabel.remove();
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1930,7 +1930,7 @@ $.fn.dropdown = function(parameters) {
               : value
             ;
           },
-          values: function() {
+          values: function(raw) {
             var
               value = module.get.value()
             ;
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

prudho (approved):
Woa, good catch ! This one was a nasty bug.

LGTM 👍

ColinFrick (approved):

PR #1774, opened 2020-11-19
feat(modal): dynamic modal content

### LINKED ISSUE #1407: [Modules] Add user prompt
# Feature Request

Since Fomantic UI presents not only UI but some interactive modules it would be great to add a user prompt module. 

The module should ask user for some decision like "_File 1.txt already exists. Would you like to overwrite this file?_" And also can be invoked from js with text, label and buttons (actions) with callbacks.

## Example

From one hand poor prompt is already can be constructed via toast actions but it displays only toast but prompt (no dimmer, no visual highlights). From another hand prompt can be displayed via modal but with no dynamic content, actions and callbacks.
So I suggest to merge modal UI with toast actions.

## Example

Run `confirm('Implement this?')` in browser console to see default implementation.

lubber-de:
## Description
This PR adds the same dynamic functionality like toast to create a modal just out of properties or reuse a DOM node modal as a template with its content being dynamically set via js properties.
Everything is optional and backward compatible 🙂 
Most of the code is copied /adopted from the toast component so the usage is equal.

This goes along with @prudho fui-alert plugin approach (#1716) which could be more simplified then

### Create a temporary modal just by JS properties
By providing new properties to the modal module and targeting to body you create temporary modals without the need to create markup on your own. The modal is temporary and will be removed from the DOM once closed.

```javascript
$('body').modal({
  title: 'Important Notice',
  class: 'mini',
  closeIcon: true,
  content: 'You will be logged out in 5 Minutes',
  actions: [{
    	text: 'Alright, got it',
        class: 'green'
  }]
}).modal('show');
```
### Reuse an existing modal
Also, you can still prepare a modal markup in your domtree as usual, but set the related content via js properties to reuse the general style
```javascript
$('.ui.modal').modal({
  title: 'Chat rules',
  content: '<div>You can put HTML content right here</div>',
  class: 'inverted fullscreen',
  classContent: 'scrolling',
  actions: [{
    	text: 'That is awesome',
        class: 'green',
        icon: 'exclamation',
        click: function() {
            // do something. return false to prevent closing the modal
        }
  }]
}).modal('show');
```

## Testcase
https://jsfiddle.net/lubber/eo8bdyp3/48/

## Screenshot
dynamicmodal

## Closes
#1407

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -58,6 +58,7 @@ $.fn.modal = function(parameters) {
         selector        = settings.selector,
         className       = settings.className,
         namespace       = settings.namespace,
+        fields           = settings.fields,
         error           = settings.error,
 
         eventNamespace  = '.' + namespace,
@@ -74,7 +75,7 @@ $.fn.modal = function(parameters) {
         $dimmer,
 
         element         = this,
-        instance        = $module.data(moduleNamespace),
+        instance        = $module.hasClass('modal') ? $module.data(moduleNamespace) : undefined,
 
         ignoreRepeatedEvents = false,
 
@@ -91,6 +92,39 @@ $.fn.modal = function(parameters) {
       module  = {
 
         initialize: function() {
+          if(!$module.hasClass('modal')) {
+            module.create.modal();
+            settings.onHidden = function(){
+              module.destroy();
+              $module.remove();
+            };
+          }
+          $module.addClass(settings.class);
+          if (settings.title !== '') {
+            $module.find(selector.title).html(module.helpers.escape(settings.title, settings.preserveHTML));
+          }
+          if (settings.content !== '') {
+            $module.find(selector.content).html(module.helpers.escape(settings.content, settings.preserveHTML)).addClass(settings.classContent);
+          }
+          if(module.has.configActions()){
+            var $actions = $module.find(selector.actions);
+            settings.actions.forEach(function (el) {
+              var icon = el[fields.icon] ? '<i class="' + module.helpers.deQuote(el[fields.icon]) + ' icon"></i>' : '',
+                  text = module.helpers.escape(el[fields.text] || '', settings.preserveHTML),
+                  cls = module.helpers.deQuote(el[fields.class] || ''),
+                  click = el[fields.click] && $.isFunction(el[fields.click]) ? el[fields.click] : function () {};
+              $actions.append($('<button/>', {
+                html: icon + text,
+                class: className.button + ' ' + cls,
+                click: function () {
+                  if (click.call(element, $module) === false) {
+                    return;
+                  }
+                  module.hide();
+                }
+              }));
+            });
+          }
           module.cache = {};
           module.verbose('Initializing dimmer', $context);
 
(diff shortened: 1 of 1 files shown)


src/definitions/modules/modal.js
```diff
@@ -91,6 +92,39 @@ $.fn.modal = function(parameters) {
       module  = {
 
         initialize: function() {
+          if(!$module.hasClass('modal')) {
+            module.create.modal();
+            settings.onHidden = function(){
```
prudho, line 97: We should add an option to define what we want to do when closing the model (`hide` by default, or `remove`).

src/definitions/modules/modal.js
```diff
@@ -58,6 +58,7 @@ $.fn.modal = function(parameters) {
         selector        = settings.selector,
         className       = settings.className,
         namespace       = settings.namespace,
+        fields           = settings.fields,
```
prudho, line 61: Please align with others 😉

prudho (changes requested):

src/definitions/modules/modal.js
lubber-de: O, but i would use `remove` as default, because such modals are basically always created temporary. If we hide them only as default they will be kept in the domtree, because each call to $('body').modal() will create a complete new DOM node.
Imagine 200 "alert" modals in  an app session. It will blew up the DOM tree makes the app slower and slower and, unless the user references to the temporary modal afterwards, it will never be referenced anymore.

Btw, i just copied over your logic from fui-alert, where you also remove the modal once closed 😉

prudho: It was a comment for the whole modal system, including the old style, but yeah, you're right, if we implements this, `remove` should be the default. Let's postpone this until further.

lubber-de: I will change it that way, so the `onHidden` function will only be generated (and removing the modal) when there isn't already a `onHidden` function set, so if anybody wants to keep the modal DOM, then thoese people can simply provide their own `onHidden` function which will do whatever they need.
For that, i will make the default of `onHidden` false and check for their existance before actually calling it. Agreed?

prudho: Yup ! That'll do the trick 👍

lubber-de: Implemented by fc073b4b6d8a1f1067889d259bfcbc8566182757
Please review again

src/definitions/modules/modal.js
lubber-de: fc073b4b6d8a1f1067889d259bfcbc8566182757

src/definitions/modules/modal.js
ko2in: I also agree to use `remove` as default. I've been creating dynamic modals in my project mostly for Ajax content and they are removed from DOM tree after closing. Otherwise, there'll be the duplicate modals when calling again to render and they'll conflict with the ID using by the same modal.

prudho (approved):
LGTM !

ko2in (approved):
LGTM. Nice feature :1st_place_medal:

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/287

PR #1775, opened 2020-11-20
[Modal] Centered variant for any content

lubber-de:
## Description
There currently is no declarative way to display content of a modal center aligned.

- Workaround for header is `icon header` or `ui centered header` or `ui center aligned header`
- Workaround for `actions` is to create an own theme modifying the `@actionAlign` variable
- For the content  itself there is no workaround than creating a css class manually (would also work for actions

This PR copies over the existing variant from `header` (which supports `centered` as well as `center aligned` already) and applies it to any direct sibling of the modal itself, so  `header` or `content` or even `actions` can be displayed centered out of the box in the fomantic way

## Testcase
https://jsfiddle.net/lubber/qb8pe0rj/4/

#### Before
image

#### After
image

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -182,6 +182,11 @@
   border-top:none;
 }
 
+.ui.modal > .centered,
+.ui.modal > .center.aligned {
+  text-align: center;
+}
+
 /*-------------------
        Responsive
 --------------------*/
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -185,6 +185,10 @@
 .ui.modal > .centered,
 .ui.modal > .center.aligned {
   text-align: center;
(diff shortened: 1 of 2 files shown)


prudho (dismissed):
LGTM

ko2in (dismissed):
LGTM.

lubber-de:
I just recognized the action buttons and header was not really centered because of different margins/paddings.
I added a fix, so the right header padding is only set when a close icon is in place (that was the intention of the right padding only) , so if a close icon does not exist at all there is no need of a different right padding.

For the action buttons i added an equal margin when those should be centered

Please review again

ko2in (approved):
LGTM.

prudho (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/253

PR #1777, opened 2020-11-23
fix(calendar): return null instead of undefined in sanitizer

### LINKED ISSUE #1776: [Calendar] Resetting minDate and maxDate to null doesn't re-enable full date range in popup
# Bug Report
Resetting minDate and maxDate to null doesn't re-enable the full date range in the popup.

## Steps to reproduce
1. Initialize a calendar
2. Set its minDate and maxDate
3. Reset its minDate and maxDate to null
4. Open the calendar widget

## Expected result
The full date range should be enabled in the popup.

## Actual result
Only the date range set in step 2 is enabled.

## Testcase
```js
    //just a helper
    Date.prototype.addDays = function(days) {
        var date = new Date(this.valueOf());
        date.setDate(date.getDate() + days);
        return date;
    };

    var today = new Date(new Date().toDateString()),
        $widget = $('#standard_calendar').calendar({ type: 'date' });
        
    $widget.calendar('set minDate', today.addDays(-1));
    $widget.calendar('set maxDate', today.addDays(1));
    $widget.calendar('set minDate', null);
    $widget.calendar('get minDate'); //correctly returns null
    $widget.calendar('set maxDate', null);
    $widget.calendar('get maxDate'); //correctly returns null    
    $widget.calendar('refresh'); //doesn't help either
    $widget.calendar('popup', 'show'); //still only enables the date range set in step 2
```
[Play with my fiddle.](https://jsfiddle.net/cr7myt25/2/)

## Version
2.8.7

ko2in:
## Description
The date sanitizer returns `undefined` when the date argument is not a valid date object or date string which doesn't allow to reset the date value especially for min and max date by passing `null` value via `set minDate` and `set MaxDate`.

This PR makes the `sanitiseDate` method to return `null` if the value of date argument is not valid date object or date string.

## Testcase
**Unable to reset min date and max date:** https://jsfiddle.net/cr7myt25/2/

**Enable to reset min date and max date:** https://jsfiddle.net/ko2in/8ym30fju/8/

## Closes
#1776

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1074,14 +1074,11 @@ $.fn.calendar = function(parameters) {
             return null;
           },
           sanitiseDate: function (date) {
-            if (!date) {
-              return undefined;
-            }
             if (!(date instanceof Date)) {
               date = parser.date('' + date, settings);
             }
-            if (!date || date === null || isNaN(date.getTime())) {
-              return undefined;
+            if (!date || isNaN(date.getTime())) {
+              return null;
             }
             return date;
           },


lubber-de (approved):
LGTM

prudho (approved):
LGTM

h0lg:
Thanks for the fix, guys <3

PR #1779, opened 2020-11-24
fix(dropdown): calculate correct height of current menu if it contains child menu

### LINKED ISSUE #1711: [Dropdown] Search In-Menu positioning does not work correctly
1. Go to Search In-menu example in Docs: https://fomantic-ui.com/modules/dropdown.html#search-in-menu
2. Scoll the window so that the bottom of the window is close to "Filter posts" button. About 100-200px below is the sweet spot.
3. Open menu.

Result: 
Menu goes out of screen. If the dropdown is at the bottom of the page it can even be unaccessible.

Expected result: 
Menu should open above the field if it would go outside of view from the bottom.

search-in-menu

ko2in:
## Description
When the menu contain the child menu, the calculation of menu height is incorrect while evaluating to open upward or downard.

It's because the child menu is hidden when the current menu is in the transition of being open and it doesn't occupy any space in the current menu and the height of the current menu is incorrectly calculated without the height of child menu which let the menu opens upward or downward incorrectly.

This PR checks if the current menu has the child menu and adds the child menu height to the current menu height to give correct assumption whether the menu fits in the screen upward or downward.

Dropdown

## Testcase

Open the fiddle and adjust the page where the dropdown is not fit to open downward. Click the button to open dropdown. The dropdown should open upward.
- Bug: https://jsfiddle.net/ko2in/k6Lwzhjf/
- Fixed: https://jsfiddle.net/ko2in/c28t6wkx/3/

## Screenshot
**Incorrect:**
Dropdown-Downward

**Correct:**
Dropdown-Upward

## Closes
#1711

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3295,7 +3295,7 @@ $.fn.dropdown = function(parameters) {
             return ($menu.length > 0);
           },
           subMenu: function($currentMenu) {
-            return $currentMenu.find(selector.menu).length > 0;
+            return ($currentMenu || $menu).find(selector.menu).length > 0;
           },
           message: function() {
             return ($menu.children(selector.message).length !== 0);
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3527,7 +3527,7 @@ $.fn.dropdown = function(parameters) {
               calculations.menu.offset.top += calculations.context.scrollTop;
             }
             if(module.has.subMenu()) {
-              calculations.menu.height += $menu.find(selector.menu).first().outerHeight();
+              calculations.menu.height += $currentMenu.find(selector.menu).first().outerHeight();
             }
             onScreen = {
               above : (calculations.context.scrollTop) <= calculations.menu.offset.top - calculations.context.offset.top - calculations.menu.height,
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3294,8 +3294,8 @@ $.fn.dropdown = function(parameters) {
           menu: function() {
             return ($menu.length > 0);
           },
-          subMenu: function() {
-            return $menu.find(selector.menu).length > 0;
+          subMenu: function($currentMenu) {
+            return $currentMenu.find(selector.menu).length > 0;
           },
           message: function() {
             return ($menu.children(selector.message).length !== 0);
@@ -3526,7 +3526,7 @@ $.fn.dropdown = function(parameters) {
             if(module.is.verticallyScrollableContext()) {
               calculations.menu.offset.top += calculations.context.scrollTop;
             }
-            if(module.has.subMenu()) {
+            if(module.has.subMenu($currentMenu)) {
               calculations.menu.height += $currentMenu.find(selector.menu).first().outerHeight();
             }
             onScreen = {
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3294,6 +3294,9 @@ $.fn.dropdown = function(parameters) {
           menu: function() {
             return ($menu.length > 0);
           },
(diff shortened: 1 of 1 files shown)


ko2in (commented):
Please review this one. I chose the selector incorrectly in previous commit.

src/definitions/modules/dropdown.js
```diff
@@ -3294,6 +3294,9 @@ $.fn.dropdown = function(parameters) {
           menu: function() {
             return ($menu.length > 0);
           },
+          subMenu: function() {
+            return $menu.find(selector.menu).length > 0;
```
lubber-de, line 3298: This has to respect a multilevel submenu
```suggestion
          subMenu: function($subMenu) {
            return ($subMenu || $menu).find(selector.menu).length > 0;
```

src/definitions/modules/dropdown.js
```diff
@@ -3523,6 +3526,9 @@ $.fn.dropdown = function(parameters) {
             if(module.is.verticallyScrollableContext()) {
               calculations.menu.offset.top += calculations.context.scrollTop;
             }
+            if(module.has.subMenu()) {
```
lubber-de, line 3529: Submenu has alto to respect multi nested submenus
```suggestion
            if(module.has.subMenu($currentMenu)) {
```

lubber-de (changes requested):
When this code is used on a multilevel submenu the opening of the 2nd level submenu will return NaN for outerHeight on undefined, because it will always check the most parent menu. Supplying the currentmenu to the `.has.subMenu` method to support nested dropdowns fixes that . There won't be a console error though, but the resulting calculations.height is NaN in those cases.

I suggested some changes at the related codelines to fix it.

Try this dropdown to reproduce
```javascript
<div class="ui dropdown button">
  <span class="text">Choose Category</span>
  <i class="dropdown icon"></i>
  <div class="menu">
    <div class="item">
      <span class="text">Category 1</span>
    </div>
    <div class="item">
      <i class="dropdown icon"></i>
      <span class="text">Category 2</span>
      <div class="menu">
        <div class="item">Item 2A</div>
        <div class="item">Item 2B</div>
        <div class="item">Item 2C</div>
      </div>
    </div>
    <div class="item">
      <i class="dropdown icon"></i>
      <span class="text">Category 3</span>
      <div class="menu">
        <div class="item">Item 3A</div>
        <div class="item">Item 3B</div>
        <div class="item">Item 3C</div>
      </div>
    </div>
  </div>
</div>
```

ko2in (commented):
@lubber-de Please review again. I used the argument name `subMenu: function($currentMenu)` because, it refers to the current menu.

Shouldn't the `openDownward` check only for first-level menu? I thought only the first level menu opens upward or downward, but from second level, all menus open leftward or rightward.

lubber-de:
> I used the argument name `subMenu: function($currentMenu)` because, it refers to the current menu.
> 
While this is fine as long as this method is only used internally, i still suggest to make the menu argument optional, so it can be used as behavior or in other parts of the code.

```javascript
return ($currentMenu || $menu).find(selector.menu).length > 0;
```
making `$currentMenu` optional will make this also work as externally called behavior. All other `.has` methods also work this way (yes, those always refer to the most parent menu anyway)


> Shouldn't the `openDownward` check only for first-level menu? I thought only the first level menu opens upward or downward, but from second level, all menus open leftward or rightward.

That's logical, yes. But nevertheless the fix was needed anyway (`has.submenu` was checking for $menu always, while the inner `if` was referencing `$currentMenu` instead)

ko2in (commented):
OK. Here we go again. The menu selector is optional now. If it doesn't pass specific menu node, it'll use the root menu node instead.

lubber-de (approved):
LGTM  👍

exoego (approved):
LGTM
Document needed ?

ko2in:
No. The fix and changes are internal use only.

PR #1782, opened 2020-11-25
fix(dropdown): Clearing multiple selection dropdown supports to prevent onChange callback

### LINKED ISSUE #1781: [Dropdown] Clearing a multiple dropdown ignores preventChangeTrigger setting
# Bug Report

When using the behavior `$('#blah').dropdown('clear',true)` on a dropdown, it's supposed to not fire the dropdown's `onChange` function. However if the dropdown happens to be a multiple dropdown, the `onChange` is instead fired once for every selected option.

## Steps to reproduce
1. Make a multiple select dropdown
2. Init the dropdown, adding an `onChange` function with a console.log or whatever
3. Clear the dropdown with the "preventChangeTrigger" flag set to true (via `$('#blah').dropdown('clear',true)`)

## Expected result
The `onChange` should not be triggered.

## Actual result
The `onChange` is triggered once for every selected element, annihilating the world as we know it.

## Testcase
https://jsfiddle.net/designosis/ob4wLf6h/

## Screenshot
Screen Shot 2020-11-25 at 1 54 51 PM

## Version
2.8.7

## Note
This is perhaps related to but not the same as https://github.com/fomantic/Fomantic-UI/issues/165

ko2in:
## Description
When clearing the selected items from multiple dropdown via `clear` method, the `onChange` callback triggers although the `preventChange` is set as true.

It's because, when the dropdown is multiple selection, the `clear` calls the specific internal method to remove multiple selected items without forwarding the `preventChange` setting which results triggering the `onChange` callback.

This PR makes the `clear` method to call the internal method by passing the `preventChange` setting which supports to prevent triggering the `onChange` callback.

## Testcase
**Bug:** https://jsfiddle.net/ko2in/vadg8nr2/1/

**Fix:** https://jsfiddle.net/ko2in/vadg8nr2/2/

## Screenshot
**Bug:**

Dropdown-Bug

**Fix:**

Dropdown-Fix

## Closes
#1781

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2368,7 +2368,7 @@ $.fn.dropdown = function(parameters) {
 
         clear: function(preventChangeTrigger) {
           if(module.is.multiple() && settings.useLabels) {
-            module.remove.labels();
+            module.remove.labels($module.find(selector.label), preventChangeTrigger);
           }
           else {
             module.remove.activeItem();
@@ -3094,7 +3094,7 @@ $.fn.dropdown = function(parameters) {
           userAddition: function() {
             $item.filter(selector.addition).remove();
           },
-          selected: function(value, $selectedItem) {
+          selected: function(value, $selectedItem, preventChangeTrigger) {
             $selectedItem = (settings.allowAdditions)
               ? $selectedItem || module.get.itemWithAdditions(value)
               : $selectedItem || module.get.item(value)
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -3210,12 +3210,12 @@ $.fn.dropdown = function(parameters) {
...
                 module.remove.message();
                 if(isUserValue) {
-                  module.remove.value(stringValue);
+                  module.remove.value(stringValue, stringValue, module.get.item(stringValue), preventChangeTrigger);
                   module.remove.label(stringValue);
                 }
                 else {
                   // selected will also remove label
-                  module.remove.selected(stringValue);
+                  module.remove.selected(stringValue, module.get.item(stringValue), preventChangeTrigger);
```
lubber-de, line 3218: Inside the `.remove.selected` function the auto generation of `$selectedItem` is dependent on the `allowAddition` setting. by providing `module.get.item(stringValue)` here directly, this would be ignored and probably won't if `allowAdditions` is true.

If, instead of creating the value by yourself, just provide `false`, this would still work as before (and you dont need to repeat the logic here again) but the additional `preventChangeTrigger` can be provided as awell

```suggestion
                  module.remove.selected(stringValue, false, preventChangeTrigger);
```

lubber-de (changes requested):
Please see my comment

src/definitions/modules/dropdown.js
ko2in: Thanks. I updated the PR, please review again.

ko2in (commented):
Just pass the `false` to avoid duplicate look up when `allowAdditions` is used.

lubber-de (approved):
LGTM

exoego (approved):
LGTM

PR #1784, opened 2020-12-02
[Button] labeled icon buttons have wrong margin

### LINKED ISSUE #1783: [Button] Margins are too large when button is multiple with a placeholder.
# Bug Report
When using a dropdown with multiple selection, the margin is a little too big. It's not when removing the multiple part.

## Steps to reproduce

cf. jsfiddle

## Expected result
button shouldn't be that large in multiple like the no multiple dropdown

## Actual result
text is wrapped

## Testcase
This is the test case for latest version of fomantic (2.8.7 at the time) https://jsfiddle.net/jLn0wkv9/1/

## Screenshot (if possible)
2.8.7 (multiple is on right)
image

## Version

2.8.7

Problem related:
https://github.com/fomantic/Fomantic-UI/pull/1571

lubber-de:
## Description
A `labeled icon button` got a wrong margin especially when used for multiple dropdown buttons as the specificity changed in  #1571 

However `labeled icon buttons` are never supposed to be displayed without text so the selector for single icon buttons should not select labeled buttons.

## Testcase
- Adjust the selector via console accordingly
https://jsfiddle.net/jLn0wkv9/1/

## Screenshot
labelediconbutton

## Closes
#1783

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -735,7 +735,7 @@
 ---------------*/
 
 .ui.icon.buttons .button,
-.ui.icon.button:not(.animated):not(.compact) {
+.ui.icon.button:not(.animated):not(.compact):not(.labeled) {
   padding: @verticalPadding @verticalPadding ( @verticalPadding + @shadowOffset );
 }
 .ui.animated.icon.button > .content > .icon,


ko2in (approved):
LGTM.

exoego (approved):
LGTM

PR #1795, opened 2020-12-16
[Nag] Refactored: no jquery cookie plugin dependency, colors, sizes, fixes

lubber-de:
## Description

This PR finishes the `nag` module. It was hidden from the docs, because there were still bugs inside and not completely done (but already used in several projects). I refactored some code and added usual features so this module can be officially supported again (I'll do another PR for the docs then) 

### What has been improved?

- Removed the dependency of the jquery cookie plugin. Cookie storage method is now natively supported inside the module. Now every storage method also got the same api for easier abstraction
- Added expiration functionality for localstorage method as well 
- Added color support
- Added sizing support
- fixed grouping (wrong border radius)
- fixed animation (wrong direction)
- slightly fixed text colors for better contrast
- fully backward compatible to existing code

## Testcase
https://jsfiddle.net/lubber/p67wLyb0/41/

## Screenshot
image
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/pull/2905
https://github.com/Semantic-Org/Semantic-UI/pull/6505
https://github.com/Semantic-Org/Semantic-UI/issues/211#issuecomment-227877711
https://github.com/Semantic-Org/Semantic-UI/issues/211#issuecomment-140584891

### DIFF
--- a/src/definitions/modules/nag.js
+++ b/src/definitions/modules/nag.js
@@ -58,14 +58,14 @@ $.fn.nag = function(parameters) {
 
         element         = this,
         instance        = $module.data(moduleNamespace),
-
+        storage,
         module
       ;
       module = {
 
         initialize: function() {
           module.verbose('Initializing element');
-
+          storage = module.get.storage();
           $module
             .on('click' + eventNamespace, selector.close, module.dismiss)
             .data(moduleNamespace, module)
@@ -110,9 +110,9 @@ $.fn.nag = function(parameters) {
 
         hide: function() {
           module.debug('Showing nag', settings.animation.hide);
-          if(settings.animation.show == 'fade') {
+          if(settings.animation.hide == 'fade') {
             $module
-              .fadeIn(settings.duration, settings.easing)
+              .fadeOut(settings.duration, settings.easing)
             ;
           }
           else {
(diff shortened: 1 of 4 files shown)


src/definitions/modules/nag.less
```diff
@@ -95,64 +95,116 @@ a.ui.nag {
...
-  border-radius: @bottomBorderRadius;
-  top: auto;
-  bottom: @bottom;
+  .ui.bottom.nags,
+  .ui.bottom.nag {
+    border-radius: @bottomBorderRadius;
+    top: auto;
+    bottom: @bottom;
+  }
 }
```
ko2in, line 128: Can we have all rounded borders if it's not fixed nag? Otherwise, it looks awkward as it's bottom borders are rounded, but the top borders are not.

lubber-de: Yes, right 👍 
Did that by e1a940b97c2a26d64ff5d9fba340982ba5a58e62 and also supported fixed groups as well 🙂 
Updated screenshots and jsfiddle in de PR description.

Please review again

src/definitions/modules/nag.js
```diff
@@ -181,39 +246,30 @@ $.fn.nag = function(parameters) {
...
-            if(settings.storageMethod == 'localstorage' && window.localStorage !== undefined) {
-              window.localStorage.setItem(key, value);
-              module.debug('Value stored using local storage', key, value);
-            }
-            else if(settings.storageMethod == 'sessionstorage' && window.sessionStorage !== undefined) {
-              window.sessionStorage.setItem(key, value);
-              module.debug('Value stored using session storage', key, value);
+            if(storage === window.localStorage && options.expires) {
+              module.debug('Storing expiration value in localStorage', key, options.expires);
+              storage.setItem(key + settings.expirationKey, options.expires );
```
exoego, line 251: [Cookie also supports expiration](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Example_3_Do_something_only_once) in the form of `; expires=date-in-GMTString-format`. 
How about supporting expiration on cookie storage ?

lubber-de: Cookie expiration is already built in via the given cookie options.

I only added the extra expiration here for localstorage, because localstorage does not expire values by default

exoego (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/257

PR #1796, opened 2020-12-17
[Form]: Support range values for decimal and number validation, add max/minValue and display range in default prompt

lubber-de:
## Description
It was not possible to allow for an optional range check for the decimal or number validator. This was only supported for integer values.
This PR now 

- supports decimal and number validators to supply a range by making the check for a range value a separate `range` validator so it can be reused.
- adds 2 new validators `maxValue` and `minValue` accordingly
- displays the range values in the default error prompt as well if no prompt is given 

### Before
```javascript
rules: [
  {
    type   : 'decimal'
  }
]
```

### After
```javascript
rules: [
  {
    type   : 'decimal[0.5..1.9]'
  }
]
rules: [
  {
    type   : 'minValue[50]'
  }
]
rules: [
  {
    type   : 'maxValue[10]'
  }
]
```

## Testcase
https://jsfiddle.net/lubber/g3nte1m6/14/

### Before
image

### After
image
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5496

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -559,8 +559,8 @@ $.fn.form = function(parameters) {
               parts = ancillary.split('..', 2);
               if(!rule.prompt) {
                 suffixPrompt = (
-                    parts[0] === '' ? settings.prompt.maxValue.replace(/\{ruleValue\}/,'{max}') :
-                    parts[1] === '' ? settings.prompt.minValue.replace(/\{ruleValue\}/,'{min}') :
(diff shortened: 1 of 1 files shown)


exoego (approved):
Nice !!

prudho (approved):
... with a nice and efficient refactoring, LGTM !

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/250

PR #1797, opened 2020-12-17
[Input / Form] Firefox has different placeholder opacity by default

lubber-de:
## Description
[Firefox does not set the default placeholder to 100% opacity](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder#Opaque_text)

This results into a visual difference when an input field is shown next to a dropdown, whereas the placeholder opacity is 100%, while firefox defaults it around 54%

## Testcase
Open in Firefox to see the difference: The normal input fields placeholders have a lower opacity than the dropdown placeholder

### Before
https://jsfiddle.net/lubber/eh8v6y35/2/

### After
https://jsfiddle.net/lubber/eh8v6y35/4/

## Screenshots

|Before|After|
|-|-|
|image|image|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4937

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -377,6 +377,7 @@
 }
 .ui.form ::-moz-placeholder {
   color: @inputPlaceholderColor;
+  opacity: 1;
 }
 
 .ui.form :focus::-webkit-input-placeholder {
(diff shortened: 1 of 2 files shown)


exoego (approved):
Eagle eye !!

prudho (approved):
This one was annoying as hell, great fix @lubber-de !

LGTM 👍

PR #1800, opened 2020-12-17
[Message] centered and right aligned variant

lubber-de:
## Description
This PR adds a `centered` or `center aligned` as well as a `right aligned` variant to the `message` component just like many other components already support.

## Testcase
https://jsfiddle.net/lubber/03qx4gr5/39/

## Screenshot
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5049

### DIFF
--- a/src/definitions/collections/message.less
+++ b/src/definitions/collections/message.less
@@ -169,6 +169,12 @@
 /*******************************
             Variations
 *******************************/
+& when (@variationMessageCentered) {
+  .ui.centered.message,
+  .ui.center.aligned.message {
+    text-align: center;
+  }
+}
 
 & when (@variationMessageCompact) {
   /*--------------
--- a/src/definitions/collections/message.less
+++ b/src/definitions/collections/message.less
@@ -183,6 +183,11 @@
     }
   }
 }
+& when (@variationMessageRightAligned) {
+  .ui.right.aligned.message {
+    text-align: right;
+  }
+}
 
 & when (@variationMessageCompact) {
   /*--------------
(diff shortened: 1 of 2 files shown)


exoego (approved):
LGTM

prudho (commented):
Nice ! But the icon message seems a little away now.

Capture

Maybe we should align it with content ?

lubber-de:
> Maybe we should align it with content ?

@prudho Fixed by edbc0c4586b01795ed12dcce4d57457d5b8ba2ad 😉 
Updated jsfiddle and screenshot in issue description accordingly. Please review again

src/definitions/collections/message.less
```diff
@@ -173,6 +173,10 @@
   .ui.centered.message,
   .ui.center.aligned.message {
     text-align: center;
+    justify-content: center;
+    & > i.icon + .content {
+      flex: 0 0 auto;
+    }
```
ko2in: I think it's better to omit adjacent sibling selector for `content`. So, we can also place icon at the right side (https://jsfiddle.net/ko2in/8r9axfjo/2/). The icon left margin needs to be defined in this case. Maybe, another PR for the icon's margin adjustments.

lubber-de: @ko2in Thanks for the hint! I adjusted it according to your suggestion by 0bdd81fe4614ec38321ed69b2240b3677ac8a261
Please review again 🙂

lubber-de: And while the right icon also works now, i added a `right aligned` variant as well to complete it 🙂

prudho (approved):
Well, you're really on fire today 😄 

LGTM !

ko2in (approved):
LGTM :+1: My mailbox is flooding today about your PR notifications :smile:

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/252

PR #1807, opened 2020-12-21
[Dropdown] Clearable inline support

### LINKED ISSUE #924: [Dropdown] Inline clearable dropdown is not clearable
# Bug Report

## Steps to reproduce
1. Create inline dropdown
2. Initialize dropdown with `clearable: true` option
3. Can't clear selection

## Expected result
Should have a little "x" icon to clear

## Actual result
Does not

## Testcase
https://jsfiddle.net/tfmbgz3e/
(Taken directly from the docs [here](https://fomantic-ui.com/modules/dropdown.html#clearable))

## Screenshot (when possible)
image

## Version
2.7.7

lubber-de:
## Description
Inline (or no selection) dropdowns did not support `clearable` (while SUI does and even the FUI docs mention it)

## Testcase
Remove CSS to see the issues
https://jsfiddle.net/lubber/hemuypkc/8/

## Screenshot
|Before|After|
|-|-|
|image|image|

## Closes
#924 
https://github.com/Semantic-Org/Semantic-UI/issues/5800 (loading / min-width fix)

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -359,7 +359,7 @@ $.fn.dropdown = function(parameters) {
             if( !module.has.menu() ) {
               module.create.menu();
             }
-            if ( module.is.selection() && module.is.clearable() && !module.has.clearItem() ) {
+            if ( module.is.clearable() && !module.has.clearItem() ) {
               module.verbose('Adding clear icon');
               $clear = $('<i />')
                 .addClass('remove icon')
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -808,7 +808,7 @@ select.ui.dropdown {
     margin-left: @clearableIconMargin;
   }
   .ui.dropdown:not(.selection) > .remove.icon {
-    margin-top: -@clearableIconMargin;
+    margin-top: -@clearableIconMarginTop;
   }
 }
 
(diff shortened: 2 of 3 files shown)


ko2in:
The clear icon doesn't looks align very well on my screen as your screenshot.

inline-dropdown

`margin-top: -1.35em` looks the best for me. Can you also see how does it look for you with `-1.35em`?
inline-dropdown2

lubber-de:
🤔 Margin 1.35em looks worse for me in the jsfiddle (win 10/chrome/firefox)
See
image
I also added a usual `selection dropdown` to the jsfiddle to see the reason, why i put the remove icon more to the top with margin 1.5 (because it's already the case for usual selection dropdowns)
https://jsfiddle.net/lubber/hemuypkc/4/

lubber-de:
ok, i believe it's the same reason as the discussion you had with the labels.
I am also in 125% default zoom in Win10 when i browserzoom tp 110%, then -1,35em are indeed fine

100%
image

110%
image

I'll do some tests on browserstack and will change to 1.35 i all browsers there do it well

lubber-de:
@ko2in Ok, you were right. I tested several browsers using browserstack and indeed -1.35em is the best visual appearance. It's also ok to have it a bit different than for `selection dropdown` because  there is no border for non-selection dropdowns where the remove icon has to be centered.
I updated the jsfiddle and screenshots in the description accordingly
Please review again 🙂

ko2in (approved):
LGTM :tada:

exoego (approved):
LGTM

PR #1611, opened 2020-07-28
[Form]: Remove existing error messages on clear and reset and adjust error messages on revalidation when not inline

lubber-de:
## Description
Whenever a form was cleared or reset by their respective behaviors `clear` or `reset`, any existing error messages were not removed
Additionally , when form fields are revalidated after they have been invalidated the error message was still visible in the error messages. However this was not the case when the form had inline errors.

#### Broken
https://jsfiddle.net/lubber/70c8tapg/10/

#### Fixed
https://jsfiddle.net/lubber/70c8tapg/12/

## Screenshot
|Broken|Fixed|
|-|-|
|formvlearerror_wrong|formvlearerror_right|

## Closes
#1610 
https://github.com/Semantic-Org/Semantic-UI/issues/4363
https://github.com/Semantic-Org/Semantic-UI/issues/702

PR #1809, opened 2020-12-22
[Form] Simplify Blur validation and respect revalidate:false

lubber-de:
## Description
The validation triggered by change and blur are similar. However, the code was unneccesary different.
Additionally the blur validation did not respect a possible given `revalidate:false` resulting in revalidating error fields all the time when the field blurs (even if one only wants revalidation on submit)

It also fixes not showing/updating the error messages on blur as change does

## Testcase
- Click on "submit" to invalidate the fields
- Enter something in the first input field. It does not immediatly revalidate while typing, because `revalidate:false` prevents that (which is correct)
- Now blur the field by pressing tab or click somewhere outside the field

### Broken
The field gets revalidated
http://jsfiddle.net/lubber/g8n9c6kb/4/

### Fixed
The field does not get validated as expected
http://jsfiddle.net/lubber/g8n9c6kb/5/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/4889
https://github.com/Semantic-Org/Semantic-UI/issues/3574
https://github.com/Semantic-Org/Semantic-UI/issues/4475

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -444,15 +444,11 @@ $.fn.form = function(parameters) {
                 $fieldGroup     = $field.closest($group),
                 validationRules = module.get.validation($field)
               ;
-              if( $fieldGroup.hasClass(className.error) ) {
+              if(validationRules && (settings.on == 'blur' || ( $fieldGroup.hasClass(className.error) && settings.revalidate) )) {
                 module.debug('Revalidating field', $field, validationRules);
-                if(validationRules) {
-                  module.validate.field( validationRules );
-                }
-              }
-              else if(settings.on == 'blur') {
-                if(validationRules) {
-                  module.validate.field( validationRules );
+                module.validate.field( validationRules );
+                if(!settings.inline) {
+                  module.validate.form(false,true);
                 }
               }
             },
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -444,15 +444,11 @@ $.fn.form = function(parameters) {
...
                 module.debug('Revalidating field', $field, validationRules);
-                if(validationRules) {
-                  module.validate.field( validationRules );
-                }
-              }
-              else if(settings.on == 'blur') {
-                if(validationRules) {
-                  module.validate.field( validationRules );
+                module.validate.field( validationRules );
+                if(!settings.inline) {
```
exoego, line 450: Why this condition `!settings.inline` added ?

lubber-de: This is also used in the `change` variant. It's needed because of #1611 .
Without that, the error message is not getting updated after revalidation.
That's also the main reason why this issue fixes the other 2 SUI issues

lubber-de: And the removement of inline errors is already done inside the `validate.field` method, that's why the code first checks if the form should have inline errors

ko2in (approved):
LGTM.

exoego (approved):
LGTM

PR #1813, opened 2020-12-23
fix(build): upgrade css minify and remove specialcomments in concat build

lubber-de:
## Description
This PR upgrades gulp-clean-css (respectively clean-css) to the latest version.
Now the removement of special header comments in the concatenated minified css file `semantic.min.css` is properly done and works
Before, the minified concatenated css file still contained every special `/*! ... */` header from each single .css file which increased the minified css file significantly.

> #### Note for reviewers in case you wonder, why i removed nearly all previous options for the minify plugin
> Just because those previous values are default now in clean-css 4.2 (gulp-clean-css 4.3 depends in clean-css 4.2 ... don't get confused)🙂 

## Testcase
`npm install && npx gulp build`
The created `dist/semantic.min.css` does only contain one single comment header, but not every header from each concatenated single css file anymore

## Screenshot
|Before|After|
|-|-|
|image|image|

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6930

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1143,9 +1143,9 @@
       }
     },
     "clean-css": {
-      "version": "4.2.1",
-      "resolved": "https://registry.npmjs.org/clean-css/-/clean-css-4.2.1.tgz",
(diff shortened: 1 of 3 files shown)


ko2in (approved):
LGTM.

exoego (approved):
LGTM

PR #1814, opened 2020-12-23
[Form] Support shouldtrim option per fieldrule

### LINKED ISSUE #1278: Validation does not police leading and trailing spaces.
Originally identified by @Gwi in discord. 

Issue: Validation does not account for leading and trailing spaces. For example: 

is[dog]. '          dog', ' dog           ' are both valid but should not be.
isExactly[dog] '            dog', 'dog         ' are both valid but should not be.

Repro: Using the [docs demo](https://fomantic-ui.com/behaviors/form.html#specified-content) you can test this out.

The documentation infers otherwise. 'isExactly Dog' would imply, in standard English, that there would be no other characters than D-o-g.  However, in other use cases I could see the need to allow leading or trailing spaces.

Proposed solution:  Firstly I propose that validation should not change the input value unless the following classes are in place. I recommend a class be added to inputs, for example something like 'trimming-both', 'trimming-left', trimming-right' & 'trimming-none'. Fomantic validation would be modified so that the use of one of these classes would operate as inferred by the semantic name of the class. The trimming operation inferred by the class would happen __before__ the validation process was carried out.   This would retain the semantic intention of Fomantic without needing to have the developer revert to script to handle this chore.

lubber-de:
## Description
This PR adds support to use the form wide `shouldTrim` setting individually per field rule 

## Testcase
Enter "    dog" (leading spaces) into the field. Although the global `shouldTrim` setting is set to false, the field still validates, because `shouldTrim:true` is set to the fieldrule itself
https://jsfiddle.net/lubber/6349nfh0/1/

## Closes
#1278

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1327,7 +1327,7 @@ $.fn.form = function(parameters) {
                 // cast to string avoiding encoding special values
                 value = (value === undefined || value === '' || value === null)
                     ? ''
-                    : (settings.shouldTrim) ? String(value + '').trim() : String(value + '')
+                    : (settings.shouldTrim && rule.shouldTrim !== false) || rule.shouldTrim ? String(value + '').trim() : String(value + '')
                 ;
                 return ruleFunction.call(field, value, ancillary, $module);
               }


src/definitions/behaviors/form.js
```diff
@@ -1327,7 +1327,7 @@ $.fn.form = function(parameters) {
                 // cast to string avoiding encoding special values
                 value = (value === undefined || value === '' || value === null)
                     ? ''
-                    : (settings.shouldTrim) ? String(value + '').trim() : String(value + '')
+                    : (settings.shouldTrim && rule.shouldTrim !== false) || rule.shouldTrim ? String(value + '').trim() : String(value + '')
```
ko2in, line 1330: Isn't the same as `settings.shouldTrim || rule.shouldTrim !== false`?

lubber-de: No, there is one usecase there you may want to have a rule **not** trim althouh the overall form setting shouldTrim=true.
So if form:shouldTrim=true, your proposed solution will always be true, even if one explicisely defines: rule: shouldTrim:false

ko2in (approved):
LGTM.

exoego (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/250

PR #1815, opened 2020-12-25
feat(icon): use corner offset variables

lubber-de:
## Description
- Moved corner icon offsets into a variable, so one can easily adjust them
- Added render condition to possibly reduce CSS


exoego (approved):
LGTM

ko2in (approved):
LGTM.

PR #1816, opened 2020-12-26
[Modal] bodyfixed selector did not work with existing padding values

lubber-de:
## Description
When a modal opens and the screen ist large, so there exists a scrollbar, selected fixed elements on the screen are adjusted so they dont move while the screen is shown/closed.

- The `fixed nag` element was missing to be supported
- existing values for `padding` were not taken into account. For all previous selectors (toast-container, menu, sidebar) this was not a problem, because they haven't had a padding by default. But the new `nag` selector actually has a padding of 1em.
- In case a non fixed selector was used, the `padding` was the wrong approach (for example the `close` icon inside the `nag`)
- When the modal hides, the previously set values for padding were overridden by 0. Whenever there was a default padding before, this would not work anymore. I now completely removed the inline style instead.

## Testcase
Watch the right side of the bottom nag next to the scrollbar
The whole nag including the close icon moves to the right when the modal opens

### Broken
https://jsfiddle.net/lubber/43ndykwq/26/

### Fixed
https://jsfiddle.net/lubber/43ndykwq/28/

## Screenshot
|Broken|Fixed|
|-|-|
|nagmodalbroken|nagmodalfixed|

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -680,7 +680,12 @@ $.fn.modal = function(parameters) {
           bodyMargin: function() {
             var position = module.can.leftBodyScrollbar() ? 'left':'right';
             $body.css('margin-'+position, initialBodyMargin);
-            $body.find(selector.bodyFixed.replace('right',position)).css('padding-'+position, initialBodyMargin);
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

prudho (approved):
LGTM

PR #1823, opened 2021-01-03
fix(nag): missing callbacks and fixed unused code

lubber-de:
## Description
The `nag` module was still missing the usual callbacks `onshow`, `onvisible`, `onhide`, `onhidden` we have in other modules like modal , calendar or toast.
Even worse, existing code was not even used anywhere or non existing options (`speed` (exists, not used) -> `duration` (used, non existing)

All of this has now been corrected 😊 

## Testcase
Turn on developer console and watch the output while showin/hiding/dismissing the nag
https://jsfiddle.net/lubber/k7xodeqz/2/

### DIFF
--- a/src/definitions/modules/nag.js
+++ b/src/definitions/modules/nag.js
@@ -94,47 +94,47 @@ $.fn.nag = function(parameters) {
 
         show: function() {
           if( module.should.show() && !$module.is(':visible') ) {
+            if(settings.onShow.call(element) === false) {
+              module.debug('onShow callback returned false, cancelling nag animation');
+              return false;
+            }
             module.debug('Showing nag', settings.animation.show);
-            if(settings.animation.show == 'fade') {
+            if(settings.animation.show === 'fade') {
               $module
-                .fadeIn(settings.duration, settings.easing)
+                .fadeIn(settings.duration, settings.easing, settings.onVisible)
               ;
             }
             else {
               $module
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/257
If you find time, please review, because ;)

ko2in (approved):
LGTM.

src/definitions/modules/nag.js
```diff
@@ -94,47 +94,47 @@ $.fn.nag = function(parameters) {
...
-          if (settings.onHide) {
-            settings.onHide();
-          }
-        },
-
         dismiss: function(event) {
-          if(settings.storageMethod) {
+          if(module.hide() !== false && settings.storageMethod) {
+            module.debug('Dismissing nag', settings.storageMethod, settings.key, settings.value, settings.expires);
             module.storage.set(settings.key, settings.value);
```
exoego, line 136: Is it safe to skip `module.storage.gset` when `module.hide() === false` ?

exoego (approved):
Basically LGTM
I left a small comment just for clarification.

src/definitions/modules/nag.js
lubber-de: Yes, the only situation where the storage should set the dismisss value is of course when hiding the nag is allowed. (if i cant hide it, i also cannot dismiss it, most logical)

PR #1836, opened 2021-01-09
fix(toast): attached toast margin and border

lubber-de:
## Description
an `attached` positioned toast container still showed an 1px background on the right of the toast

## Testcase
https://jsfiddle.net/lubber/9vkoqp8m/21/

## Screenshot
Look closely to the very right of the toast. You can see a 1px background green color
|Before|After|
|-|-|
|image|image|


ko2in (approved):
LGTM.

exoego (approved):
LGTM

Issue #1846, opened 2021-01-20
[dropdown] onQuery callback


jamessampford:
Unless there is a way to listen to changes to when a query is typed/pasted into a dropdown search, possibly adding an onQuery callback to the dropdown could be a useful addition and could also be used as an alternative to using remoteData

lubber-de:
Implemented by #1853

PR #1853, opened 2021-01-24
feat(dropdown) adds onSearch callback for when searching in the dropdown

jamessampford:
## Description
This PR adds a new `onSearch` callback to allow access to the search term being entered so that external processes can be informed about the term being typed in. Such processes may include an API that requires authentication with a token, especially if the website in question already has an existing API

## Testcase
https://jsfiddle.net/oafqnm3v/

Relates to #1846

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -297,6 +297,12 @@ $.fn.dropdown = function(parameters) {
             : module.get.query()
           ;
           module.verbose('Searching for query', query);
+          if(settings.fireOnInit === false && module.is.initialLoad()) {
+            module.verbose('Skipping callback on initial load', settings.onSearch);
+          }
+          else {
+            settings.onSearch.call(element, query);
+          }
           if(module.has.minCharacters(query)) {
             module.filter(query);
           }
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -297,6 +297,12 @@ $.fn.dropdown = function(parameters) {
             : module.get.query()
           ;
           module.verbose('Searching for query', query);
+          if(settings.fireOnInit === false && module.is.initialLoad()) {
+            module.verbose('Skipping callback on initial load', settings.onSearch);
+          }
+          else {
+            settings.onSearch.call(element, query);
```
lubber-de, line 304: What about checking for a return value from that function to probably decline the search at all? We have this in other callbacks as well, so i suggest adding this to the minCharacters check one LOC later like this instead
```javascript
if(module.has.minCharacters(query) && settings.onSearch.call(element, query) !== false) {
```

jamessampford: Seems like a good idea to remove that else statement and put it one line later

lubber-de (approved):
LGTM

ko2in (approved):
LGTM.

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/263

Issue #1847, opened 2021-01-20
[dropdown] actionable item + callback


jamessampford:
Allow a dropdown to have an actionable item, this is, doesn't get chosen as the value - by this, imagine having a dropdown where at first there is no data, however you could append an actionable item in the list, which when selected could launch a model to create a new record and the developer could add it to the dropdown before/after the actionable item as the selected item

For example, the actionable item is called "+New"
`<div class="item actionable" data-value="new">+New</div>`
Clicking/Picking this item fires off a callback, possibly with the data-value and text of the item (dropdown value+text remains at existing value)
```
$('.ui.dropdown').dropdown({
    onActionable: function(value, text, $selected){
        ....
    }
});
```

PR #1854, opened 2021-01-24
feat(dropdown) onActionable callback

jamessampford:
### Description
The intention of this PR is to add a new onActionable callback to allow the possibility of providing a way of having a way of performing actions when an actionable item has been selected

The remaining items are outstanding as am unsure:
- [ ] prevent what seems to be the method being called twice
- [ ] ensure that the original selection is not overwritten with the actionable item (nothing should happen upon selection, aside from firing the callback)
- [ ] ensure that when using the keyboard, that the callback is not triggered unless hitting the enter key and do not use as the value of the dropdown
- [ ] possibly need an `actionableCollapse` field to prevent collapse on selection of an actionable item

### Testcase
https://jsfiddle.net/n1yvb07w/

The gives an example of dynamically adding additional options to the dropdown, but could also fire the additional item to an API to add to a database (validation could also happen on the server to ensure it is unique) and then refresh the list of options.

An alternative option could be that the dropdown could contain something like a navigation tree showing options:
https://jsfiddle.net/n1yvb07w/1/

Although this second example may not be so great, since the menu collapses when selecting the actionable item, so it could be that an additional option `actionableCollapse` may be wanted, where if `false` the menu doesn't collapse on selection

Related to #1847

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2677,7 +2677,11 @@ $.fn.dropdown = function(parameters) {
                 : value,
               newValue
             ;
-            if(hasInput) {
+            if($selected && $selected.hasClass(className.actionable)){
+              settings.onActionable.call(element, value, text, $selected);
+              return;
+            }
+            else if(hasInput) {
               if(!settings.allowReselection && stringValue == currentValue) {
                 module.verbose('Skipping value update already same value', value, currentValue);
                 if(!module.is.initialLoad()) {
@@ -4038,6 +4042,7 @@ $.fn.dropdown.settings = {
   onChange      : function(value, text, $selected){},
   onAdd         : function(value, text, $selected){},
   onRemove      : function(value, text, $selected){},
+  onActionable  : function(value, text, $selected){},
 
   onLabelSelect : function($selectedLabels){},
   onLabelCreate : function(value, text) { return $(this); },
@@ -4097,7 +4102,8 @@ $.fn.dropdown.settings = {
     icon         : 'icon',     // optional icon name
     iconClass    : 'iconClass', // optional individual class for icon (for example to use flag instead)
     class        : 'class',    // optional individual class for item/header
-    divider      : 'divider'   // optional divider append for group headers
+    divider      : 'divider',  // optional divider append for group headers
+    actionable   : 'actionable' // optional actionable item
   },
 
   keys : {
@@ -4167,7 +4173,8 @@ $.fn.dropdown.settings = {
     header      : 'header',
     divider     : 'divider',
     groupIcon   : '',
-    unfilterable : 'unfilterable'
+    unfilterable : 'unfilterable',
+    actionable : 'actionable'
   }
 
 };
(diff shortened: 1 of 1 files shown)


prudho (commented):
Took a look at it, and the idea does not seem bad.

But you will have to handle a lot of cases if you want to make it work. For example if your dropdown is a `multiple` it will not work anymore.

jamessampford:
I have updated the PR to take into account keyboard up/down arrow keys causing `onActionable` to trigger when just being selected - hitting enter on the keyboard will allow it to fire off

Have updated my sample JSFiddles:
https://jsfiddle.net/u18h5b3n/
https://jsfiddle.net/z4gj1wy8/

I think just need to add that additional `actionableCollapse` boolean setting to prevent the dropdown from collapsing upon selecting the `onActionable` item (the 2nd JSFiddle can show it is needed?) - suggest the default to be false (ie. do not collapse upon selection). Trying to find the best way to implement this, but may need help/suggestions

jamessampford:
> Took a look at it, and the idea does not seem bad.
> 
> But you will have to handle a lot of cases if you want to make it work. For example if your dropdown is a `multiple` it will not work anymore.

Good point about `multiple` - it could be that this explicitly doesn't support multiple, or we will need to bake support for that in too (somehow)

I think the whole concept around `onActionable` has a few useful applications, so would certainly be a good addition :)

lubber-de:
Is this PR ready for review now? You still have an unmarked task in your description. If it's still in progress, please convert this PR into Draft

jamessampford:
@lubber-de Should be complete now - updated the previous sample to by adding `collapseOnActionable` which seems to do the trick

https://jsfiddle.net/dq8pz39g/

src/definitions/modules/dropdown.js
```diff
@@ -2774,16 +2778,20 @@ $.fn.dropdown = function(parameters) {
                       module.save.remoteData(selectedText, selectedValue);
                     }
                     if(settings.useLabels) {
-                      module.add.label(selectedValue, selectedText, shouldAnimate);
                       module.add.value(selectedValue, selectedText, $selected);
```
lubber-de, line 2781: Could you please explain, why setting the value is still done even if the item is actionable, but setting is label/text will be prevented? Shouldn't even the value not being updated then?

src/definitions/modules/dropdown.js
```diff
@@ -2774,16 +2778,20 @@ $.fn.dropdown = function(parameters) {
...
+                        module.set.activeItem($selected);
+                      }
                       module.filterActive();
                       module.select.nextAvailable($selectedItem);
                     }
                     else {
                       module.add.value(selectedValue, selectedText, $selected);
-                      module.set.text(module.add.variables(message.count));
-                      module.set.activeItem($selected);
+                      if(!$selected.hasClass(className.actionable)){
```
lubber-de, line 2791: Same question here: value is still set, but text/activeItem is not. 🤔

src/definitions/modules/dropdown.js
```diff
@@ -2679,7 +2679,11 @@ $.fn.dropdown = function(parameters) {
                 : value,
               newValue
             ;
-            if(hasInput) {
+            if($selected && $selected.hasClass(className.actionable)){
```
lubber-de, line 2682: Here the setting of a value is skipped when it's actionable. This seems logical, but later in `set.selected` the value is always written and only label/text is skipped. (See other review comments)

lubber-de (commented):
Please explain the logic when differentiating between setting value/text/label. (See my comments)
Perhaps i just don't get the logic 😄 
Thanks

jamessampford:
@lubber-de I have taken the feedback onboard and realised it could be simplified and be less of a workaround - the latest commit will hopefully be the last, as the value will not be set at all and calls the additional function directly

ColinFrick:
@jamessampford After pushing changes, please update your jsfiddles to the latest version. Stumbled upon some bugs, that were already fixed.

lubber-de:
@ColinFrick I updated his jsfiddle testcases in the description using the latest commit

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/376

PR #1856, opened 2021-01-24
feat(icon): colored variant for bordered or circular

lubber-de:
## Description
This PR adds a `colored` variant to `circular` or `bordered` icon(s).

Currently the border of a circular or bordered icon is always grey. Even if the icon itself has got a color.
By adding a `colored` class to such icons the border will inherit its color from the icon itself.

## Testcase
https://jsfiddle.net/lubber/8xqfopah/3/

## Screenshot
image

PR #1864, opened 2021-02-06
fix(icon): colored selector was wrongly placed

lubber-de:
## Description
I accidently placed the `.colored` selector outside of the icon scope with #1856 which made the specificity useless. 🙄 
This PR is fixing this


exoego (approved):
Nice catch

ColinFrick (approved):

PR #1866, opened 2021-02-06
feat(dropdown): columns were ignored inside menus

lubber-de:
## Description
`column dropdown` was not working when used inside menus

## Testcase
The right "Leaderboard" Menu entry is supposed to be a `two column` dropdown
### Broken
- It's ignored and still displayed as one column
- https://jsfiddle.net/lubber/or6nd413/10/
### Fixed
- Two columns as expected
https://jsfiddle.net/lubber/or6nd413/9/
 
## Screenshot
|Before|After|
|-|-|
|image|image|


exoego (approved):
LGTm 👍

ColinFrick (approved):

PR #1896, opened 2021-02-22
fix(search): dont set value when onselect returns false

### LINKED ISSUE #1895: Input search onSelect not working prevent default action
# Bug Report

## Steps to reproduce
1. İnclude js and css
2. Set js code
3. Type something
4. Click result

## Expected result
When click element then prevent default action 

## Actual result
The default action continued to work 

## Testcase
http://jsfiddle.net/b17ys2pt/2/

## Version
2.8.4 and up

lubber-de:
## Description
As the docs tell, when the `onSelect` callback returns `false` there should be no selection and menu hiding happen.
But infact only the menu is not hidden, the value is still selected.

This is because of #83 where it was made possible to change the value inside the onSelect callback.
To still support this i moved the value setting back to the point after the callback but now the value is only overridden if the previous value is still unchanged (to recognize a possible manual change in the callback.

## Testcase
- Enter something in the searchfield to bring up the menulist
- Select something from the list

### Broken
Although the onSelect callback returns false, the value is still overridden, while the menu remains
http://jsfiddle.net/lubber/9z4j5wqo/9/

### Fixed
Value is not overridden, menu remains, just as the docs say (and SUI does)
http://jsfiddle.net/lubber/9z4j5wqo/8/

## Proof of previous logic
Callback does not return false, but changes value inside the callback. It will be kept as before :)
http://jsfiddle.net/lubber/9z4j5wqo/10/

### Broken
searchcallback_bad

### Fixed
searchcallback_good

## Closes
#1895

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -227,9 +227,7 @@ $.fn.search = function(parameters) {
                 results = module.get.results(),
                 result  = $result.data(metadata.result) || module.get.result(value, results)
               ;
-              if(value) {
-                module.set.value(value);
-              }
+              var oldValue = module.get.value();
               if( $.isFunction(settings.onSelect) ) {
                 if(settings.onSelect.call(element, result, results) === false) {
                   module.debug('Custom onSelect callback cancelled default select action');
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

ColinFrick (approved):

PR #1909, opened 2021-03-05
fix(header): icon headers should not have a left content padding

### LINKED ISSUE #1908: Header with icon has unnecessary padding-left
# Bug Report

A header with icon class produces an unnecessary padding-left (looks to be left over from the case where icon and text are on the same line)

This padding-left is only necessary on a `header` without the `icon` class.

Possible offending lines: https://github.com/fomantic/Fomantic-UI/blob/fcd221b001d45ec6bc820d5f43a3d8bcad377bf0/dist/components/header.css#L101-L106

## Steps to reproduce
```HTML
<h2 class="ui icon header">
  <i class="settings icon"></i>
  <div class="content">
    Account Settings
    <div class="sub header">Manage your account settings and set e-mail preferences.</div>
  </div>
</h2>
```

## Expected result
Should not have `padding-left` on the `content` div. 

## Actual result
`padding-left` exists, creates off centered header.

## Testcase

Bug present in the docs itself: https://fomantic-ui.com/elements/header.html#icon-headers

## Screenshot (if possible)
Screen Shot 2021-03-05 at 2 34 48 PM

## Version

2.8.7

lubber-de:
## Descripton
An `icon header` has got a left padding which is only supposed to be needed when it's a normal header having an icon next to the header text

## Testcase
https://jsfiddle.net/lubber/c0vapqbz/2/

## Screenshot
iconheaderpadding

## Closes
#1908

### DIFF
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -113,8 +113,8 @@
 }
 
 /* After Icon */
-.ui.header:not(.centered):not(.aligned) > .icons + .content,
-.ui.header:not(.centered):not(.aligned) > i.icon + .content {
+.ui.header:not(.icon):not(.centered):not(.aligned) > .icons + .content,
+.ui.header:not(.icon):not(.centered):not(.aligned) > i.icon + .content {
   padding-left: @iconMargin;
   display: table-cell;
   vertical-align: @contentIconAlignment;


exoego (approved):
LGTM

PR #1924, opened 2021-03-20
fix(segment): support placeholder segments for tab usage

### LINKED ISSUE #1920: [Segment] Placeholder segments do not play well with tabs
# Bug Report
* A placeholder tab segment does not hide when not active
* A placeholder tab segment does not have the correct `display` value when active

## Steps to reproduce

1. Create tab with segments
2. Add placeholder to one of the segments
3. Call js `tab` function

## Expected result
Placeholder tab should work like a regular segment

## Actual result
Placeholder tab does not work like a regular segment

## Testcase
https://jsfiddle.net/eo7sut0g/

image
image

## Version

2.8.7

lubber-de:
## Description
When a `placeholder segment` is used for tabs, it messes up the display.

## Testcase
Switch between the tabs, the placeholder tab should only be visible when tab 2 gets activated

### Before
Messed up, the placeholder is shown all the time and also totally misaligned
https://jsfiddle.net/eo7sut0g/

### After
All fine
https://jsfiddle.net/lubber/bf1t4nzq/12/

### Before
placeholdersegmenttabbroken

### After
placeholdersegmenttabfixed

## Closes
#1920

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -139,6 +139,12 @@
     background: @placeholderBackground;
     border-color: @placeholderBorderColor;
     box-shadow: @placeholderBoxShadow;
+    &.tab {
+      display: none;
+      &.active {
+        display: flex;
+      }
+    }
   }
 
   .ui.placeholder.segment .button,


exoego (approved):
LGTM 👍

PR #1950, opened 2021-04-18
fix(form): file upload fields trigger error using jquery > 3.4.0

lubber-de:
## Description
When a form contains a file upload field it is also bound to several form events which are not supposed to happen on those and do not work because flie upload fields are limited in access.

That said, when using jQuery > 3.4.0, it triggers a `Maximum callstack exceeded instead`

## Testcase
- Click on the "Upload image" Button and select something you like 
- Watch console

### Broken
"maximum callstack exceeded"
https://jsfiddle.net/vzctb5pu/2/

### Fixed
Nothing 🙂 
https://jsfiddle.net/lubber/7f269p84/8/

## Screenshot
|Broken|Fixed|
|-|-|
|formfilebroken|formfilefixed|

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1602,9 +1602,9 @@ $.fn.form.settings = {
   selector : {
     checkbox   : 'input[type="checkbox"], input[type="radio"]',
     clear      : '.clear',
-    field      : 'input:not(.search), textarea, select',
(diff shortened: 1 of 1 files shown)


exoego (approved):
Nice catch

Yahav:
That was fast, thanks.

PR #1962, opened 2021-05-09
feat(build): check node 13 and 14 for github actions

lubber-de:
## Decription
Increases node version up to 14 LTS

## Screenshot
image


exoego (approved):
Nit: 8, 10, 11 and 13 could be removed since those are no longer receive patches.

Otherwise LGTM.

lubber-de:
> Nit: 8, 10, 11 and 13 could be removed since those are no longer receive patches.
> 
> Otherwise LGTM.

Will do once node 15/16 will work.

PR #1965, opened 2021-05-15
fix(search): fetch "Cannot read property 'top' of undefined

### LINKED ISSUE #1964: [Search] Uncaught TypeError: Cannot read property 'top' of undefined when ".message" div shows
# Bug Report
Try a local search or API search on the search bar. Make the "No Results" div display then hit the arrow down button, this <Uncaught TypeError: Cannot read property 'top' of undefined> error will appear.

## Expected result
Should it be doing nothing when we hit the arrow down button? Because no results found.

## Testcase
Try on any Fomantic Search Bar.

## Screenshot (if possible)
issue-1
"Hit the arrow down key now. ERROR will be shown!"
issue-2
issue-3

## Version
JQuery - 3.5.1
FUI - 2.8.6

lubber-de:
## Description
When a search has no result and one presses the "arrow down" key on the keyboard, a javascript error
> Cannot read property 'top' of undefined

occurs

This PR also fixes dead, unused and unnecessary code of the affected function "ensureVisible"

### Broken
https://jsfiddle.net/vphxjf61

### Fixed
https://jsfiddle.net/vphxjf61/1/

## Closes
#1964

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -252,16 +252,16 @@ $.fn.search = function(parameters) {
             }
           }
         },
-        ensureVisible: function ensureVisible($el) {
+        ensureVisible: function($el) {
           var elTop, elBottom, resultsScrollTop, resultsHeight;
-
+          if($el.length === 0) {
+            return;
+          }
           elTop = $el.position().top;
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

PR #1966, opened 2021-05-17
feat(dropdown,popup): support individual show and hide transition and duration

lubber-de:
## Description
This PR adds optional support for individual setting of transition method and duration to the dropdown and popup module.

We already have the same optional setting in toast, modal and dimmer, so this PR concludes a unique syntax.
Because the settings are optional, this change is 100% backwards compatible. 

```javascript
// duration is used as before from settings.duration
transition: {
  showMethod: 'zoom',
  hideMethod: 'fly'
}

// complete individual settings
transition: {
  showMethod: 'zoom',
  showDuration: 1000,
  hideMethod: 'fly',
  hideDuration: 500
}

```

## Testcase
https://jsfiddle.net/lubber/4o6b8Lqp/37/

## Screenshot
individualdropdowntransition

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/2911

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1869,7 +1869,7 @@ $.fn.dropdown = function(parameters) {
             return count;
           },
           transition: function($subMenu) {
-            return (settings.transition == 'auto')
+            return (settings.transition === 'auto')
               ? module.is.upward($subMenu)
                 ? 'slide up'
                 : 'slide down'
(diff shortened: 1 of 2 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -3684,7 +3684,7 @@ $.fn.dropdown = function(parameters) {
                 $currentMenu
                   .transition({
                     animation  : transition + ' out',
-                    duration   : settings.duration,
+                    duration   : settings.transition.hideDuration || settings.duration,
```
exoego, line 3687: I think we should document `settings.transition` has higher precedence than `settings.duration`.
Perhaps deprecating `settings.duration` ?

exoego (approved):
LGTM

src/definitions/modules/dropdown.js
lubber-de: Yes, will update the docs accordingly 👍🏼

ColinFrick (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/280

PR #1974, opened 2021-05-30
feat(dimmer): support individual show/hide duration

lubber-de:
## Description
This PR adds the same syntax for the optional `hideDuration/showDuration` value of the `transition` setting to the `dimmer` module as we already have in toast, modal, popup and dropdown.

As dimmer was the only module which was already/also supporting an object for the duration value, i kept the logic in for backward compatibility

## Testcase
https://jsfiddle.net/lubber/x2dzht4b/6/

### DIFF
--- a/src/definitions/modules/dimmer.js
+++ b/src/definitions/modules/dimmer.js
@@ -335,15 +335,12 @@ $.fn.dimmer = function(parameters) {
             return $dimmer;
           },
           duration: function() {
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/280

prudho (approved):
LGTM

PR #1985, opened 2021-06-16
build(deps): upgrade gulp-less to fully support less 4

lubber-de:
## Description
Upgrade to gulp-less 5 which finally removed the unmaintained accord dependency and now supports LESS 4


exoego (approved):
Finally:+1:

PR #1819, opened 2020-12-26
[Chore] Added parenthesis to math divisions to support LESS 4

lubber-de:
## Descriptions
LESS 4 enforces parenthesis to divider math operations. This was missing at a few places so Fomantic did not build with LESS 4

## Testcase
A codesnippet from FUI
The div should have a red background with a visible top margin.

- using LESS 3.13.1. All fine
https://jsfiddle.net/lubber/8mqpo3ne/2/

- using LESS 4.0.0. Breaks. No color, no padding
https://jsfiddle.net/lubber/8mqpo3ne/8/

- using LESS 4.0.0 and added parenthesis. Works again
https://jsfiddle.net/lubber/8mqpo3ne/10/

## Closes
#1818

PR #1986, opened 2021-06-16
fix(slider): less 4 support needs parenthesis

lubber-de:
## Description

LESS 4 forces parenthesis for math division. There was one line missing to take care of that. Everything else was covered by #1819  already


exoego (approved):
LGTM

PR #1987, opened 2021-06-17
[slider] Add option to disable touch-/mouse-sliding on the track

mhthies:
## Description
Currently, the slider is quite sensible to mouse and touch events: Any mouse drag and touch event that *starts* anywhere on the slider will move the knob (not only when hitting the knob). This is especially annoying when trying to scroll a page with many sliders on a mobile device with touch screen. You'll find yourself moving a slider accidentially quite often.

Thus, this PR introduces a setting `slideEverywhere` to sliders to change the click and touch behaviour. By default, it is `true`, which corresponds to the old behaviour. With `slideEverywhere=true`, the slider can only be moved by either dragging **the knob** or by clicking / tapping (not dragging) anywhere on the slider. 

## Testcase
https://jsfiddle.net/81cs7dbg/1/

## Screenshot (if possible)
Does not change anything visually.

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -283,7 +283,7 @@ $.fn.slider = function(parameters) {
               event.preventDefault();
               module.event.down(event);
             });
-            if(!slideEverywhere) {
+            if(!settings.slideEverywhere) {
               $module.on('touchstart' + eventNamespace, module.event.down);
             }
           },
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -786,7 +786,11 @@ $.fn.slider = function(parameters) {
             return value;
           },
           eventPos: function(event) {
-            if(event.type !== 'click') {
+            // unwrap jQuery event objects
+            if(!(event.pageX || event.changedTouches)) {
+              event = event.originalEvent;
+            }
+            if(event.changedTouches) {
               var
                 touchEvent = event.changedTouches ? event : event.originalEvent,
                 touches = touchEvent.changedTouches[0] ? touchEvent.changedTouches : touchEvent.touches,
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -284,6 +284,13 @@ $.fn.slider = function(parameters) {
               module.event.down(event);
             });
             if(!settings.slideEverywhere) {
+              $module.find('.track, .inner').on('click' + eventNamespace, function(event) {
+                event.stopImmediatePropagation();
+                event.preventDefault();
+                module.event.up(event);
+              });
+              $module.on('click' + eventNamespace, module.event.up);
+            } else {
               $module.on('touchstart' + eventNamespace, module.event.down);
             }
           },
@@ -779,7 +786,7 @@ $.fn.slider = function(parameters) {
             return value;
           },
           eventPos: function(event) {
-            if(module.is.touch()) {
+            if(event.type !== 'click') {
               var
                 touchEvent = event.changedTouches ? event : event.originalEvent,
                 touches = touchEvent.changedTouches[0] ? touchEvent.changedTouches : touchEvent.touches,
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -252,12 +252,22 @@ $.fn.slider = function(parameters) {
           },
           mouseEvents: function() {
             module.verbose('Binding mouse events');
-            $module.find('.track, .thumb, .inner').on('mousedown' + eventNamespace, function(event) {
+            var downElements = settings.slideEverywhere ? '.thumb,.inner,.track' : '.thumb';
+            $module.find(downElements).on('mousedown' + eventNamespace, function(event) {
               event.stopImmediatePropagation();
               event.preventDefault();
               module.event.down(event);
             });
-            $module.on('mousedown' + eventNamespace, module.event.down);
+            if(!settings.slideEverywhere) {
+              $module.find('.track, .inner').on('click' + eventNamespace, function(event) {
+                event.stopImmediatePropagation();
+                event.preventDefault();
+                module.event.up(event);
+              });
+              $module.on('click' + eventNamespace, module.event.up);
+            } else {
+              $module.on('mousedown' + eventNamespace, module.event.down);
+            }
             $module.on('mouseenter' + eventNamespace, function(event) {
               isHover = true;
             });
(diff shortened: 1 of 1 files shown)


src/definitions/modules/slider.js
```diff
@@ -267,12 +277,15 @@ $.fn.slider = function(parameters) {
...
+            var downElements = settings.slideEverywhere ? '.thumb,.inner,.track' : '.thumb';
+            $module.find(downElements).on('touchstart' + eventNamespace, function(event) {
               event.stopImmediatePropagation();
               event.preventDefault();
               module.event.down(event);
             });
-            $module.on('touchstart' + eventNamespace, module.event.down);
+            if(!settings.slideEverywhere) {
+              $module.on('touchstart' + eventNamespace, module.event.down);
+            }
```
lubber-de, line 288: This should be the same logic as you did for mouse events, shouldn't it?

lubber-de (changes requested):
Here is a testcase using your latest code
Try this on real mobile devices (i did on my Android Xperia 5ii)
https://fomantic-ui.com/jsfiddle/#!lubber/thr6dvLp/1/

The current code does not seem to work on real mobile devices at all. No matter if i touch or swipe on the track: It is always doing the same for me (as before: changing the slider immediatly)

It is also not completely working as expected when using on desktop browsers:
https://jsfiddle.net/lubber/thr6dvLp/1/
The mouse event does indeed not change the slider on mousedown and drag... but when i release the mouse button after the drag, the slider thumb still get repositioned to where i released the mouse button. I think your intention was to not change the thumb at all then.

mhthies:
> The current code does not seem to work on real mobile devices at all. No matter if i touch or swipe on the track: It is always doing the same for me (as before: changing the slider immediatly)

Oh, yes, you are right. This is due to the logics screwup in line 286 that you already commented. I actually used the new setting the wrong way round for touch events on the `$module`. This is easy to fix.

However, the `click` events are still not triggered on touch devices. I'll try to find a solution for this.

> It is also not completely working as expected when using on desktop browsers:
> https://jsfiddle.net/lubber/thr6dvLp/1/
> The mouse event does indeed not change the slider on mousedown and drag... but when i release the mouse button after the drag, the slider thumb still gets repositioned to where i released the mouse button. I think your intention was to not change the thumb at all then.

This is due to the definition of the ´click` event. From MDN:

> An element receives a click event when a pointing device button (such as a mouse's primary mouse button) is both pressed and released while the pointer is located inside the element.

The only way to prevent this, would be implementing a custom "click detection" which uses move events to cancel a click when moving more than a fixed pixel delta from the original click position. I could do this, but I don't think it's worth the additional required code.

src/definitions/modules/slider.js
mhthies: Fixed in b48c4f69718c8f777df8f4a36dc89a9b211fc3a9.

I have mistakenly assumed that both functions (`touchEvents` and `mouseEvents`) are called on touch devices. In addition, I got the logic meaning of `settings.slideEverywhere` wrong in that if statement.

lubber-de (changes requested):
Here is a jsfiddle with your latest code.
Unfortunatley it is completely broken on desktop for when slideEverywhere is true (which is default and would break every existing code)
https://jsfiddle.net/lubber/thr6dvLp/2/
The first example does not work at all anymore on desktop and throws a console error when you click/hover on the slider

image

On mobile i can still drag from inside the track in the second example.
https://fomantic-ui.com/jsfiddle/#!lubber/thr6dvLp/2/

mhthies:
Ah, sh*t. I'm sorry for wasting your time with testing my broken patches.
The exceptions on non-touch devices should be fixed now.

Unfortunately, I cannot reproduce the dragging in the second example. I tested the latest version locally and the previous version with you JS fiddle link on Chrome mobile 91.0.4472.101 on Android, Firefox mobile 89.1.1 on Android and Safari mobile on iOS. On all three browsers the behaviour is as intended.

src/definitions/modules/slider.js
```diff
@@ -267,12 +277,22 @@ $.fn.slider = function(parameters) {
...
-            $module.find('.track, .thumb, .inner').on('touchstart' + eventNamespace, function(event) {
+            var downElements = settings.slideEverywhere ? '.thumb,.inner,.track' : '.thumb';
+            $module.find(downElements).on('touchstart' + eventNamespace, function(event) {
               event.stopImmediatePropagation();
               event.preventDefault();
               module.event.down(event);
             });
-            $module.on('touchstart' + eventNamespace, module.event.down);
+            if(!settings.slideEverywhere) {
+              $module.find('.track, .inner').on('click' + eventNamespace, function(event) {
```
lubber-de, line 287: Shouldn't the event be "touchstart" instead of "click" inside the touchevents?

lubber-de (changes requested):
New jsfiddle using your latest commit:
https://jsfiddle.net/lubber/thr6dvLp/3/
https://fomantic-ui.com/jsfiddle/#!lubber/thr6dvLp/3/

Please see my question about the different event name

Also, the setting still does not work on desktop as mentioned. When releasing the mousebutton the thumb still gets repositioned even if dragged from within the track. @fomantic/maintainers Your opinion?

src/definitions/modules/slider.js
mhthies: No, "touchstart" is (somehow) equivalent to "mousedown" for touch events.

However, all modern browsers have a built-in mouse emulation, such that they trigger touch and mouse events when a touch screen is used. See this MDN page for more detail on the triggered events: https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent 

Here, we only want to trigger the `up` function when a user taps the slider's track, i.e. they touch and release but do not drag on the track. This is (more or less – see the other discussion) what the browser checks for to trigger the emulated `click` event.

mhthies:
> Also, the setting still does not work on desktop as mentioned. When releasing the mousebutton the thumb still gets repositioned even if dragged from within the track.

As I explained above:

> This is due to the definition of the click event. […]
> 
> The only way to prevent this, would be implementing a custom "click detection" which uses move events to cancel a click when moving more than a fixed pixel delta from the original click position. I could do this, but I don't think it's worth the additional required code.

lubber-de:
Yes, that's why i wanted the other maintainers to have an opinion if/how we deal with that. I otherwise smell people complaining the feature does not work on desktop...just saying 😉 
I still think there is a possibility  to fix/fetch this without implementing a complicated pixel detection mechanism. For example remember the DOM node on mousedown (was is "thumb" or something else?) and then use that infomation to deny the thumb move on mouse up again.

mhthies:
Sorry, I misinterpreted that as a criticism on my fixes.

Another compromise could be to restrict the new setting on touch events. I.e. with `slideEverywhere == false`, mouse interaction would stay the same as before but touch events would not allow dragging on the track. However, the trivial approach to do this would interfere with the solutions for #1988 that I have currently in mind.

prudho:
Mmmmmh... Can't we determine if pointer get out of the component zone, and then cancel the movement ? This seem to be a delicate issue tbh.

lubber-de:
Since #2327 is merged, i think we can close this?

lubber-de:
I made a simple replacement PR #2330 based on the current code. Please check

PR #1992, opened 2021-06-22
fix(modal): dont stack same actions buttons

lubber-de:
## Description

While preparing the docs for #1774 , i recognized that modal

- does not create action buttons at all when the actions container is not available in the DOM
- does always stack action buttons on DOM modals when reused

#### Missing action div container
https://jsfiddle.net/lubber/5f6rm830/2/
#### Actions stacked on each call
https://jsfiddle.net/lubber/5f6rm830/3/

#### Actions do not stack on each call
https://jsfiddle.net/lubber/5f6rm830/6/
#### Missing action div container will be created at runtime
https://jsfiddle.net/lubber/5f6rm830/5/

## Screenshot
stackedactionsbug

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -110,6 +110,10 @@ $.fn.modal = function(parameters) {
           }
           if(module.has.configActions()){
             var $actions = $module.find(selector.actions).addClass(settings.classActions);
+            if ($actions.length === 0) {
+              $actions = $('<div/>', {class: className.actions + ' ' + (settings.classActions || '')}).appendTo($module);
+            }
+            $actions.empty();
             settings.actions.forEach(function (el) {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/modal.js
```diff
@@ -110,6 +110,10 @@ $.fn.modal = function(parameters) {
           }
           if(module.has.configActions()){
             var $actions = $module.find(selector.actions).addClass(settings.classActions);
+            if ($actions.length === 0) {
+              $actions = $('<div/>', {class: className.actions + ' ' + (settings.classActions || '')}).appendTo($module);
+            }
+            $actions.empty();
```
exoego, line 116: Nit) I think empty can be skipped when the action container is created dynamically.

```suggestion
            } else {
              $actions.empty();
            }
```

exoego (approved):
LGTM
Left a minor suggestion

prudho (approved):
LGTM

PR #1998, opened 2021-06-26
fix(nightly): check for already uploaded versions and support downgrades

lubber-de:
## Description
The current nightly build is broken, because the next nightly version would be `2.9.0-beta-0`, which was already uploaded (by accident) to npm in 2019, so it already exists, which breaks the build.

This PR now checks for possible existing npm versions and increases the beta number accordingly.
> i found out that `2.9.0-beta.0` is not available in the `versions` property from the npm api, but it is in the `time` property of https://registry.npmjs.org/fomantic-ui

This PR also allows for possible downgrades.
Means, if we have a `2.9.0-beta.1234` but , by some reason, we would have to create a 2.8.9 at a later stage, the script would not allow that (the next github milestone was always supposed to be greater than the latest nightly)

### DIFF
--- a/scripts/nightly-version.js
+++ b/scripts/nightly-version.js
@@ -35,23 +35,38 @@ const getNpmPreRelease = async function () {
     .then(pr => pr === null ? ['beta', 0] : pr)
 }
 
+const getAllNpmVersions = async function () {
+  // The versions property sometimes does not include versions which are in "time"!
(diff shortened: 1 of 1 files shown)


prudho (approved):
LGTM

hammy2899 (approved):

Issue #1518, opened 2020-06-10
Calendar: Add onAfterSelect callback


PhilippGrashoff:
# Feature Request
Currently, there is the onSelect Callback. This callback is called BEFORE the selected date is transferred into the input field.
It would be nice to have a callback which is called AFTER the selected date was transferred to the input field.
My use case is simple: I implemented an auto-Submit form which submits whenever there is a **sensible** change in an input. Especially for a date, not every input is sensible, e.g. if someone types a date by keyboard. Here, submitting is only sensible when the enter key is pressed.
At the moment, I cant simply call the form submit in onSelect, as the value was not transferred to the input field at that point yet. Hence it is not sent to the server. A onAfterSelect callback would fix this.

At the moment I try to work around this in onSelect by something like:
```
                    function(date, mode) {
                        if(!date) {
                            return;
                        }
                       //set date to input before submitting the form
                        $(this).find('input').val(dateTodmY(date));
                        $("#someForm").form('submit');
                    }
```

PhilippGrashoff:
But I do not want to disable the manual date input. Its very handy in case you have to enter the date which is for Example 3 Years in the future. Would need 36 clicks in the calendar.

lubber-de:
Regarding the 36 clicks, its not necessary as i mentioned here https://github.com/fomantic/Fomantic-UI/issues/1517#issuecomment-642484258
We may put those two issues together into one to avoid postig the same ?

I still believe that additional suggested "AfterSelection" Event is not necessary because onChange and onBeforeChange are called right after onSelect is.
So, according to the usecase to do something _after_ the input field has been updated, you go with `onChange`

I understand, it's tricky tricky to find out when the user has actually stopped typing when onChange triggers, but this is part of your other issue #1517 

I would not even know where to call such an additional "afterSelect" in the code ( after onChange?  after onBeforeChange? after onSelect (as said, the input field is not updated at that time)

PhilippGrashoff:
Could we pass an additional parameter to onChange like source which would e.g. "keyboard" or "calendar" (when someone clicks a date)? This would make working in onChange easier.
I still think the most important part is that incomplete dates are not guessed and transformed into actual dates.

lubber-de:
@PhilippGrashoff Possibly 😉 
But you can already figure this out by checking for the `onSelect` event. It it was called right before the `onChange` you know ithe value comes from a cell selection. Otherwise it was a pure keyboard entry and you should validate the date value against your desired format.
[...]

Issue #1517, opened 2020-06-10
Calendar: onChange/onBeforeChange and keyboard input do work strange


PhilippGrashoff:
# Bug Report
When entering a value via keyboard into the input field of a calendar, it gets magically turned into a date in onChange/onBeforeChange. Even typing only a single "1" results in the date 2001-01-01.

## Steps to reproduce
1. use this simple jsFiddle: https://jsfiddle.net/nyg4vph1/3/
2. type anything in the date input. You will see in the console that even on the first sign you type, the parameters passed to onChange represent a real date. This is what IMO the bug is: Some user input which is not a valid date gets magically turned into some date, even if you only type "1" in the empty input.
3. Due to this, I cant see a way to determine in onChange() if the user actually typed a date or just anything that gets converted into a date.

## Expected result
Do not magically convert any typed input into a date, only if its in the defined format.

##Why
Think of an application which submits the form in the onChange handler. You will want to check first if the date the user entered is a valid date, and only then proceed. This is not possible, as any input gets turned into some date. Think of someone buying a ticket for a movie, accidently types the date incomplete, and instead of being noticed about this the user goes on buying a ticket for a date he never selected.

## Version
2.8.5

PhilippGrashoff:
But I do not want to disable the manual date input. Its very handy in case you have to enter the date which is for Example 3 Years in the future. Would need 36 clicks in the calendar.

lubber-de:
>  you have to enter the date which is for Example 3 Years in the future. Would need 36 clicks in the calendar.

Do you know you can click on the Year to select the year much more faster? So its about 6-7 clicks rather than 36 :wink:
|||
|-|-|
|![datein3years](https://user-images.githubusercontent.com/18379884/84360709-d2680f80-abca-11ea-8a17-b2c12a600e92.gif)|![datein3years_2](https://user-images.githubusercontent.com/18379884/84360720-d5fb9680-abca-11ea-94e9-1be6d1eabbd0.gif)|

However, i do understand entering the date manually is probably much faster anyway

PhilippGrashoff:
Hi, I know that the date can be clicked, but from experience with the average user, I am very sure 95% dont :)

lubber-de:
Yes, that might be true 🤣 , i just wanted to clarify 36 clicks are not really needed technical speaken 😉 
So, back to the main topic: I initially suggestd to implment either a `minCharacter` setting or some kind of format check before a date is even tried to be parsed.
[...]

PR #2001, opened 2021-06-29
 feat(calendar): trigger onSelect() event on enter key down

prudho:
## Description
This PR adds triggering of the `onSelect` event when user press the Enter key after entering a date in the input field. It also allow user to rollback if the date field is not correct for him by returning `false` (can be usefull for #1517 and #1518).

## Screenshot (if possible)
Before:
Before

After:
After


ColinFrick (approved):

lubber-de (approved):
LGTM

PR #2002, opened 2021-06-29
fix(calendar): blur input on hidden

prudho:
## Description
This PR just blur the calendar when a date is selected and the popup has closed. Before, selecting a date just hide the popup but it still was focused. Now if you select a date the popup will hide and the input will be blurred. So you can click again to the input to reshow the popup.

## Screenshot
Before:
Before

After:
After


exoego (approved):
LGTM

lubber-de (changes requested):
This code is now duplicated and can be removed from here, because it is now the default

https://github.com/fomantic/Fomantic-UI/blob/e39a0ae5183c12808ba1eee82082efbb15f4d450/src/definitions/modules/calendar.js#L165-L170

prudho:
Ah, good catch @lubber-de 👍 Now it should be good :)

lubber-de:
I think we should make sure that #1508 is not affected. Maybe #1508 makes this PR even unneccessary (?)
So, maybe we should only add the autoblur to the input field when the listen event is 'focus' instead?
I also fear that a auto blur will probably autoopen another datefield (think of a range calendar)

I will prepare some testcase to make this #1508 and this will work together

lubber-de (approved):
LGTM
Testcase for this PR and also having the new default "click" from #1508 
https://jsfiddle.net/lubber/ztdobucf/12/

Issue #2009, opened 2021-07-06
[Checkbox] More Features for Checkbox


KiddoV:
# Feature Request
We are at v2.8.8 and checkboxes still look...uh... plain.
I think FUI checkboxes deserve to be better :P. Would be nice to see more features for [Checkbox] in the future.

## Example (if possible)
Something like this library to be consider...
https://lokesh-coder.github.io/pretty-checkbox/

Issue #550, opened 2019-03-08
[Feature Request] Toggle text with toggle or slider button


ko2in:
# Feature Request
It would be nice if we can switch the text upon the state of toggle or slider [checkbox](https://fomantic-ui.com/modules/checkbox.html). Sometimes, we need to display different text after changing the state of toggle button.

**Example:** If the toggle button is for On/Off, or Show/Hide option, when the user toggle the button, it should change 'On' to 'Off' or 'Show' to 'Hide' and vice versa.

## Example and testcase
I created a sample here (https://jsfiddle.net/8dn1yes5/)

Issue #393, opened 2019-01-14
Toggle checkbox - Background color


etshy:
# Feature Request

## Description
Add the possibility to change the Chackbox toggle background

## Example
Example rule for that
```
.ui.toggle.checkbox[class*="red background"] input:checked~.box:before, .ui.toggle.checkbox[class*="red background"] input:checked~label:before {
    background-color: red!important;
}
```
Put that in an `each` and replace class `red` with the `@color`.

## Testcase (when possible)

https://jsfiddle.net/pbuok0w5/

PR #2010, opened 2021-07-07
[Checkbox][WIP] New features

prudho:
## Description
Following #2009, I think that indeed checkboxes need some fancy new features too. So I open this PR as a draft and discussion topic.

For now, I just add `colored` and `outline` checkboxes too see what it's possible to implement.

The others options i'm thinking are:
- icons instead of check mark.
- animated checks (js and `Transition` are paving the way).
- label text depending on checkbox state...

What do you guys think about this ?

## Testcase
[JSFiddle](https://jsfiddle.net/j3xmnq4t/)

## Screenshot
<a href="#">Capture</a>

## Related issues
#393, #550, #2009

### DIFF
--- a/src/definitions/modules/checkbox.less
+++ b/src/definitions/modules/checkbox.less
@@ -793,6 +793,62 @@
     }
 }
 
+& when (@variationCheckboxColored) {
+  /*--------------
+       Colored
+  ---------------*/
+
+  each(@colors, {
+    @color: replace(@key, '@', '');
+    @c: @colors[@@color][color];
+    @l: @colors[@@color][light];
+    @isVeryDark: @colors[@@color][isVeryDark];
+  
+    .ui.@{color}.checkbox input:checked ~ label::after {
+      color: @color;
+    }
+    .ui.@{color}.radio.checkbox input:checked ~ label::after {
+      background-color: @color;
+    }
+    .ui.@{color}.toggle.checkbox input:checked ~ label::before,
+    .ui.@{color}.slider.checkbox input:checked ~ label::before {
+      background-color: @color !important;
+    }
+
+    & when (@variationCheckboxOutline) {
+      /*--------------
+        Outline
+      ---------------*/
+      .ui.@{color}.outline.checkbox input:checked ~ label::before {
+        border-color: @color;
+      }
+      .ui.@{color}.outline.toggle.checkbox input:checked ~ label::after,
+      .ui.@{color}.outline.slider.checkbox input:checked ~ label::after {
+        box-shadow: @subtleShadow,
+        0 0 0 1px @color inset;
+      }
+    }
+  })
+}
+
+/*--------------------
+        outline
+---------------------*/
+
+& when (@variationCheckboxOutline) {
+  .ui.outline.checkbox input:checked ~ label::before {
+    border-color: @checkboxActiveCheckColor;
+  }
+  .ui.outline.toggle.checkbox input:checked ~ label::after {
+    box-shadow: @subtleShadow,
+    0 0 0 1px @toggleOnLaneColor inset;
+  }
+  .ui.outline.slider.checkbox input:checked ~ label::after {
+    box-shadow: @subtleShadow,
+    0 0 0 1px @checkboxActiveCheckColor inset;
+  }
+}
+
 /*--------------------
         Size
 ---------------------*/
--- a/src/themes/default/globals/variation.variables
+++ b/src/themes/default/globals/variation.variables
@@ -388,6 +388,8 @@
 @variationCheckboxFitted: true;
 @variationCheckboxRightAligned: true;
 @variationCheckboxSizes: @variationAllSizes;
+@variationCheckboxColored: true;
+@variationCheckboxOutline: true;
 
 /* Dimmer */
 @variationDimmerInverted: true;


lubber-de:
I like the approach to make `outline` optional 👍🏼 !
I also like your additional ideas 🙂

src/definitions/modules/checkbox.less
```diff
@@ -793,6 +793,62 @@
...
+    .ui.@{color}.checkbox input:checked ~ label::after {
+      color: @color;
+    }
+    .ui.@{color}.radio.checkbox input:checked ~ label::after {
+      background-color: @color;
+    }
+    .ui.@{color}.toggle.checkbox input:checked ~ label::before,
+    .ui.@{color}.slider.checkbox input:checked ~ label::before {
+      background-color: @color !important;
+    }
```
lubber-de, line 816: Beside the variable name: Any chance to increase specificity without using `!important`? (Haven't tested myself yet!)
```suggestion
    .ui.ui.@{color}.toggle.checkbox input:checked ~ label::before,
    .ui.ui.@{color}.slider.checkbox input:checked ~ label::before {
      background-color: @c;
    }
```

src/definitions/modules/checkbox.less
```diff
@@ -793,6 +793,62 @@
...
+
+    & when (@variationCheckboxOutline) {
+      /*--------------
+        Outline
+      ---------------*/
+      .ui.@{color}.outline.checkbox input:checked ~ label::before {
+        border-color: @color;
```
lubber-de, line 823: Same here 😉 
```suggestion
        border-color: @c;
```

src/definitions/modules/checkbox.less
```diff
@@ -793,6 +793,62 @@
...
+      }
+      .ui.@{color}.outline.toggle.checkbox input:checked ~ label::after,
+      .ui.@{color}.outline.slider.checkbox input:checked ~ label::after {
+        box-shadow: @subtleShadow,
+        0 0 0 1px @color inset;
```
lubber-de, line 828: Same here :wink:
```suggestion
        0 0 0 1px @c inset;
```

lubber-de (changes requested):
I know, this PR is marked as Draft and WIP, but i wanted to give "official" Feedback beside my other comment anyway 🙂 for the current code.

src/definitions/modules/checkbox.less
```diff
@@ -793,6 +793,62 @@
...
+  ---------------*/
+
+  each(@colors, {
+    @color: replace(@key, '@', '');
+    @c: @colors[@@color][color];
+    @l: @colors[@@color][light];
+    @isVeryDark: @colors[@@color][isVeryDark];
+  
+    .ui.@{color}.checkbox input:checked ~ label::after {
+      color: @color;
```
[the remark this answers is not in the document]
prudho: Fixed by 84d43de.

src/definitions/modules/checkbox.less
```diff
@@ -793,6 +793,62 @@
...
+    }
+    .ui.@{color}.radio.checkbox input:checked ~ label::after {
+      background-color: @color;
```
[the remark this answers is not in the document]
prudho: Fixed by 84d43de.

src/definitions/modules/checkbox.less
prudho: Fixed by 84d43de.

src/definitions/modules/checkbox.less
prudho: Fixed by 84d43de.

src/definitions/modules/checkbox.less
prudho: I tried `.ui.ui` and `.ui.ui.ui`, but still need that `!important` to work... 😞

lubber-de: It's because there is an existing `!important` at
```css
.ui.toggle.checkbox input:checked ~ label:before {
    background-color: #2185D0 !important;
}
```

We would need to remove the `!important` from there too

lubber-de: If you check the checkbox.less there are still a few `!important` for toggle and slider which are causing this. This PR is a good chance to get rid of them as well 😉

prudho:
Icon instead of checkmark is doable:

image

But it will require more tests and code than previous modifications. Should I create another PR ?

lubber-de:
> But it will require more tests and code than previous modifications. Should I create another PR ?

Up to you 😉 Just make sure that each PR is based on current develop (or do everything in one PR)

lubber-de:
I have a suggestion/idea/question:
What about replacing `outline` by `bordered`?
We already have `bordered icon` and `bordered image`

However, a checkbox/radio always has an "outline" or is "borderd". Your new feature does only make the outline/border more prominent/bold/colored

Opinion?

prudho:
Yeah I don't know how to really name it, since I "copy" class name from another framework. And I struck to find another class name for a no bordered checkbox as well (well the border disapear when checked).

prudho:
I've added the `icon` variation. Works well overall, except for sizes, i'm not comfortable with advanced scss yet 😞

lubber-de (changes requested):

PR #2011, opened 2021-07-08
fix(docs): various typos

luzpaz:
Found via `codespell -q 3 -S *.min.js,CHANGELOG.md -L adn,ba,te,ue,varius`


lubber-de (approved):
LGTM, thanks!
In future PRs, please do not change the files in the `dist` folder, as they will be created on the next release and would make the review a bit easier, because less files would be changed

prudho (approved):

PR #2012, opened 2021-07-08
fix(api): source typos

luzpaz:
Found via `codespell -q 3 -S ./dist.*.min.js,CHANGELOG.md -L adn,ba,te,ue,varius

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -334,7 +334,7 @@ $.api = $.fn.api = function(parameters) {
           cancelled: function() {
             return (module.cancelled || false);
           },
-          succesful: function() {
+          successful: function() {
             module.verbose('This behavior will be deleted due to typo. Use "was successful" instead.');
             return module.was.successful();
           },
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -335,10 +335,6 @@ $.api = $.fn.api = function(parameters) {
             return (module.cancelled || false);
           },
           successful: function() {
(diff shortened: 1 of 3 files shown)


src/definitions/behaviors/api.js
```diff
@@ -334,7 +334,7 @@ $.api = $.fn.api = function(parameters) {
           cancelled: function() {
             return (module.cancelled || false);
           },
-          succesful: function() {
+          successful: function() {
```
lubber-de, line 337: This will break the code because the function "was.successful" is already available in Line 341
See line 338
> module.verbose('This behavior will be deleted due to typo. Use "was successful" instead.');

Instead, you can completely remove the typo'ed function "was.succesful", as the next version 2.9.0 is supposed to be a minor release which allows for breaking changes.

lubber-de (changes requested):
The function name change sill break the code but you can remove it instead. See my review

src/definitions/behaviors/api.js
luzpaz: IOW, I should revert this change ?

lubber-de: Either revert the change or, as already suggested, remove the whole function "successful". The correct spelled "successful" function is already there and used. The old spelled function was left in the code for backward compatibility

luzpaz: Thanks for the clarification. Done in 9281165

lubber-de (approved):
LGTM

luzpaz:
Rebased and force pushed

lubber-de (dismissed):
LGTM

lubber-de (approved):
LGTM

PR #1801, opened 2020-12-18
[Tooltip] Bottom center tooltip was misaligned

lubber-de:
## Description

A `bottom center` tooltip had it's arrow misaligned and not correctly centered

## Testcase
https://jsfiddle.net/lubber/9mbupL4t/16/

## Screenshot
|Before|After|
|-|-|
|image|image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/7021

PR #2021, opened 2021-07-20
fix(tooltip): fix arrow position for bottom center

lubber-de:
## Description
The Fix from #1801 was basically correct, however the specificity for transform-origin in later selectors override it, so the result was even worse

## Testcase
https://fomantic-ui.com/modules/popup.html#position

## Screenshot
|Before|After|
|-|-|
|image|image|


exoego (approved):
LGTm

PR #2022, opened 2021-07-21
fix(toast): dedicated horizontal container

lubber-de:
## Description
Whenever a horizontal toast was created, all further toast, in the same container position, were shown horizontal as well, even if they were supposed to be displayed normal/vertically.

Reason for this was the container selector, which was only defined for the positioning, but ignored a possible horizontal setting. This PR now creates a dedicated `horizontal` container, which fixes such situations.

## Testcase

- Click the first Button to show horizontal toasts
- Now click the second button, which should show normal/vertical toast in the same default position (top right)

### Broken
The vertical toasts are shown horizontal as well 
https://jsfiddle.net/lubber/znp9gr7o

### Fixed
The vertical toasts are shown vertical, no matter if a horizontal toast was shown before or not 
https://jsfiddle.net/lubber/znp9gr7o/2/

### Broken
toasthorizontalbroken

### Fixed
toasthorizontalfixed


prudho (approved):
LGTM

PR #2036, opened 2021-08-06
feat(modal): a11y support, tabbing in modal, close icon focus/keypress

lubber-de:
## Description
This PR adds accessibility (a11y) support to the modal module

- add proper aria attributes
- make sure tabbing only takes place inside the modal and does not tab to fields hidden by the dimmer
- make a close button a button and tabbable as well

I also separated the closeIcon from the previous close selector which was supposed to be able to also close the modal on each action button when the selector is enhanced.
Reason for separation of closeIcon is the new ability to trigger the closeicon by space or enter which is needed as accessibility requirement and was not possible before.

However, the tabbing feature works independently of existing aria labels and was a general bug anyway.

## Testcase
- Try to tab through the modals fields / buttons
- Try to focus the close icon and press space or enter
- Use a screen reader or check developers tools tab "accessibility" on the generated modal dom node to look for aria labels

### Before
- Tabbing after last action button leads into tabbing something else of your browser or even OS
- Tabbing before the first action button will focus the button behind the dimmer and after that your browser address bar and OS elements
- The close icon cannot be focussed, thus trying to press space or enter is useless and has no effect
- No aria attributes / labels, modal is not recognized in a screen reader properly
- Icon only Button is not recognized by screen reader

https://jsfiddle.net/lubber/7tqo3xkm/15/

### After
- Tabbing only works inside the modal
- Close icon can be focussed
- Pressing space or enter when close icon is focussed will close the modal
- A screen reader will recognize the icon only button as "mail" buttons
- A screen reader (or developer tools, tab "accessibility" will recognize the modal with proper aria attributes/labels

https://jsfiddle.net/lubber/7tqo3xkm/14/

### Before
aria_modal_wrong

### After
aria_modal_right

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5773
https://github.com/Semantic-Org/Semantic-UI/issues/733

## Related
https://github.com/Semantic-Org/Semantic-UI/issues/348
https://github.com/go-gitea/gitea/issues/7057


prudho (approved):
Woah @lubber-de this is a great addition ! I'm not an expert in accessibility/aria support but from my perspective all looks good to me. Great job 👍

PR #2041, opened 2021-08-09
feat(form): new default true for errorfocus

lubber-de:
## Description
As proposed in https://github.com/fomantic/Fomantic-UI/discussions/1745#discussion-29935 this PR changes the default value for `errorFocus` to `true`.
The new default will always focus the first error field of a form when the form is invalid, which is more logical. By the previous Logic to adjust a form field content, you almost always have to click into that field first.


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/295

exoego (approved):
This sounds better from end user PoV.

prudho (approved):

ColinFrick (approved):

PR #2043, opened 2021-08-09
feat(calendar): new default constantheight false

lubber-de:
## Description
As proposed in https://github.com/fomantic/Fomantic-UI/discussions/1745#discussioncomment-120894 this PR changes the default value for `constantheight` to `false`.

Whenever the first of a month is shown within the first 5 columns of a calendar, then, according to the months amount of days, a full adjacent row is added to the calendar, which is unnecessary in most cases, so i suggest to turn this setting off by default

### Screencapture
calendar-height

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1433,7 +1433,7 @@ $.fn.calendar.settings = {
 
   type               : 'datetime', // picker type, can be 'datetime', 'date', 'time', 'month', or 'year'
   firstDayOfWeek     : 0,          // day for first day column (0 = Sunday)
-  constantHeight     : true,       // add rows to shorter months to keep day calendar height consistent (6 rows)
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/298

exoego (approved):
LGTM
Added screen capture in the PR description.

prudho (approved):

lubber-de:
Mmh, there is one situation where this might be annoying if it's default: When the calendar is displayed on top of the calendar field: Because of the height changes you cannot constantly click on the month arrows.

Should we change the setting anyway?

https://jsfiddle.net/lubber/2gw7f8ze/1/

calendartopconstantheight

prudho:
Well this can be annoying... At my work for example, they always need to change month in calendars. Maybe we should not define it to `true` by default 🤔

lubber-de:
Agree, i will close this PR, let's keep the default as it is.

PR #2044, opened 2021-08-09
feat(dropdown): new default showonfocus false

lubber-de:
## Description
As proposed in https://github.com/fomantic/Fomantic-UI/discussions/1745#discussioncomment-129264 this PR changes the default value for `showOnFocus` to `false`.

This will stop the dropdown from opening but allowing it to retain focus and would be better in terms of accessibility (in a manner of speaking)

I don't believe <select> tags normally open when they gain focus (at least with autofocus) anyway (only when the enter key is hit does it open) and using the arrow keys goes through the options without it opening at all. Of course if it is clicked or tapped on, it should open...


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/297

exoego (approved):
LGTM

prudho (approved):

PR #2046, opened 2021-08-10
fix(dropdown): add namespace to sessionstorage

lubber-de:
## Description
Whenever more than 1 dropdown is using remoteData, the sessionstorage is sharing the keys because of a missing namespace.
Now, if different dropdowns use the same ids, but different values then the second dropdown will return the wrong names from the first dropdown.
This PR adds the dropdowns id/namespace to the sessionstorage keys, so those are now unique to their respective dropdowns.

### Example
Dropdown 1 has the following elements (key, value)
1, monday
2, tuesday
3, wednesday

Dropdown 2 has the following elements (key, value)
1, january
2, february
3, march

The second dropdown will get the Weeknames out of the sessionstorage instead of the monthnames, because the keys are the same.

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5211

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2324,7 +2324,7 @@ $.fn.dropdown = function(parameters) {
               module.error(error.noStorage);
               return;
             }
-            name = sessionStorage.getItem(value);
+            name = sessionStorage.getItem(value + elementNamespace);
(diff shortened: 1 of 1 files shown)


prudho (approved):
LGTM

exoego (approved):
LGTM

PR #2047, opened 2021-08-10
fix(api): respect post vars for caching

lubber-de:
## Description
When the API uses local caching, it stores the response to the key, which only consists of the url.
While this is fine for the default method GET (which contains the get parameter as part of the url), this won't work if POST is used (and the url is most probably always the same)

This PR stringifies all existing url data and adds this to the url for a unique sessionstorage key to make sure the response cache corresponds to the parameters.

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5093


src/definitions/behaviors/api.js
```diff
@@ -148,8 +148,8 @@ $.api = $.fn.api = function(parameters) {
               module.error(error.noStorage);
               return;
             }
-            response = sessionStorage.getItem(url);
-            module.debug('Using cached response', url, response);
+            response = sessionStorage.getItem(url + JSON.stringify(settings.data));
```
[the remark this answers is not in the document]
lubber-de: @exoego Thanks for the hint! Should be fine with 3f950f4ad71d0bda22d4dda776bd814ffd5d548c now
I intergrated https://github.com/hongymagic/jQuery.serializeObject/blob/master/jquery.serializeObject.js to make sure it's always a json object and to remove a dependency

exoego (dismissed):
LGTM

lubber-de:
@exoego I added a fix for IE11 which does not support a string for `Object.keys()`
This could happen whenever somebody provides a string for settings.data (which is a valid setting for jquery ajax) and keeps `serializeForm:false`

ColinFrick (approved):

src/definitions/behaviors/api.js
```diff
@@ -148,8 +148,8 @@ $.api = $.fn.api = function(parameters) {
               module.error(error.noStorage);
               return;
             }
-            response = sessionStorage.getItem(url);
-            module.debug('Using cached response', url, response);
+            response = sessionStorage.getItem(url + (typeof settings.data === "string" ? settings.data : JSON.stringify(settings.data,Object.keys(settings.data).sort())));
```
exoego, line 151: I suggest to extract function, since these duplications are non-trivial.

```javascript
function normalize(data) {
  return (typeof data === "string" ? data : JSON.stringify(data, Object.keys(data).sort())
}

// use-site
normalize(settings.data)
```

lubber-de: Done by b5aa6fca667e03759d1f1eb841f2a6a78ad05a02 
I decided to do it for the dedicated settings.data variable and put it under the `get` bahavior. This way it is also possible to retrieve this separately via behavior if needed

exoego (approved):
LGTM

lubber-de:
Docs regarding included serializeObject dependecy  added by https://github.com/fomantic/Fomantic-UI-Docs/pull/313

PR #2049, opened 2021-08-10
[Checkbox]!: Remove old v1 checkbox box code

prudho:
This PR delete the old code for v1 checkboxes that isn't even documented anymore.

Since target 2.9 is a breaking change, I think that we can delete it before the 3.0 release safely.


lubber-de (approved):
LGTM

PR #2085, opened 2021-08-21
feat(input): support invalid pseudoclass

lubber-de:
## Description
New native input types like "week", "date", "email", "url" etc. have a browser inbuild validation.
When those get invalid the pseudo selector `:invalid` is set. In general all input field types, which have the `required` attribute set and are empty, are considered invalid. 

This PR adds the error class styling to such input elements by default. If nothing has been entered in such fields, those are not considered invalid (as long as they do not also have the `required` attribute set).

I decided to only show the invalid error colors when there is either something entered invalid or no placeholder is given. This way one can decide to not show the error colors on empty fields (initial forms) by defining a placeholder (even a " " works , if somebody does not want a textual placeholder set).

## Testcase
Enter something invalid and the fields get automatically styled, even though no js form validation is triggered
https://jsfiddle.net/lubber/947q8xve/66/

## Screenshot
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6618

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -578,7 +578,14 @@
       box-shadow: @formStates[@@state][boxShadow];
     }
     & when (@state=error) {
-      .ui.form .field input:invalid {
+      .ui.form .field input:not(:placeholder-shown):invalid {
+        color: @c;
+        background: @bg;
+        border-color: @formStates[@@state][borderColor];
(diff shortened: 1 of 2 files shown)


exoego (approved):
Ah, I did not know :invalid !!

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/333

PR #2088, opened 2021-08-26
feat(chore): color list for each component separately

lubber-de:
## Description

This PR allows to define a color list for each component individually out of the central `colors.less` file.

For example, now it is possible to have only red, blue and orange buttons, but only yellow labels and no colors for message at all. See example below.

All of this can be defined in the central `variation.variables` file as before (or in custom theme file)

Before this change you were only able to reduce the whole color list by modifying the colors.less file and change the available colors for all components at once.

> ### Tip for reviewers: 
> For an easier overview add `?w=1` to the review URL which will ignore all the whitespace changes. github also has an  option for this via the gear icon
image

### Example config for the above mentioned different colors per component
```less
// Only red, blue and orange buttons
@variationMessageColors: red, blue, orange;
// Only yellow labels
@variationLabelColors: yellow;
// No colors for message at all
@variationMessageColors: false;
```

### Example to reduce the colors in all components inside the `variation.variables` file only without the need to modify the `colors.less` (unless one needs totally new color names of course)

```less
// Simply adjust the central variable
// only green and black for all components
@variationAllColors: green, black;
```

#### I also applied the same logic for states and consequences:
```less
// Only info and warning textstates
@variationTextStates: info, warning;

// only Error form states
@variationFormStates: error;

// No dropdrop formstates
@variationDropdownStates: false;

// only positive and negative message consequences
@variationMessageConsequences: positive, negative;
```

The default theme will still build all colors into every component as before.

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -310,9 +310,9 @@
 /*--------------------
    Types of Messages
 ---------------------*/
-& when (@variationFormStates) {
-  each(@formStates, {
-    @state: replace(@key, '@', '');
+& when not (@variationFormStates = false) {
+  each(@variationFormStates, {
+    @state: @value;
     .ui.form .@{state}.message,
     .ui.form .@{state}.message:empty {
       display: none;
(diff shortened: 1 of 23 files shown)


lubber-de:
@exoego I merged the previous PR #2106 so, this one can be reviewed again, please.
As before, for easier review remove the whitespace changes
https://github.com/fomantic/Fomantic-UI/pull/2088/files?w=1

ColinFrick (approved):

lubber-de:
Doc hint added by https://github.com/fomantic/Fomantic-UI-Docs/pull/377

PR #2093, opened 2021-09-04
feat(dimmer): use modern blurring to improve performance

lubber-de:
## Description
The `blurring` dimmer is inperformant because every single node inside the dimmable context is blurred separately which takes lots of time. 
It also causes positioning glitches when applied (Try the blurring example at https://fomantic-ui.com/modules/modal.html#dimmer-variations and see the sidebar suddenly vanishes!) 
Additionally when used in modals it is not nicely animated when the modal hides.

This PR switches from the old filter approach on each single node to the modern backdrop-filer approach on the dimmer itself only which is much more performant and nicely animates now.
For each browser not supporting this technique, the old behavior will still apply as before.

## Testcase
Remove CSS to see the difference
https://jsfiddle.net/lubber/egzbcoL2/35/

## Screenshot
|Before|After|
|-|-|
|oldblur|newblur|

## Fixes
https://github.com/Semantic-Org/Semantic-UI/issues/5932
This PR fixes the mentioned issue in chrome.
One has to enable the backdrop-filter in the firefox settings and it will be fixed in firefox as well.

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -171,17 +171,21 @@ body.dimmable > .dimmer {
       Blurring
   ---------------*/
 
-  .blurring.dimmable > :not(.dimmer) {
-    filter: @blurredStartFilter;
-    transition: @blurredTransition;
-  }
-  .blurring.dimmed.dimmable > :not(.dimmer):not(.popup) {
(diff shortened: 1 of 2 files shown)


exoego (approved):
Great!!

PR #2097, opened 2021-09-08
feat(chore): same custom scrollbars in chrome, FF and IE

lubber-de:
## Description

Custom scrollbars are already implemented since a long time for non body elements.
However, only webit based browsers were targeted.

This PR now adds the same look and feel to firefox and IE as well (as far, as those browsers support it)

## Testcase
https://jsfiddle.net/lubber/h0n7bur6/3/
Remove CSS to see the behavior before.

## Screenshot
|webkit before|firefox before|IE before|webkit AFTER|firefox AFTER|IE AFTER|
|-|-|-|-|-|-|
|image|image|image|image|image|image|


exoego (approved):
LGTM

PR #2108, opened 2021-09-17
fix(modal): respect approve deny callback on js modals

### LINKED ISSUE #2105: [Modal] onApprove cannot prevent hiding when modal is created with JS API.
# Bug Report
As per documentation (and module code), if `onApprove` callback returns `false` then modal should not be hidden on clicking approval button. But it still hides.

`onApprove` return value checked in: https://github.com/fomantic/Fomantic-UI/blob/e39a0ae5183c12808ba1eee82082efbb15f4d450/src/definitions/modules/modal.js#L322

## Steps to reproduce
1. Create modal with JS API
2. Show modal
3. Click approval button

## Expected result
Modal should not be hide.

## Actual result
Modal hides.

## Testcase
https://jsfiddle.net/pbcr1qL5/5/

## Version
2.8.8

lubber-de:
## Description
When using modals via JS properties only having any approve/deny buttons declared as actions, those did not respect the `onApprove` or `onDeny` callbacks returning a possible false to prevent hiding the modal.

## Testcase
Try to click on either buttons. Both should _not_ hide the modal as the onApprove and onDeny callbacks return false.

### Broken
https://jsfiddle.net/lubber/np2bdLew/14/

### Fixed
https://jsfiddle.net/lubber/np2bdLew/15/

## Closes
#2105

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -127,7 +127,8 @@ $.fn.modal = function(parameters) {
                 'aria-label': $('<div>'+(el[fields.text] || el[fields.icon] || '')+'</div>').text(),
                 class: className.button + ' ' + cls,
                 click: function () {
-                  if (click.call(element, $module) === false) {
+                  var button = $(this);
+                  if (button.is(selector.approve) || button.is(selector.deny) || click.call(element, $module) === false) {
                     return;
                   }
                   module.hide();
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/311

exoego (approved):
LGTM

PR #2109, opened 2021-09-17
fix(icon): adjust icons width to match single icon 

shadoWalker89:
## Description
I have a dropdown with two items

The first item has a simple icon

`<i class="blue file excel outline icon"></i>`

The second one has two grouped icons

```html
<i class="icons">
    <i class="red file excel outline icon"></i>
    <i class="red bottom right corner key icon"></i>
</i>
```

Because the `.icon` has a width and the `.icons` does not visually they look off. The grouped icons has a smaller width then the simple icon

This PR adds a min-width to the `.icons` to make sure that the .icons group will be at least the same size of the `.icon`
I didn't want to use `width` because in some cases maybe the grouped icons need more space.

For the same reason i added a min-width i added also a min-height.
the height is not the cause of my problem, but i think it's best to add it as well.

## Testcase
https://jsfiddle.net/xoaj6kgh/

## Screenshot (if possible)

image

### DIFF
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -376,6 +376,8 @@ i.icons {
     display: inline-block;
     position: relative;
     line-height: 1;
+    min-width: @width;
+    min-height: @height;
   }
 
   i.icons .icon {
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -105,6 +105,7 @@ i.emphasized.icon:not(.disabled), i.emphasized.icons:not(.disabled) {
           Fitted
   --------------------*/
 
+  i.fitted.icons,
   i.fitted.icon {
     width: auto;
     margin: 0 !important;
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -378,6 +378,8 @@ i.icons {
     line-height: 1;
     min-width: @width;
     min-height: @height;
+    margin: 0em @distanceFromText 0em 0em;
+    text-align: center;
   }
 
   i.icons .icon {
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -379,7 +379,7 @@ i.icons {
     line-height: 1;
     min-width: @width;
     min-height: @height;
-    margin: 0em @distanceFromText 0em 0em;
+    margin: 0 @distanceFromText 0 0;
     text-align: center;
   }
 


src/definitions/elements/icon.less
```diff
@@ -376,6 +376,8 @@ i.icons {
     display: inline-block;
     position: relative;
     line-height: 1;
+    min-width: @width;
+    min-height: @height;
```
lubber-de, line 380: Good catch. 
We should also add the same margin and text-align as a single icon, so there definately is no visual difference anymore
See https://jsfiddle.net/lubber/axyf20j5/8/
```suggestion
    min-width: @width;
    min-height: @height;
    margin: 0 @distanceFromText 0 0;
    text-align: center;
```

lubber-de (changes requested):
Thanks! Please see my suggestion and compare changes here https://jsfiddle.net/lubber/axyf20j5/8/

lubber-de (changes requested):
Please also adjust the `fitted icon` to support `icons` as well.
https://jsfiddle.net/lubber/axyf20j5/14/

https://github.com/fomantic/Fomantic-UI/blob/dada7018a1fdfbf5f085758737839af90108dc48/src/definitions/elements/icon.less#L107-L112
```css
   i.fitted.icons, 
   i.fitted.icon { 
     width: auto;
     margin: 0 !important; 
   } 
```

lubber-de:
I adjusted my second suggestion. Even for fitted icons, the min-width should not change back to auto, so please just add `i.fitted.icons` to the selector.

src/definitions/elements/icon.less
shadoWalker89: Changes applied

shadoWalker89:
I applied all changes as requested

lubber-de (dismissed):
Awesome. Thanks!

src/definitions/elements/icon.less
```diff
@@ -376,6 +377,10 @@ i.icons {
...
+    margin: 0em @distanceFromText 0em 0em;
```
lubber-de, line 382: Could you please remove the "em" from the zero values so it becomes
```css
margin: 0 @distanceFromText 0 0;
```

lubber-de: Nevermind, i just did it ;)

lubber-de (approved):

lubber-de:
@all-contributors please add @shadoWalker89 for code

PR #2111, opened 2021-09-17
feat(checkbox): use left/right to switch off/on toggle+slider variant

### LINKED ISSUE #2102: Toggle checkbox and slider checkbox should react to Left and Right key presses
https://fomantic-ui.com/, checkbox section, second and third items (slider checkbox and toggle checkbox): when the knob is on the left side, pressing the Right arrow key should shift it to the right side, and when it’s on the right side, pressing Left should shift it to the left side. (This is fairly standard behaviour.)

lubber-de:
## Description
This PR adds support to use the left/right keys on a `slider checkbox` or `toggle checkbox`  once focussed and initialted via JS.

## Testcase
- Focus a checkbox
- Press right to move the knot to the right
- Press left to move the knot to the left
https://jsfiddle.net/lubber/ne18mzk2/7/

## Closes
#2102

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -239,18 +239,26 @@ $.fn.checkbox = function(parameters) {
               }
             }
 
+            shortcutPressed = false;
             if(key == keyCode.escape) {
               module.verbose('Escape key pressed blurring field');
               $input.blur();
               shortcutPressed = true;
             }
-            else if(!event.ctrlKey && ( key == keyCode.space || (key == keyCode.enter && settings.enableEnterKey)) ) {
-              module.verbose('Enter/space key pressed, toggling checkbox');
-              module.toggle();
-              shortcutPressed = true;
-            }
-            else {
-              shortcutPressed = false;
+            else if(!event.ctrlKey && module.can.change()) {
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

src/definitions/modules/checkbox.js
```diff
@@ -239,18 +239,26 @@ $.fn.checkbox = function(parameters) {
...
               shortcutPressed = true;
             }
-            else if(!event.ctrlKey && ( key == keyCode.space || (key == keyCode.enter && settings.enableEnterKey)) ) {
-              module.verbose('Enter/space key pressed, toggling checkbox');
-              module.toggle();
-              shortcutPressed = true;
-            }
-            else {
-              shortcutPressed = false;
+            else if(!event.ctrlKey && module.can.change()) {
```
exoego, line 248: `module.can.change` is added to fix some issue?

lubber-de: No, i added this, because any possible change change inside the if statement can only happen when the checkbox is not disabled/reachonly (which can.change checks). 
The toggle function tests this itself, but the checked/unchecked dont, so i decided to check this in the first place

PR #2119, opened 2021-09-23
[build] fix exclusion of table component when tab is active

### LINKED ISSUE #2118: [Build] Can't exclude Table component if Tab is included
# Bug Report

I want to exclude the Table component from semantic.css, to reduce file size. It's still listed as an individual in semantic.json.

This leads to a funky glitch, where the CSS styles for Table are not removed from semantic.css _if_ the Tab component is also included _and if_ table.css is still present under dist/components. Probably a globbing issue in one of the Gulp tasks.

## Steps to reproduce

1. Remove 'table' as component in semantic.json (but don't remove table.css in dist/components)
2. Make sure 'tab' is still listed as component
2. Run 'gulp build'
3. Check semantic.css

## Expected result

All table classes should be removed from semantic.css if 'table' is not in the list of components.

## Actual result

No reduction in file size for semantic.css. All .ui.table classes are still there.

## Version

2.8.8

hugopeek:
## Description

When you remove the 'table' component from semantic.json while there is still a table.css file present in dist/components, it will not be removed from semantic.css.

This is caused by a collision in the globbing mechanism: 'tab' is included in the word 'table', so table.css is included in the package task.

## Closes
Fixes #2118


lubber-de:
The same logic is in the javascript build script as well. Could you please also apply your fix to the same logic in javascript.js ? 😉 

https://github.com/fomantic/Fomantic-UI/blob/f66e62d712f475c102af1f8fc6bdc4d36f11357f/tasks/build/javascript.js#L75

lubber-de (commented):
Please see my comment. I believe the same fix should be applied to the javascript build task as well

tasks/build/css.js
```diff
@@ -99,7 +99,12 @@ function pack(type, compress) {
     concatenatedCSS = compress ? filenames.concatenatedMinifiedCSS : filenames.concatenatedCSS;
   }
 
-  return gulp.src(output.uncompressed + '/**/' + globs.components + ignoredGlobs)
+  let src = output.uncompressed + '/**/' + globs.components + ignoredGlobs;
+  if (globs.components.indexOf('table') < 0 && globs.components.indexOf('tab') > 0) {
```
exoego, line 103:
```suggestion
  if (globs.components.indexOf('table') < 0 && globs.components.indexOf('tab') >= 0) {
```
Just in case `tab` is a prefix of components.

hugopeek: I'm not exactly sure what you mean with a prefix of components... The comparison is purely to see if the word 'tab' is present in the list, and 'table' is not.

indexOf will return the start position of the substring. So even if tab is the first component in the list, the output of indexOf is already 1:
```
{tab,reset,site,button,container,divider,header,icon,image,input,label,list,segment,text,breadcrumb,grid,menu,message,card,feed,item,statistic,accordion,checkbox,dropdown,embed,popup,rating,sidebar,sticky,transition,visibility}
1
```
Probably because of the { at the beginning.

hugopeek: It would be much better if the components list is processed as a proper array of course, but my JS-fu is too limited to make that happen (quickly).

hugopeek:
> The same logic is inside the javascript build script as well. Could you please also apply your fix to the same logic in javascript.js ? wink

@lubber-de Currently that is not necessary, because the table component doesn't have a table.js part. I could add it to future-proof this solution, but I'm never that eager to add code that doesn't do anything useful yet...

lubber-de:
> @lubber-de Currently that is not necessary, because the table component doesn't have a table.js part. I could add it to future-proof this solution, but I'm never that eager to add code that doesn't do anything useful yet...

😬 Oups... I totally forgot about that... Yes, then my suggestion is of course nonsense and doesn't make sense 😄

tasks/build/css.js
exoego: > Probably because of the { at the beginning.

Ah, I see.

exoego (approved):
LGTM

lubber-de (approved):
LGTM

PR #2131, opened 2021-10-01
feat(toast): image variant, a11y support, title html, click support

### LINKED ISSUE #2125: [toast] Click callback not called when closeOnClick is set to false
# Bug Report
When creating a toast, `onClick` callback not called when `closeOnClick` is set to false.

## Expected result
The `onClick` callback should work

## Actual result
The `onClick` callback doesn't work

## Testcase
https://jsfiddle.net/7fv4k0ag/

## Version
2.8.8

lubber-de:
## Description
Once again: Another year, another toast enhancement 🙂 

This PR

- adds accessibility support using aria labels to toasts just like we did for modals in #2036 
- adds a `ui image toast` variant to show images without any padding
- fixes missing html support in titles
- fixes wrong placement of vertical actions when the message was too short
- supports the `onClick` handler in any case now even if `closeOnClick` is set to false 
  - onClick can now also return false to prevent closing the toast
  - the close icon gets a dedicated event now independent of the onClick event

## Testcase
https://jsfiddle.net/lubber/xym8qg9j/37/

#### ui image toast
image

### wrong placement of actions
|Before|After|
|-|-|
|image|image|

### html titles
|Before|After|
|-|-|
|image|image|

## Closes
#2125

### DIFF
--- a/src/definitions/modules/toast.js
+++ b/src/definitions/modules/toast.js
@@ -69,12 +69,14 @@ $.fn.toast = function(parameters) {
         element          = this,
         instance         = isToastComponent ? $module.data(moduleNamespace) : undefined,
 
+        id,
         module
       ;
       module = {
 
         initialize: function() {
           module.verbose('Initializing element');
+          module.create.id();
           if (!module.has.container()) {
             module.create.container();
           }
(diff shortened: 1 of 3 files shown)


exoego (approved):
LGTM !!

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/338

PR #2142, opened 2021-10-23
[list] fix styling of multi-digit ordered lists (#1130)

### LINKED ISSUE #1130: Ordered and suffix list styles bug after 9.
# Bug Report

When ordered list (with added dots from list.variables) or suffix list has numbers over 9. the styling fails.

## Steps to reproduce
https://jsfiddle.net/qgms8p07/

## Expected result
Numbers have the same margin relative to text no matter how big the number is.

## Actual result
When list number is bigger than 9. the number/dot will have no margin relative to text and will even overlap.

## Testcase
https://jsfiddle.net/qgms8p07/

## Screenshot (when possible)
list

## Version
2.7.8

SnirBroshi:
## Description
Fixes styling of multi-digit ordered lists - made the number aligned to the right of the text, so that when more digits are required they will be added on the left and the right edge of the number will stay in place.

I used `translateX(-100%)` to keep the right edge in a constant position, and adjusted the margin a bit since it previously accounted for the width of the number itself. The number I adjusted it to is to keep the look of numbers with single digits the same.

## Testcase
https://jsfiddle.net/futybd2v/

## Screenshot
Current (left) vs fixed (right)
image

## Closes
Fixes #1130

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -703,6 +703,7 @@ ol.ui.list ol li,
     left: auto;
     user-select: none;
     pointer-events: none;
+    transform: translateX(-100%);
     margin-left: -(@orderedCountDistance);
     counter-increment: @orderedCountName;
     content: @orderedCountContent;
--- a/src/themes/default/elements/list.variables
+++ b/src/themes/default/elements/list.variables
@@ -179,7 +179,7 @@
 @orderedCountContent: counters(@orderedCountName, @orderedCountSeparator) " ";
 @orderedCountContentSuffixed: counters(@orderedCountName, @orderedCountSeparator) @orderedCountSuffix;
 @orderedCountColor: @textColor;
-@orderedCountDistance: 1.25rem;
+@orderedCountDistance: 0.65rem;
 @orderedCountOpacity: 0.8;
 @orderedCountTextAlign: right;
 @orderedCountVerticalAlign: middle;


lubber-de (changes requested):
Thanks for the PR. I changed your jsfiddle to use the same selector as used by the patch itself.
https://jsfiddle.net/lubber/L8f627oy/9/
You will recognize that nested ordered lists are not properly positioned anymore:
image
To fix this, you should ~remove the following code, because it is not needed by your translatex(-100%) approach anymore (as the variable `@orderedChildCountOffset` is then not used, you can also remove it from `list.variables` OR simply~ set the variable to 0.65rem as well (somehow more backward compatible if people might have changed it in their custom themes)..
https://github.com/fomantic/Fomantic-UI/blob/67babf538aeb681d8e476e84efc42b07f842a9be/src/definitions/elements/list.less#L737-L740

See https://jsfiddle.net/lubber/L8f627oy/11/
image

Also, such kinda long lists now do not have a nice left padding space anymore (at least when the counter reaches 100), so i suggest 2 additional changes
- Change the suffixed list margin left from 1.25rem to 1.5rem (because the dot moves even single numbers too much to the left otherwise) ONLY to the first list element (not for nested ones)
- Invent a new `long` variant which further increases the left margin to 2rem, again ONLY to the first list element (not for nested ones)

```css
ol.ui.suffixed.list,
.ui.suffixed.ordered.list,
ol.ui.suffixed.list ol {
    margin-left: 1.5rem;
}

ol.ui.long.list,
.ui.ordered.long.list,
ol.ui.long.list ol {
    margin-left: 2rem;
}
```

See https://jsfiddle.net/lubber/L8f627oy/24/

lubber-de:
@NotWearingPants Are you interested in continuation to work on this PR? Please see my review 😉

NotWearingPants:
@lubber-de I am, but finding the time isn't always easy :) I'mma try now

lubber-de:
@NotWearingPants Do you think you can work on this again?

lubber-de:
@NotWearingPants Maybe this got forgotten: Do you think you can work on this again? Please see my review

lubber-de:
> @lubber-de It's been too long for not replying. Should I take over or would you like to finish with your reviews?

My review was done already, so if you like you can take over

PR #2145, opened 2021-10-29
feat(input): file input variant

### LINKED ISSUE #105: [Form] file input styles
Add file upload input styles as mentioned here https://github.com/Semantic-Org/Semantic-UI/issues/403, would also be cool to implement as a module and have drag & drop functionality.

lubber-de:
## Description
This PR styles the file input field so it has the same look as normal input fields by adding a `file` variant to the `input` element.
It also supports using action inputs for file inputs, if one needs additional features like icons.
If not using an action input i also added color variants to support colors for the native file input trigger button.

It also adds an `invisible file input`, in case one does not need the filepath being shown and only needs a button.
See examples in the jsfiddle

## Screenshot
image
image

### Before
https://jsfiddle.net/lubber/xv0zLadf/3/

### After
https://jsfiddle.net/lubber/xv0zLadf/12/

## Closes
#105 
https://github.com/Semantic-Org/Semantic-UI/issues/403

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -66,10 +66,12 @@
 .ui.form .field > label {
   display: block;
   margin: @labelMargin;
-  color: @labelColor;
   font-size: @labelFontSize;
   font-weight: @labelFontWeight;
   text-transform: @labelTextTransform;
+  &:not(.button) {
+    color: @labelColor;
+  }
 }
 
 /*--------------------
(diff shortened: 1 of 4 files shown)


exoego (approved):
LGTM 👍 

This is awesome, 8-year-long-awaited feature https://github.com/Semantic-Org/Semantic-UI/issues/403

PR #2134, opened 2021-10-07
feat(table): scrolling variant, thead and tfoot colors, inverted colors

lubber-de:
## Description
This PR adds a `scrolling` variant, which offers a scrollable table body having a fixed header and footer. 
Just like for dropdown, i added height variants `short`, `very short`, `long` and `very long` respecting responsive layouts

I also fixed support for inverted colors  for better readability and some specificity fixes to the disabled variant.
You can now also add colors to the thead or tfoot section of a table (using the new `scrolling` feature one has to add the color directly to `thead` or `tfoot`, while with normal tables one would have to add the colors to the `tr` tag as before (however, it was was not working inside thead/tfoot before this PR, not even with non inverted tables)

## Testcase
https://jsfiddle.net/lubber/5df6jc2a/30/

## Screenshot
scrollingtable

## Closes
#535
#1519
#2081
https://github.com/Semantic-Org/Semantic-UI/issues/1357
https://github.com/Semantic-Org/Semantic-UI/issues/6781
https://github.com/Semantic-Org/Semantic-UI/issues/6778
https://github.com/Semantic-Org/Semantic-UI-React/issues/4378

PR #2148, opened 2021-11-01
 feat(table): all scroll stuck head foot first last variant

### LINKED ISSUE #1519: [table] scrollable tbody with sticky th
# Feature Request

It would be great with native support for scrollable tables with sticky header

https://stackoverflow.com/questions/29906134/semantic-ui-keep-thead-visible-when-scrolling-tbody

lubber-de:
## Description
This PR adds a `stuck` variant which supports to have the head, foot, first column and last column of a table being stuck while scrolling (using position `sticky`)

You can freely stick any combination like `head first stuck`, `foot first last stuck`, `head foot first stuck`, `first last stuck`, etc.
This feature needs to have the table being wrapped by a overflowing container having a fixed height (and/or optional width).
That said, i also added a new `scrolling container` and `scrolling segment` variant, which can be used for that purpose.

### Why `stuck` instead of `sticky` or `fixed`?
Because the other classnames already exists and provide different functionality

### Why not use classnames "header, footer, top, bottom, left, right" ?
Because the other classnames (beside "footer") already exists and provide different functionality

### What about the already merged `scrolling` table #2134 ? Why another approach? 
Unfortunately the other approach cannot properly support having the first or last column fixed (I tried some very ugly approaches which needed a fixed width of the columns and especially were not able to fill the background color into the whole cell.)

The new `stuck` variant makes use of sticky positioning which supports using a usual table element and does not have the above mentioned issues.

The `scrolling` table works in any browser, while `stuck` needs modern browsers with [position:sticky support](https://caniuse.com/?search=position%20sticky) to table elements . 

`stuck` possibly needs a wrapper when horizontal scrolling is not needed, while `scrolling`  does not.
That said, i also added new `scrolling container` and `scrolling segment` variants, which can be used for that purpose.

If you don't want to have an extra wrapper, you can use `overflowing stuck table`. However, this only works if your overall table columns exceed available width. 

Each new variant of `scrolling container`, `scrolling segment` or `overflowing table` support predefined heights using `short`, `very short`, `long` and `very long` 

See comparison table for details.

### Comparison

|Feature|stuck table|scrolling table|
|-|-|-|
|Supports fixed thead|:heavy_check_mark:|:heavy_check_mark:|
|Supports fixed tfoot|:heavy_check_mark:|:heavy_check_mark:|
|Supports fixed left column|:heavy_check_mark:||
|Supports fixed right column|:heavy_check_mark:||
|Supports combinations|:heavy_check_mark:||
|Supports column sizing (`wide`)[^1]|:heavy_check_mark:||
|Supports (`single line`)|:heavy_check_mark:||
|Supports vertical scrolling|:heavy_check_mark:|:heavy_check_mark:|
|Supports horizontal scrolling|:heavy_check_mark:||
|Needs wrapper[^2]|:heavy_check_mark:||
|Needs modern browser|:heavy_check_mark:||
|Scrollbar only for tbody||:heavy_check_mark:|

[^1]: on vertical scrolling
[^2]: to keep tables with less columns to 100% width

## Testcase
https://jsfiddle.net/lubber/hxugjqLk/6/

## Screenshot
stucktable

## Closes
#1519

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -783,6 +783,115 @@
   }
 }
 
+& when (@variationSegmentScrolling) {
+  /*--------------
+    Scrolling
+  ---------------*/
+  .ui.scrolling.segment {
+    overflow: auto;
+  }
+  @media only screen and (max-width : @largestMobileScreen) {
+    & when (@variationSegmentScrollingShort) {
+      .ui.scrolling.segment.short {
+        max-height: @scrollingMobileMaxHeight * 0.75;
+      }
+    }
+    & when (@variationSegmentScrollingVeryShort) {
+      .ui.scrolling.segment[class*="very short"] {
+        max-height: @scrollingMobileMaxHeight * 0.5;
+      }
+    }
+    .ui.scrolling.segment {
+      max-height: @scrollingMobileMaxHeight;
+    }
+    & when (@variationSegmentScrollingLong) {
+      .ui.scrolling.segment.long {
+        max-height: @scrollingMobileMaxHeight * 2;
+      }
+    }
+    & when (@variationSegmentScrollingVeryLong) {
+      .ui.scrolling.segment[class*="very long"] {
+        max-height: @scrollingMobileMaxHeight * 3;
+      }
+    }
+  }
+
+  @media only screen and (min-width: @tabletBreakpoint) {
+    & when (@variationSegmentScrollingShort) {
+      .ui.scrolling.segment.short {
+        max-height: @scrollingTabletMaxHeight * 0.75;
+      }
+    }
+    & when (@variationSegmentScrollingVeryShort) {
+      .ui.scrolling.segment[class*="very short"] {
+        max-height: @scrollingTabletMaxHeight * 0.5;
+      }
+    }
+    .ui.scrolling.segment {
+      max-height: @scrollingTabletMaxHeight;
+    }
+    & when (@variationSegmentScrollingLong) {
+      .ui.scrolling.segment.long {
+        max-height: @scrollingTabletMaxHeight * 2;
+      }
+    }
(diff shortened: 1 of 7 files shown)


exoego (approved):
Ah, this was a feature I needed but found difficult to implement.
Great 🥇

lubber-de:
I made the `overflowing` selector independent of the `stuck` variant  by c18c4619781e676321267a069fb32b201f8fb815 now.
@exoego Please review again

src/definitions/elements/segment.less
```diff
@@ -783,6 +783,115 @@
...
+        max-height: @scrollingComputerMaxHeight * 2;
+      }
+    }
+    & when (@variationSegmentScrollingVeryLong) {
+      .ui.scrolling.segment[class*="very long"]{
+        max-height: @scrollingComputerMaxHeight * 3;
+      }
+    }
+  }
+  @media only screen and (min-width: @widescreenMonitorBreakpoint) {
```
exoego, line 869: I think it might be better to take the screen height into account.
16:9 or 16:10 wide monitors usually have many vertical pixels, but some of recent ultra-wide monitors do not.

lubber-de: Well, we would need to implement overall additional height breakpoints then and basically adopt that for other components as well .🤔 it would probably also need make use of additional media queries to detect landscape mode.

The way i used it here was the same as you invented it for the dropdown long/short variants back then ☺️ 

Would you agree to figure that out in a separate PR ?

exoego: SGTM:+1:

exoego (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/353

PR #2150, opened 2021-11-03
fix(table): sortable border only when celled

### LINKED ISSUE #2149: [Table] Sortable, non-celled table header has border between cells
# Bug Report
Table header for sortable, non-celled table variant shouldn't have borders between cells for consistency with the table's body

## Steps to reproduce

1. Create a `ui sortable table`

## Expected result
Table header cells should have no left or right border, just as the rest of the table

## Actual result
Table header cells have the left border

## Testcase

[https://jsfiddle.net/efncg5d0/](https://jsfiddle.net/efncg5d0/)

## Version

2.8.8

lubber-de:
## Description

A sortable table header always got a left border, even when the table is not celled

## Testcase
https://jsfiddle.net/efncg5d0/

## Closes
#2149

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -1419,11 +1419,12 @@
   .ui.sortable.table > thead > tr > th {
     cursor: pointer;
     white-space: nowrap;
-    border-left: @sortableBorder;
     color: @sortableColor;
   }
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

PR #2164, opened 2021-11-22
feat(tab): callback to support tab change cancel

### LINKED ISSUE #796: Tab loading can't be cancelled
# Feature Request

## Description
I have forms in tabs and I'd like to implement a functionality similar to `beforeunload` in order to prevent loading other tabs (thus losing data) when a form has already been modified.

The contents of tabs are loaded via AJAX so I tried to manipulate the `onRequest` event in the `apiSettings` as well as the `beforeLoad` event of the tab but I had no success.

I think it would be useful to make tab changing cancellable with a callback returning `false`. I hope that you also like this idea.

Thanks for keeping the legacy of Semantic uI alive! Have a nice day:

Máté

lubber-de:
## Description
This PR adds a new callback "onBeforeChange"  before changing a tab, so you can cancel the tab change if needed.

## Testcase
The second tab cannot be clicked/changed anymore
https://jsfiddle.net/lubber/f5knryej/7/

## Closes
#796

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -379,6 +379,10 @@ $.fn.tab = function(parameters) {
                   module.verbose('Tab parameters found', nextPathArray);
                 }
               }
+              if (settings.onBeforeChange.call(element, currentPath) === false) {
+                module.debug('onBeforeChange returned false, cancelling tab change', $tab);
+                return false;
+              }
               if(isLastTab && remoteContent) {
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

prudho (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/320

PR #2167, opened 2021-12-01
fix(dropdown): split delimiter values on paste

### LINKED ISSUE #2166: [Dropdowns][Tags] Add in tokenization for Dropdowns when allowAdditions is true so user can input comma-delimited strings
# Feature Request

It would be really nice if fomantic had an option for tokenization in dropdowns when allowAdditions is set to true.  What I have in mind is the equivalent of [Select2's tokenization](https://select2.org/tagging#automatic-tokenization-into-tags) so that if a user input a comma delimited string (or any other character-delimited string), it would automatically separate those values. This way, users won't have to individually type in tags if they have a comma-delimited string elsewhere that they'd rather copy and paste.

## Example (if possible)

https://jsfiddle.net/9r57vm3e/2/

and then enter "third tag, fourth tag, fifth tag" in the dropdown. This will be saved as a single value. It would be really nice if it tokenized this and made them into separate tags.

## Screenshot (if possible)

Here's the desired behavior (screengrab from [Select2's tokenization](https://select2.org/tagging#automatic-tokenization-into-tags)):

In this example, i took the string `third, fourth, fifth` and pasted it into the box and it automatically separated it into three different values.

tokenization

lubber-de:
## Description
When pasting a string inside a `search selection dropdown` existing comma values (the default delimiter) is ignored. Even worse, when pressing enter afterwards the whole pasted string, containing delimiter values is used as a value itself.
This MR now supports the paste event and splits the pasted string into separate values and adds them when either `allowAdditions:true` or if the value exists in the dropdown menu.
Therefore the delimiter key is now also supported on multiple dropdowns which do not allow additions for consistency

## Testcase
https://jsfiddle.net/lubber/jc3xtshd 

### Before
pastedelimiterwrong

### After
pastedelimiterright

## Closes
#2166 
https://github.com/Semantic-Org/Semantic-UI/issues/5897
https://github.com/Semantic-Org/Semantic-UI/issues/3521

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -629,6 +629,11 @@ $.fn.dropdown = function(parameters) {
             $module
               .on('change' + eventNamespace, selector.input, module.event.change)
             ;
+            if(module.is.multiple() && module.is.searchSelection()) {
+              $module
+                  .on('paste' + eventNamespace, selector.search, module.event.paste)
+              ;
+            }
           },
           mouseEvents: function() {
             module.verbose('Binding mouse events');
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

PR #2175, opened 2021-12-08
fix(api): reset existing formdata for proper renewing

### LINKED ISSUE #2174: [API] Cannot delete element between two api call with form serialization
# Bug Report
I use the API module to send a form with the "serializeForm" parameter. In my form, I use a "multiple" dropdown and when I send it twice (with less elements the second time), I find elements from the first send in the second.

## Steps to reproduce
1. Create form (in my case with a "multiple" dropdown)
2. Select two elements in the dropdown
3. Send the form via API module (with the "serializeForm" parameter)
4. Remove elements in the dropdown and select another
5. Send the form again via API

## Expected result
Have only the items selected each time

## Actual result
We have old selected values

## Testcase
https://jsfiddle.net/54mnafog/6/

## Investigations
Apparently, the "settings.data" variable is not cleared between API calls and "module.add.formData" ("serializeForm") merges old and new form data. I tried with empty "settings.defaultData" but this property does not seem to be used. 

## Version
2.8.8

lubber-de:
## Desctiption
The apis `formData` collections always extended an already existing data value. If this however already contained properties of the same names as the formfields those were extended as well.
This led into situations where you want to reuse the same api several times. Especially for multiple name form fields like dropdown those always kept the previous values when submitted multiple times

## Testcase
- Open console
- Select three languages (English + French + Dutch)
- submit
- console show all three selected languages
- remove one (for example French)
- submit again.
 
### Broken
console again shows all three languages 
https://jsfiddle.net/lubber/05msby9h

### Fixed
console only shows the two languages
https://jsfiddle.net/lubber/05msby9h/5/

## Closes
#2174

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -443,6 +443,9 @@ $.api = $.fn.api = function(parameters) {
               } else {
                 formData[element.name] = element.value;
               }
+              if(hasOtherData && data[element.name]) {
+                delete data[element.name];
+              }
             });
 
             if(hasOtherData) {


prudho (approved):
LGTM

PR #2181, opened 2021-12-17
fix(api): empty dropdown data was overriden by previous data

### LINKED ISSUE #2174: [API] Cannot delete element between two api call with form serialization

lubber-de:
## Description
Revert and proper fix of #2175 
I forgot to check for an empty dropdown value. Jquery unfortunately does not return the field at all on `serializeArray` when multiple named fields (which should return an array) are empty.

## Testcase 
- Watch console
- Select something then submit
- The selected values will be logged
- Remove all selected entries and submit

### Before
The previous selected values are logged again
https://jsfiddle.net/lubber/05msby9h/5/

### After
No value is logged as the field is empty
https://jsfiddle.net/lubber/05msby9h/6/

## Closes 
#2174

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -74,6 +74,7 @@ $.api = $.fn.api = function(parameters) {
         url,
         data,
         requestStartTime,
+        originalData,
 
         // standard module
         element         = this,
@@ -86,6 +87,7 @@ $.api = $.fn.api = function(parameters) {
 
         initialize: function() {
           if(!methodInvoked) {
+            originalData = settings.data;
             module.bind.events();
           }
           module.instantiate();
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

PR #2184, opened 2021-12-19
fix(themes): less3/4 needs font import interpreted as raw css

lubber-de:
## Description

When using webpack 4+ and LESS 3/4 the font imports have to be interpreted as raw CSS, otherwise the build breaks.
This is already fixed for the default theme since 3 years, but any of the included optional themes will fail.

I also updated the google font api usage to v2

### DIFF
--- a/src/themes/instagram/views/card.overrides
+++ b/src/themes/instagram/views/card.overrides
@@ -3,7 +3,7 @@
 *******************************/
 
 
-@import url(https://fonts.googleapis.com/css?family=Montserrat:700,400);
+@import (css) url('https://fonts.googleapis.com/css2?family=Montserrat');
 
 .ui.cards > .card,
 .ui.card {
--- a/src/themes/material/modules/dropdown.overrides
+++ b/src/themes/material/modules/dropdown.overrides
@@ -1,4 +1,4 @@
-@import url(https://fonts.googleapis.com/css?family=Roboto:400,700);
+@import (css) url('https://fonts.googleapis.com/css2?family=Roboto');
 
 .ui.dropdown {
   font-family: 'Roboto';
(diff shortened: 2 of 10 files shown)


src/themes/material/modules/dropdown.overrides
```diff
@@ -1,4 +1,4 @@
-@import url(https://fonts.googleapis.com/css?family=Roboto:400,700);
+@import (css) url('https://fonts.googleapis.com/css2?family=Roboto');
```
exoego, line 1: Forgot [weight range](https://developers.google.com/fonts/docs/css2#axis_ranges) ? 
```suggestion
@import (css) url('https://fonts.googleapis.com/css2?family=Roboto:wght@400..700');
```

src/themes/instagram/views/card.overrides
```diff
@@ -3,7 +3,7 @@
 *******************************/
 
 
-@import url(https://fonts.googleapis.com/css?family=Montserrat:700,400);
+@import (css) url('https://fonts.googleapis.com/css2?family=Montserrat');
```
exoego, line 6: Forgot [weight range](https://developers.google.com/fonts/docs/css2#axis_ranges) ? 
```suggestion
@import (css) url('https://fonts.googleapis.com/css2?family=Montserrat:wght@700,400');
```

src/themes/material/modules/dropdown.overrides
lubber-de: Only weight 400 is available in the v2 api, that's why i removed the weight completely.
Your suggestions results in an unresolvable link:
https://fonts.googleapis.com/css2?family=Roboto:wght@400..700
image

src/themes/instagram/views/card.overrides
lubber-de: Same here: only 400 is available and your suggestion results in an unresolved link
https://fonts.googleapis.com/css2?family=Montserrat:wght@700,400
image

exoego (approved):
LGTM

PR #2185, opened 2021-12-19
fix(checkbox): wrong variable usage

lubber-de:
## Description
The toggle Checkbox LESS code accidently uses the slider variables for transition duration (obviously copy paste issue). 
For consistency, and this obvioulsy was the origin intention, a dedicated, already existing but never used, variable should be used.
For backward compatibility i changed the value to the same as for the slider variant, so the resulting css remains the same.

Thanks to @ghedsouza for [the original PR ](https://github.com/Semantic-Org/Semantic-UI-LESS/pull/23)

### DIFF
--- a/src/themes/default/modules/checkbox.variables
+++ b/src/themes/default/modules/checkbox.variables
@@ -149,7 +149,7 @@
 /* Toggle */
 @toggleLaneWidth: 3.5rem;
 @toggleHandleSize: 1.5rem;
-@toggleTransitionDuration: 0.2s;
+@toggleTransitionDuration: 0.3s;
 
 @toggleWidth: @toggleLaneWidth;
 @toggleHeight: @toggleHandleSize;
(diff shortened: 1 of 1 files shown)


ghedsouza (approved):
+1

ghedsouza:
Thanks @lubber-de.

src/themes/default/modules/checkbox.variables
```diff
@@ -149,20 +149,20 @@
 /* Toggle */
 @toggleLaneWidth: 3.5rem;
 @toggleHandleSize: 1.5rem;
-@toggleTransitionDuration: 0.2s;
+@toggleTransitionDuration: 0.3s;
```
exoego, line 152: Why increased ?

exoego: `0.3s` from `toggleTransitionDuration` 
https://github.com/fomantic/Fomantic-UI/blob/863d7123323f642238daaa5c92f005073fb9e204/src/themes/default/modules/checkbox.variables#L115

exoego (approved):
LGTM

src/themes/default/modules/checkbox.variables
lubber-de: Yup, that's why :)

PR #2186, opened 2021-12-19
fix(segment): horizontal should use group variables

lubber-de:
## Description
Horizontal segments should use their respective group variables (those however still contain the same values as for the single segment variabels used before)

## Closes
https://github.com/Semantic-Org/Semantic-UI-LESS/issues/49


exoego (approved):
LGTM

PR #2193, opened 2021-12-30
fix(modal): default templates handler was not working

lubber-de:
## Description
The default modals config templates (alert/confirm/prompt) button handler does not work anymore since #2108 because the default selectors require the `onapprove` and `ondeny` selectors.

This PR now supports both: appvove/deny selectors as well as custom selectors (e.g. by globally overriding them via `$.fn.modal.settings.className.ok = 'purple';`

## Testcase
Try all three buttons. Whatever you click to close the modal, a toast message should be shown

### Broken
No toasts, because the handler is ignored
https://jsfiddle.net/lubber/jpcdh06r/10/

### Fixed
All fine as expected
https://jsfiddle.net/lubber/jpcdh06r/11/

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -1453,33 +1453,39 @@ $.fn.modal.settings.templates = {
   },
   alert: function () {
     var settings = this.get.settings(),
-        args     = settings.templates.getArguments(arguments)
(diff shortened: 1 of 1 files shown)


exoego (approved):
nice catch!!

PR #2196, opened 2022-01-06
fix(modal): outside close button handling

### LINKED ISSUE #2195: [modal] multiple modals close buttons come on each other
# Bug Report
when you have two modals on top of each other, if the height of both of them are the same and close buttons are out of the modal, the close buttons are on each other which has a bad user experience.

## Steps to reproduce

1. make the scope (browser window screen) at a big size so that the close buttons of the modal come outside of the modal scope
2. open the background and foreground modal

## Expected result
a better UI for the close buttons which are on top of each other

## Actual result
the  close buttons of both modals fall on top of each other (based on the modals sizes and headers height)

## Testcase
https://jsfiddle.net/208c45nt/1

## Screenshot (if possible)
image

## Version
2.8.8

lubber-de:
## Description
when you have two modals on top of each other, if the height of both of them are the same and close buttons are out of the modal, the close buttons are on each other which has a bad user experience.

In addition the close icon from the first modal was still clickable and would close the first modal.

In some cases the close icon could be shown over a different inner Dimmer making the close icon invisible, so i added a slight text shadow to keep it visible. This text shadow is not recognizable on usual modal usage .

At last i made the dimmer class of the inner dimmer optional, so one could customize that as well.

## Testcase
- Have a wide vieport (> tablet), so the close icon of a modal will apear outside the modal to the upper right
- Open modal via button 
- open another modal via the second button inside the first modal

Remove CSS to see the issue (overlapping close icons)
https://jsfiddle.net/lubber/dm7pj3gf/

### Before
image

### After
image

## Closes
#2195

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -216,8 +216,8 @@ $.fn.modal = function(parameters) {
             module.verbose('Creating unique id for element', id);
           },
           innerDimmer: function() {
-            if ( $module.find(selector.dimmer).length == 0 ) {
-              $module.prepend('<div class="ui inverted dimmer"></div>');
+            if ( $module.find(selector.dimmer).length === 0 ) {
+              $('<div/>', {class: className.innerDimmer}).prependTo($module);
             }
           }
         },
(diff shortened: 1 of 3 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/319

exoego (approved):
LGTM

PR #2197, opened 2022-01-07
feat(accordion): pure css, inverted, basic, right, tree, nested fixes

lubber-de:
## Description
This MR enhances the accordion module:

- Pure CSS variant now possible using the[ details/summary tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) (yes, IE/old Edge do not support that, that's ok for 2.9.0 and 2022 and you can still use the JS version which works there) 😉 . 
- `inverted` styled variant
- `tree` variant (for proper indentation which a tree view needs) (also works as CSS only!)
- new `right dropdown icon` variant, to be able to place the dropdown icon to the right (just as in accordion menu already). Usefull for example when used inside `message`
- fixed the right dropdown icon (also in accordion menu)
- fixed possible usage of `title` or `content` elements inside an accordion which are supposed to be parts of different elements
- `basic styled` variant which is equal to styled for title handling only (so no borders and always transparent background)

## Testcase
https://jsfiddle.net/lubber/4k5y0cjL/38/

## Screenshots
accordiontree

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/2790
https://github.com/Semantic-Org/Semantic-UI/issues/4928
https://github.com/Semantic-Org/Semantic-UI/issues/5238
https://github.com/Semantic-Org/Semantic-UI/issues/5330
https://github.com/Semantic-Org/Semantic-UI/issues/6578
https://github.com/Semantic-Org/Semantic-UI/issues/7048

### DIFF
--- a/src/definitions/modules/accordion.less
+++ b/src/definitions/modules/accordion.less
@@ -31,18 +31,16 @@
 }
 
 /* Title */
-.ui.accordion .title,
-.ui.accordion .accordion .title {
+.ui.accordion.menu .item > .title,
+.ui.accordion > .title,
+.ui.accordion .accordion > .title {
   cursor: pointer;
-}
-
-/* Default Styling */
-.ui.accordion .title {
   padding: @titlePadding;
   font-family: @titleFont;
   font-size: @titleFontSize;
   color: @titleColor;
   list-style: none;
+  line-height: @titleLineHeight;
 }
 
 /* Default Styling */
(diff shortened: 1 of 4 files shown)


exoego (approved):
This is a huge enhancement to Accordion module 🥇

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/334

PR #2198, opened 2022-01-07
feat(toast): allow clickable items without closing

lubber-de:
## Description
Some content inside toasts should remain clickable without closing the toast. (even if `closeOnClick` is true)
While this was already hardcoded for `a` tags if fails for possible other tags like `accordion` or `details` (Especially when used together with the new #2197 😉 )
The previously hardcoded selector is not a customizable classname setting.

Btw: `input` is not needed to be set here, because any kind of input inside a toast will already set the whole toast to not closing when clicked (because it is expected to be clicked into the input fields).
The new setting here is supposed for elements which _can_ be clicked, but are not expected to.

## Testcase
https://jsfiddle.net/lubber/fd8Lx9u2/24/
> This also shows how to nicely use the new `ui basic inverted styled accordion` inside the toast

## Screenshot
clickabletoast

### DIFF
--- a/src/definitions/modules/toast.js
+++ b/src/definitions/modules/toast.js
@@ -511,7 +511,7 @@ $.fn.toast = function(parameters) {
             module.close();
           },
           click: function(event) {
-            if($(event.target).closest('a').length === 0) {
+            if($(event.target).closest(selector.clickable).length === 0) {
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/338

PR #2204, opened 2022-01-17
fix(dropdown,search): autocomplete affects all chrome based browser

### LINKED ISSUE #2203: [Dropdown] module.is.chrome() is returning false when should be true
# Bug Report
this line: https://github.com/fomantic/Fomantic-UI/blob/31d0f3ffde8f10b294ab2a81dc5d2e70ab76ecc4/src/definitions/modules/dropdown.js#L3400

module.is.chrome return false because window.chrome.webstore and window.chrome.runtime doesn't exist (anymore?)
Im on Chrome Version 97.0.4692.71

## Steps to reproduce
1. create a tag with `<div class="ui fluid search selection dropdown">`
2. run in chrome
3.

## Expected result
result should be autocomplete="fomantic-search"

## Actual result
because is.chrome isn't detected. autocomplete is set to "off" and display the chrome autocomplete

## Screenshot (if possible)
image
image

## Version

2.9.0-beta.162

Thank you!

lubber-de:
## Description
The autocomplete bug fixed in #1887 affects all chromium based browsers, but the `is.chrome` check only checks for a real chrome browser. Even worse when using chromes incognito mode the chrome check fails, because window.chrome.runtime does not exists

## Closes
#2203


lubber-de:
@fomantic/maintainers  Please review if you find time. Thx 🙂

prudho (approved):

src/definitions/modules/dropdown.js
```diff
@@ -3397,7 +3397,7 @@ $.fn.dropdown = function(parameters) {
             return $(event.target).closest($icon).length > 0;
           },
           chrome: function() {
-            return !!window.chrome && (!!window.chrome.webstore || !!window.chrome.runtime);
+            return !!window.chrome && !window.StyleMedia;
```
exoego, line 3400: `VM180:1 Uncaught ReferenceError: StyleMedia is not defined
    at <anonymous>:1:1` for my chrome.
Perhaps `window.styleMedia` ?

lubber-de: No, because we explicitely check for old Edge here (which also has `window.chrome` but isn't chrome!)
The check for Uppercase `StyleMedia` is only available in old Edge (while lowercase is available on all modern browsers)
### Old Edge
image

exoego (approved):

PR #2205, opened 2022-01-19
feat(chore): automatic npm releases, ci fix and add node 16

lubber-de:
## Description
I would like to automate the npm publishing of new releases, so this PR adds another github workflow which publishes to npm ONLY when a new github release is published.

@hammy2899 
#### As you are the direct owner of the npm package (we dont have an npm org account) and to reduce the need to give me an NPM auth token to myself, this workflow uses the already existing NPM AUTH secret (used for nightly builds), so i dont have to know that 😉 

I would like to setup this release workflow for the Fomantic-UI-CSS and Fomantic-UI-LESS repos as well, so could you please upload a/the NPM secret to the fomantic-organization as well (and/or to the respective single repositories)
Otherwise i always have to bother you to create the releases manually yourself 😉 

I intentionally did not implement a version-adjustment inside package.json within the workflow script to avoid publishing wrong releases in case one of us does something wrong while adding tags . So unless we manually change the version number inside package json a possible npm publish will fail (because the old version already exists on npm)

This PR also fixes the CI script, because the node version was _always_ the latest LTS (16) because of a typo in the yml (you could check the CI logs to confirm. So at least this means FUI build well with node16/npm 8 :D (which i also tested manually)

That said, i also increased the node version for the workflows to 14 (not 16, because that is active and could have impact by changes) as node 12 becomes unsupported in april this year

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,15 +2,15 @@ name: CI
 on: [pull_request]
 jobs:
   test:
-    name: Test build process on node ${{ matrix.node_version }}
+    name: Test build process on node ${{ matrix.node-version }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node_version: [8, 10, 11, 12, 13, 14]
+        node-version: [12, 14, 16]
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
         node-version: ${{ matrix.node-version }}
     - name: pre-install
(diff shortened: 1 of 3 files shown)


hammy2899:
Checked it over and all seems good.

The releases are actually published using the [fomantic-bot](https://www.npmjs.com/~fomantic-bot) npm account not my account. I will actually send you the details on Discord for this account just as a precaution. Also if you want to send me your npm username I can add you too as it provides a fallback for whatever reasons in the future :)

hammy2899 (approved):

exoego (approved):
Great !!

hammy2899:
New token added to the org `NPM_AUTOMATION`, lets switch over this PR to use it so we can get rid of the other one and then also enable 2fa on the fomantic-bot npm account

hammy2899 (approved):
Sweet, LGTM

PR #2209, opened 2022-01-22
feat(build): group components in checkbox selection

### LINKED ISSUE #2180: [installation] While installing and choosing components to include, show the list in alphabetical order
I am currently installing the fomantic on the command line via "npx gulp install" and I'm selecting/unselecting components. The list is in a random order. It would be a lot nicer if this component list was in alphabetical order.

Screenshot at 2021-12-14 10-53-02

lubber-de:
## Description
This PR adds a Grouping Header to the component selection at install time.

It's more a hack by using disabled entries, but works out good enough 😄 

I, for now, decided against renaming the form.js to formvalidation.js because i fear a too big breaking change and most probably people who want form elements also want the validation anyway.
I could have just changed the checkbox selection name to "formvalidation" but effectively the script would have continued with "form" anyway which probably creates more confusion.

## Testcase
- run `npx gulp install`
- select "Yes, extend my settings"
- select "express" or "custom"
- see the checkbox components list are now grouped by unselectable "headers"

## Screenshot
image
image
image
image
image

## Closes
#2180

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -922,6 +922,11 @@
         }
       }
     },
+    "base64-js": {
+      "version": "1.5.1",
+      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
+      "integrity": "sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA=="
+    },
     "beeper": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/beeper/-/beeper-1.1.1.tgz",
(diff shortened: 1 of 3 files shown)


hugopeek:
Hacky or not, I like it! :)

>I, for now, decided against renaming the form.js to formvalidation.js because i fear a too big breaking change and most probably people who want form elements also want the validation anyway.

I understand the breaking change part. Propably not worth it.

But please don't assume that people using form elements will want validation too. One developer I know uses SUI for the HTML/CSS, but ships it with his own vanilla JS. I often rely purely on backend validation. So form elements and validation are definitely 2 separate things.

>I could have just changed the checkbox selection name to "formvalidation" but effectively the script would have continued with "form" anyway which probably creates more confusion.

You're right, thanks for not doing that ;)

The double `form` component is an odd ambiguity from SUI. Can imagine if you want to leave it as is for now, but maybe keep it in the backlog of a possible v3?

lubber-de:
> But please don't assume that people using form elements will want validation too. One developer I know uses SUI for the HTML/CSS, but ships it with his own vanilla JS. I often rely purely on backend validation. So form elements and validation are definitely 2 separate things.
> 
Totally agree! Backend validation ist basically the main point to implement any data validation. 
TBH, i was about to rename it, but also saw too much of needed changes (docs...)

I also thought about making it backward compatible somehow... but was too lazy to do all the needed implementation....for now.

> The double `form` component is an odd ambiguity from SUI. Can imagine if you want to leave it as is for now, but maybe keep it in the backlog of a possible v3?

Seems legit, yes

exoego (approved):
LGTM:+1:

PR #2211, opened 2022-01-23
fix(form): make sure validation object is cloned

lubber-de:
## Description
Whenever one has more than 2 forms on the page and also sets defaults to the global jquery forms object, any changes to the fields/rules afterwards were adopted to the _global_ validation object instead of the dedicated one from the related form affecting _all_ forms.

This is testable on the docs page

## Testcase
https://fomantic-ui.com/behaviors/form.html

The first form "Tell us about yourself" did not have rules for gender and password anymore, although they are only removed for the https://fomantic-ui.com/behaviors/form.html#adding-rules-programmatically example only.
Whe you play with the "Add additional validation" and "Remove additional validation" buttons, it on the second mentioned example, this affect also all the other forms (whose have gender and/or password fields)

When applying the fix on the docs, each validation is dedicated to each form as it should.

This ia basically caused by jquery not seeming to doing a clone copy of a prototype default when `extend` is called without a deepcopy. Using `extend` also with deepcopy option solved it in the end.

#### Before
image

### After
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/7074


prudho (approved):
LGTM

PR #2212, opened 2022-01-23
fix(modal): close text shadow appeared too early

lubber-de:
## Description
I accidently used the wrong breakpoint for #2196 which now showed the text shadow for the close icon too early (while inside the modal where is is not supposed to be used).

## Testcase
- Have the viewport between 768px and 992px
- Open the modal

### Broken
See ugly close text shadow inside modal
https://jsfiddle.net/lubber/dm7pj3gf/8/

### Fixed
All fine as intended
https://jsfiddle.net/lubber/dm7pj3gf/6/

### Broken
image

### Fixed
image

## Fixes
#2196


lubber-de:
@fomantic/maintainers  Please review if you find time. Thx 🙂

exoego (approved):
I am not able to notice shadow both on fixed and broken, but code LGTM

PR #2214, opened 2022-01-23
fix(dependencies): make browserslists an own dependecy

### LINKED ISSUE #2206: [build] Missing dependency on browserslist
The package is using "browserslist" (tasks/config/tasks.js).

However, it does not (directly) depend on "browserslist". Instead, browserslist is just pulled in via "gulp-autoprefixer", which works, but is invalid for yarn>=2.

lubber-de:
## Description
The dependency "browserslist"  is used in /tasks/config/tasks.js.

However, it does not (directly) depend on "browserslist". Instead, browserslist is just pulled in via "gulp-autoprefixer"

This PR adds the dependency directly

## Closes
#2206


exoego (approved):
LGTM

PR #2217, opened 2022-01-30
[Button]] wrapping and spaced variant for button Groups

### LINKED ISSUE #2157: [Buttons] Wrapping of buttons inside a group
This request stems from [a bug report in Semantic UI's repo](https://github.com/Semantic-Org/Semantic-UI/issues/7069).

# Feature Request

I've noticed that when using button groups inside a grid cell, if there are enough buttons (or enough text in them), rather than wrapping to the next line, they will overflow adjoining cells.

## Example

The effect of this with can be seen with Fomantic UI in this JSFiddle: https://jsfiddle.net/36x2csbt/ .

## Screenshot (if possible)

image

## Possible solution

In the can of "normal" buttons the solution is simply to use:

```
.ui.buttons {
  flex-wrap: wrap;
}
```

Example: https://jsfiddle.net/36x2csbt/1/

However, for `basic` buttons, I don't have quite such a simple solution. basic buttons use an rgba border, and the top / bottom border comes from the container. So wrapping the cells means we need a top / bottom border in order to separate them which complicates the CSS a bit:

```
.ui.buttons {
  flex-wrap: wrap;
}

.ui.basic.buttons {
  border-bottom: none;
}

.ui.basic.buttons .ui.button {
  border-bottom: 1px solid rgba(34,36,38,.15);
  margin-bottom: -1px;
}

.ui.basic.buttons .ui.button:hover {
  background: transparent !important;
}
```

But it does appear to be fairly effective: https://jsfiddle.net/36x2csbt/3/ .

lubber-de:
## Description
This PR adds new button group variants `wrapping` and `spaced`
There also exists another optional `wrapped` helper class which should be set when it's known if a button group is actually wrapped (this will adjust some border-radius values). It is not possible to detect a wrapping child automatically via pure CSS, that's why the developer should handle this using that helper class whenever needed.

The PR also fixes some adjustments when used together with `basic` and/or `inverted` and improved conditional compilation at some places

## Testcase
https://jsfiddle.net/lubber/akb4wh2n/

## Screenshots
image
image

## Closes
#2157 
https://github.com/Semantic-Org/Semantic-UI/issues/7069

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -262,61 +262,64 @@
     right: 0;
   }
 
-  /* Vertical */
-  .ui.vertical.animated.button .visible.content,
-  .ui.vertical.animated.button .hidden.content {
-    transition: top @animationDuration @animationEasing, transform @animationDuration @animationEasing;
+  & when (@variationButtonVertical) {
+    /* Vertical */
+    .ui.vertical.animated.button .visible.content,
+    .ui.vertical.animated.button .hidden.content {
+      transition: top @animationDuration @animationEasing, transform @animationDuration @animationEasing;
+    }
+    .ui.vertical.animated.button .visible.content {
+      transform: translateY(0%);
+      right: auto;
+    }
+    .ui.vertical.animated.button .hidden.content {
+      top: -50%;
+      left: 0;
+      right: auto;
+    }
+    .ui.vertical.animated.button:focus .visible.content,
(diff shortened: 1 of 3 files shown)


exoego (approved):
LGTM 👍

prudho (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/340

PR #2218, opened 2022-01-30
feat(calendar): multimonth, parser & modal usage fix

### LINKED ISSUE #2200: calendar date setting error when monthFirst: true
# Bug Report

With option "monthFirst" set to true, there seems to be a parsing problem when setting calendar value.

## Steps to reproduce

Try setting calendar value to '10. 6. 2021' (with spaces) and to '10.6.2021' (without spaces).

## Expected result

value == 2021-06-10

## Actual result

In the first case with spaces included the value is '2021-10-06' instead of '2021-06-10'. In the second case without spaces the result is OK.

## Version

2.8

lubber-de:
## Description
Several small fixes:
- Multimonth calendar offset was only possible by providing it via data attributes `data-month-offset`. This is now also possible by using the `monthOffset` value as well.
- This PR also fixes date parsing in case whitespace is part of the date.
- By disabling the observer in the popupoptions by default the calendar works properly when used inside modals
- multimonth calendars now appear in the same way inside popup as a single calendar

### Fixed
https://jsfiddle.net/lubber/mo97gk6y/12/

### Broken
https://jsfiddle.net/lubber/mo97gk6y/13/

## Closes
#2200 
https://github.com/mdehoog/Semantic-UI-Calendar/issues/99

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -746,7 +746,7 @@ $.fn.calendar = function(parameters) {
             return $module.data(metadata.maxDate) || null;
           },
           monthOffset: function () {
-            return $module.data(metadata.monthOffset) || 0;
+            return $module.data(metadata.monthOffset) || settings.monthOffset || 0;
           },
           mode: function () {
             //only returns valid modes for the current settings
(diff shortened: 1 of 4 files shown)


exoego (approved):
LGTM

prudho (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/336

PR #2219, opened 2022-01-31
fix(dropdown): fix simple upward variant

lubber-de:
## Description
The `simple dropdown` css only variant was broken when trying to use it with a `upward` menu

- menu was cut on smaller screens
- dropdown itself was overlayed by menu
- simple selection variant had wrong borders

#### Fixed
https://jsfiddle.net/lubber/cz09kaue/1/

#### Broken
https://jsfiddle.net/lubber/cz09kaue/3/

## Screenshot
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6288


exoego (approved):
LGTM

PR #2226, opened 2022-02-05
feat(docs): limit/clarify IE, old edge, old mobile support

lubber-de:
## Description
Starting from v2.9.0 i would like to reduce IE11/old Edge and old mobile platform support.

Fomantic will still work in those browsers in general, but in case there are dedicated bugs, we will not spent time into trying to fix them for those ancient browsers/platforms anymore. 

## Screenshots
As i used markdown footnotes for detailed explanation such footnotes are found at the end of the readme as follows:

* Last 2 Versions Firefox, Chrome, Safari Mac, Edge
* Last 4 Versions of Android, Chrome for Android, iOS Safari[^1]
* IE 11[^2]
* Microsoft Edge 12-44[^2]

[^1]: Fomantic-UI should basically still work in iOS Safari 7+, Android 4.4+, but, starting from v2.9.0, we won't support them anymore if anything works different than in recent versions.
[^2]: Fomantic-UI should basically still work in IE11 / old Edge, but, starting from v2.9.0, we won't support them anymore in terms of dedicated bugfixes.

> here comes lots of other readme text ...

> until the readme reaches the end

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Fomantic includes an interactive installer to help setup your project.
 * IE 11[^2]
 * Microsoft Edge 12-44[^2]
 
-[^1]: Fomantic-UI should basically still work in iOS Safari 7+, Android 4.4+, but, starting from v2.9.0, we won't support them anymore if anything works different as the recent versions.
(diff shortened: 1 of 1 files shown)


exoego (approved):
Ah, finally. Time flies 😌 

LGTM! 
I think we do not use semantic versioning, dropping browser is fine in 2.x

PR #2230, opened 2022-02-10
feat(api): dont convert array response to object by default

### LINKED ISSUE #2220: [api] Array API repsonse parsed as plain object in onResponse()
# Bug Report

When API returns a "bare" Array JSON response, Fomantic parses it as plain object instead of an `Array`.

Related: https://github.com/fomantic/Fomantic-UI/issues/263

## Steps to reproduce

**Step 1**

API endpoint returns, say `[{"id":1, "desc":"Sun"}, {"id":2, "desc":"Moon"}]`

**Step 2**

```html
<div class="ui search selection dropdown" id="myDropdown">
  <div class="default text">Select Celestial Body</div>
</div>
```

```js
$('#myDropdown').dropdown({ 
  apiSettings: {
    cache: false,
    url: "/celestialBodies",
    mockResponse: [{"id":1, "desc":"Sun"}, {"id":2, "desc":"Moon"}],
    onResponse: (data) => {
      console.log(data); // Plain object!
      return { results: data.map((item) => ({ name: item.desc, value: item.id })) } // ERROR!
      // Uncaught TypeError: data.map is not a function
    },
  }
});
````

## Expected result

Ideally `onResponse()` exposes a parsed `Array` object

`[{"id":1, "desc":"Sun"}, {"id":2, "desc":"Moon"}]`

## Actual result

Instead the object is a plain JS object that is not iterable.

`{"0":{"id":1, "desc":"Sun"}, "1":{"id":2, "desc":"Moon"}}`

## Testcase

https://jsfiddle.net/jchook/4xc5v0b3/15/

## Screenshot (if possible)

image

## Version

2.8.8

---

## Workaround

```js
$('#myDropdown').dropdown({ 
  apiSettings: {
    cache: false,
    url: "/celestialBodies",
    mockResponse: [{"id":1, "desc":"Sun"}, {"id":2, "desc":"Moon"}],
    onResponse: (data) => {
      const results = []
      for (const key in data) {
        if (data.hasOwnProperty(key)) {
          const item = data[key]
          results.push({ name: item.desc, value: item.id })
        }
      }
      return { results }
    },
  }
});
```

lubber-de:
## Description
It seems a more common case people are sending JSON array instead of JSON objects for their api response.
We already supported that since 2.7.0 #294 by the `rawResponse` setting, but people have to actively enable it.

This PR changes the default value so json arrays are not converted to objects anymore by default.

I am still against always forcing json arrays no to be converted, because it is a valid conversion by `$.extend` and probably still needed for some projects.

## Closes
#2220

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -1110,7 +1110,7 @@ $.api.settings = {
   responseAsync     : false,
 
 // whether onResponse should work with response value without force converting into an object
-  rawResponse       : false,
+  rawResponse       : true,
 
   // callbacks before request
   beforeSend  : function(settings) { return settings; },


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/329

exoego (approved):
LGTM

PR #2239, opened 2022-02-17
fix(chore): replace deprecated .substr() with .slice()

CommanderRoot:
## Description
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with the [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) method which works similarily but isn't deprecated.
.substr() probably isn't going away any time soon but as the change is trivial it probably doesn't hurt to switch.

### DIFF
--- a/examples/assets/library/iframe-content.js
+++ b/examples/assets/library/iframe-content.js
@@ -185,7 +185,7 @@
       return 'true' === str ? true : false;
     }
 
-    var data = initMsg.substr(msgIdLen).split(':');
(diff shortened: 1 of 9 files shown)


lubber-de (approved):
Thanks, LGTM

lubber-de:
@all-contributors please add @CommanderRoot  for code

Issue #2246, opened 2022-02-25
[dropdown] is corrupting data (/w multiple enabled & allowAddition)


mvorisek:
# Bug Report

## Steps to reproduce

https://jsfiddle.net/e8o3psvr/

## Expected result

```
[{
  name: "recipient",
  value: "Username &lt;user@emaildomain.de&gt"
}]
[{
  name: "recipient",
  value: "Username &lt;user@emaildomain.de&gt"
}]
[{
  name: "recipient",
  value: "Username Xuser@emaildomain.deX"
}]
```

## Actual result

```
[{
  name: "recipient",
  value: "Username &lt;user@emaildomain.de&gt;,Username &amp;lt;user@emaildomain.de&amp;gt;"
}]
[{
  name: "recipient",
  value: "Username &lt;user@emaildomain.de&gt;,Username &amp;lt;user@emaildomain.de&amp;gt;"
}]
[{
  name: "recipient",
  value: "Username Xuser@emaildomain.deX"
}]
```

## Version
2.8.8

Also and probably related, discovered in https://github.com/fomantic/Fomantic-UI/pull/2247#issuecomment-1052781930, `xx<yy` input produces very strange result.

mvorisek:
What is the reason of this behaviour and is there any workaround?

mvorisek:
Is this a bug or is the value expected to be encoded twice, eg. once for the value attribute and once for each listed value?

mvorisek:
This issue is quite dangerous as it corrupts data. Can it be fixed soon and is there any easy fix from JS when using the latest v2.8.8 libs?

lubber-de:
I am working on it already. PR is on its way. The only workaround is to already htmlentity encode the predefined values.

mkrecek234:
@lubber-de Any news on a fix for this issue? Would be welcomed a lot. Thank you

mkrecek234:
@lubber-de We are approaching our next release using Fomantic-UI - would be great if you could fix this issue? Any chance?

lubber-de:
This is the last remaining bug for 2.9.0, so....yes... 😉

lubber-de:
Fixed by #2474 
See your adjusted jsfiddle here https://jsfiddle.net/lubber/61u0xpL8/5
In your use case you now have to use `preserveHTML:false` to make the < > being rendered inside the label. However the values are not encoded and not added twice anymore

PR #2247, opened 2022-02-26
fix(dropdown): use central delimiter char instead of key

### LINKED ISSUE #2245: [dropdown] `allowAdditions` (with multiple values allowed) breaks the "words" on some characters
# Bug Report

## Steps to reproduce
1. create dropdown with multiple values allowed and `allowAdditions` option
2. focus it and write `test?bar` or `test<bar`

## Expected result

one `test?bar` (or `test<bar`) value (after hitting enter) is expected

image

## Actual result

as soon as `?` (or `<`) key is pressed, the previous value is finished, the `?` (or `<`) key is ignored, and a new value is started

this is problematic to enter values with such characters, workaround is to write such value in a text editor and copy it to the component in one shot instead

image

maybe more non alphanum characters are affected, but the enter/finish characters should be configureable (probably TAB and ENTER by default)

## Version
2.8.8

lubber-de:
## Description
On a multiple dropdown, having `allowAdditions:true`, the press of the delimiter key (comma by default) was checked via the keycode. However the keycode does not take possible pressed shift/alt keys into account.
This results in triggering a multi dropdown entry even when pressing the key together with shift/alt, so it was not possible to declare for example the semicolon as a delimiter.

As we already have the delimiter string for concatenation/splitting of existing values, this PR now also uses that setting for the keypress.

### Broken
https://jsfiddle.net/lubber/27whu5tf/9/

### Fixed
https://jsfiddle.net/lubber/27whu5tf/11/

## Closes
#2245 
https://github.com/Semantic-Org/Semantic-UI/issues/3016
https://github.com/Semantic-Org/Semantic-UI/issues/3318

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1531,7 +1531,7 @@ $.fn.dropdown = function(parameters) {
           keydown: function(event) {
             var
               pressedKey    = event.which,
-              isShortcutKey = module.is.inObject(pressedKey, keys)
+              isShortcutKey = module.is.inObject(pressedKey, keys) || event.key === settings.delimiter
             ;
             if(isShortcutKey) {
               var
@@ -1549,7 +1549,7 @@ $.fn.dropdown = function(parameters) {
                 hasSubMenu            = ($subMenu.length> 0),
                 hasSelectedItem       = ($selectedItem.length > 0),
                 selectedIsSelectable  = ($selectedItem.not(selector.unselectable).length > 0),
-                delimiterPressed      = (pressedKey == keys.delimiter && module.is.multiple()),
+                delimiterPressed      = (event.key === settings.delimiter && module.is.multiple()),
                 isAdditionWithoutMenu = (settings.allowAdditions && settings.hideAdditions && (pressedKey == keys.enter || delimiterPressed) && selectedIsSelectable),
                 $nextItem,
                 isSubMenuItem,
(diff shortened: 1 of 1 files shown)


mvorisek:
https://jsfiddle.net/lubber/27whu5tf/12/

when I paste `xs,xs`

image

the value with comma cannot be longer inserted via clipboard (via keyboard, `,` is expected to finish the previous item)

with broken https://jsfiddle.net/lubber/27whu5tf/9/ it was possible

maybe related with https://github.com/fomantic/Fomantic-UI/issues/2246

also when I type `xx<yy` (or copy paste) the value changes to two items:

image

and the 1st item cannot be even delete

lubber-de:
> [jsfiddle.net/lubber/27whu5tf/12](https://jsfiddle.net/lubber/27whu5tf/12/)
> 
> when I paste `xs,xs`

This is intended, because the fix is based on the latest develop branch and the behavior was adjusted by #2167 to take care of existing delimiters inside pasted strings.
The comma is the delimiter and as you entered 2 identical values it reduces that to only one `xs` values, thus the behavior is correct now.

> the value with comma cannot be longer inserted via clipboard (via keyboard, `,` is expected to finish the previous item)

Yes, see above this is intended and logical

> with broken [jsfiddle.net/lubber/27whu5tf/9](https://jsfiddle.net/lubber/27whu5tf/9/) it was possible
> 
> maybe related with #2246
> 
> also when I type `xx<yy` (or copy paste) the value changes to two items:

This is related to #2246 and will be fixed in a separate PR

exoego (approved):
LGTM

PR #2252, opened 2022-03-06
fix(input): date inputs were larger due to icon size

lubber-de:
## Description
Date or other new input types used in chrome based browsers are showing an additional icon to trigger the value seletcion.
Unfortunately the icons had an additional padding making the inputs appear larger than normal inputs which interferes with the overall ui when such inputs were shown next to each other.
I also adjusted the appearance of the icon to match the `icon input` appearance and added the pointer to the icon as such native icon has a bahaviour to show a dialog.

### Broken
The Email-Input to the very right has a smaller height than the others
image

### Fixed
Same height everywhere
image

## Testcase
Remove the CSS to see the issue
https://jsfiddle.net/lubber/1umaonxg/1/

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -143,6 +143,9 @@
 
 .ui.form input::-webkit-calendar-picker-indicator {
   padding: 0;
+  opacity: @iconOpacity;
+  transition: @iconTransition;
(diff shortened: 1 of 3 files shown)


prudho (approved):
LGTM

PR #2253, opened 2022-03-06
fix(segment): adjust inverted attached border color specificity

lubber-de:
## Description
Inverted attached segments got wrong border color (from non inverted variant). A slight specificity increase solves this.

## Testcase
Remove CSS to see issue
http://jsfiddle.net/lubber/bxtmyv9z/46/

### Broken
image

### Fixed
image


exoego (approved):
LGTM

prudho (approved):

PR #2256, opened 2022-03-07
[Container] Wide variation

### LINKED ISSUE #2231: [container] Add wider container variations
# Feature Request

It would be nice to have larger variations of `ui container`, in addition to the `fluid` one, for example a `ui wide container`.
I find the maximum container width (computed by default from the `@largeMonitorBreakpoint` value) to be a little bit too narrow in most cases, considering that nowadays almost all desktop monitors are (at least) 1920px wide.

prudho:
## Description
This PR implements a `wide` class for containers.

It works by applying a `wideRatio` (default `1.2`) to the actual container width, except for mobile screens.

## Testcase
[JSFiddle](https://jsfiddle.net/cvebk0f7/) (remove CSS to see old behavior)

## Screenshot
Animation

## Closes
#2231

### DIFF
--- a/src/definitions/elements/container.less
+++ b/src/definitions/elements/container.less
@@ -137,6 +137,25 @@
   }
 }
 
+& when (@variationContainerWide) {
+  /* Wide Container */
+  @media only screen and (min-width: @tabletBreakpoint) and (max-width: @largestTabletScreen) {
+    .ui.ui.ui.wide.container {
+      width: @tabletWideWidth;
+    }
+  }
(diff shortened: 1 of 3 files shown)


exoego (approved):
Maybe `very wide container` variant?
LGTM

lubber-de:
I leave that up to @prudho if he wants to add that to this PR. Otherwise I merge it and we could add it later if desired

prudho:
With the actual ratio, having a `very wide` variant would only be useful on very large screens (at least WQHD i think). For now we can merge it as it, and see what will be the feedback about it.

PR #2257, opened 2022-03-07
feat(api): serializeObject, formdata, files

### LINKED ISSUE #2113: [api] file input/upload support
# Feature Request

Current `serializeForm` setting/feature does not support file inputs (It uses jQuery serializeArray).
What about adding new `useFormData`/`useFormDataApi` setting? If it's set to `true` then ajax data can be generated by using [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData).

Instead of `useFormData` parameter it can be automatically detected by checking `enctype` form attribute. But user can also use submitting button `formenctype` attribute instead of form attribute. It might be tricky to detect.

Current workaround is probably generating `settings.data` with `FormData` API in `beforeSend`. (Not tested, it may also require not to pass `serializeForm` setting, if it's set, generated data will probably corrupted by `module.add.formData` method).

## Example
```js
$('.submit').api({
    action: 'upload file',
    serializeForm: true,
    useFormData: true
});
```

It can be added to:
https://github.com/fomantic/Fomantic-UI/blob/d11687dde71e26847e6d014246813b3e856a626a/src/definitions/behaviors/api.js#L198-L201

Theoretical/abstract imlementation:

```js
          if (settings.serializeForm) {
            if (settings.useFormData) {
              var formData = new FormData($form.get(0));
              // todo: append additional user passed data comes from settings.data
              settings.data = formData;
              // Need to set processData and contentType false
              // See: https://stackoverflow.com/questions/5392344/sending-multipart-formdata-with-jquery-ajax/5976031#5976031
              settings.processData = false;
              settings.contentType = false;
            }
            else {
              settings.data = module.add.formData(settings.data);
            }
          }
```

lubber-de:
## Description
This PR implements an optimized version of the original jquery-serializeobject [dependency of SUI](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/api.js#L425) when `serializeForm:true` was used.
 
https://github.com/macek/jquery-serialize-object/blob/master/jquery.serialize-object.js 
I enhanced the original logic to also support file inputs, convert pure number values and support same name keys (without trailing [])

I also added support to use The FormData Api when `serializeForm` is set to `'formdata'`. This way the old code stays backward compatible.

## Testcase
https://jsfiddle.net/lubber/qLu6vrty/4/

## Screenshots
The "FUI new" result covers all form inputs , while the "FUI" (old approach only in 2.9.0 beta) ignores key naming logic and the original "macek" (which is serializeObject used by SUI and FUI <= 2.8.8" does not handle files or same name keys.
image

## Closes
#2113


prudho (approved):
LGTM.

Great addition !

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/351

PR #2259, opened 2022-03-08
fix(form): rebind events on dynamic fields

### LINKED ISSUE #2035: [Form Validation] OnChange on Dynamic Error Field not Working.
# Bug Report
Dynamic error field should behave like the normal field. On dynamic field the ```error``` will not go away after user started to type something in the input, it only happens on the ```onblur```.

## Testcase
https://jsfiddle.net/lubber/0wyz2g7e/
_Add a field dynamically, then use ```submit``` button to make the form invalid. Type on each field to see the different_ 

## Version
2.8.8

Follow up on: https://github.com/Semantic-Org/Semantic-UI/issues/7030
According to @lubber-de  it is a known issue. If anyone has a work around method, please let me know.
Thanks,

lubber-de:
## Description
Whenever a form field is dynamically added to the form and added/removed via `add field` or `remove field` to add/remove related rules, the event listeners to such new fields wre not renewed resulting in not correctly validating the  new field at runtime.

## Testcase
- Click "add val2 field" to add a field dynamically to the already existing form
- Use submit button to make the form invalid.
- Type anything into each field to see the different
### Broken
- The first field (which was already there) gets validated (error class is removed) whenever you type anything into the field
- The second field stays invalid (error class) while typing and only gets valid on blur
https://jsfiddle.net/lubber/0wyz2g7e/15/
### Fixed
- Both fields get validated while typing
https://jsfiddle.net/lubber/0wyz2g7e/18/

## Closes
#2035

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -119,7 +119,7 @@ $.fn.form = function(parameters) {
           ;
         },
 
-        refresh: function() {
+        refresh: function(rebindEvents) {
           module.verbose('Refreshing selector cache');
           $field      = $module.find(selector.field);
           $group      = $module.find(selector.group);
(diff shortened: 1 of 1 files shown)


prudho (approved):
LGTM

PR #1739, opened 2020-11-01
[Grid] stackable grid had wrong width inside popup

lubber-de:
## Description
A `stackable grid` inside a popup did not have enough specificity to adjust the width on mobile views

## Testcase
- Click on the red bar to show the popup 
- Adjust screen to mobile size
- Watch the blue background
- Remove the CSS to see issue
https://jsfiddle.net/lubber/fv1hg0jm/5/

## Screenshot
gridinsidepopup

PR #2269, opened 2022-03-16
fix(grid): fix stackable nested grid

### LINKED ISSUE #1907: [Grid] stackable grid has extra left/right margin when stacked
# Bug Report

Stackable grid has extra left/right margin when stacked. See image below how the left margin of column 1 becomes wider when the grid is stacked.

## Steps to reproduce

```html
<div class="ui basic segment">
  <div class="ui two column stackable grid">
    <div class="column">
      <div class="ui segment">
        Column 1
      </div>
    </div>
    <div class="column">
      <div class="ui segment">
        Column 2
      </div>
    </div>
  </div>
</div>
```

## Expected result
No extra margin when stacked. i.e. same amount of margin in both stacked and unstacked state

## Actual result
Stackable grid has extra left/right margin when stacked.

## Testcase
https://jsfiddle.net/04rj9sx2/2/

## Screenshot (if possible)

Screenshot 2021-03-05 at 11 42 40 AM
Screenshot 2021-03-05 at 11 42 49 AM

## Version

2.8.7

### LINKED ISSUE #2263: [grid] double gutter in nested stackable grids
# Bug Report

Stackable grids inside a parent grid receive a negative left/right margin, in order to align properly with the parent grid.

This negative margin is overruled however, by the changes in this PR: https://github.com/fomantic/Fomantic-UI/pull/1739/files

The specificity is now too high, so the !important values of the negative margins are no longer applied.

## Steps to reproduce

See the following fiddle: https://jsfiddle.net/gtc4o5zr/3/

https://github.com/fomantic/Fomantic-UI/blob/e4f06996233cba1b901def23e66025c18b9776c5/src/definitions/collections/grid.less#L1692-L1696

If you disable the negative margins here, or remove the added .ui classes, things go back to normal.

## Expected result

image

There should be no gutter between the parent and inner grid.

## Actual result

image

There is another gutter between the parent and inner grid.

## Testcase

https://jsfiddle.net/gtc4o5zr/3/

## Version

2.9.0-beta.96

Loosely related: https://github.com/Semantic-Org/Semantic-UI/issues/5170

lubber-de:
## Description
This PR reverts and corrects #1739 for stackable nested grids.
The margins on mobile screens for stackable grid has now been adjusted so it only takes place when a stackable grid is not nested in another grid or a segment.

## Testcase
From #2263 
https://jsfiddle.net/lubber/bkrqfnhv/16/

From #1739 
https://jsfiddle.net/lubber/fv1hg0jm/23/

From #1907
https://jsfiddle.net/lubber/vqj0xdg1/22/

## Closes
#2263 
#1907


hugopeek:
Hi @lubber-de, thanks for your efforts. I'm not sure if the variable you introduced is a safe solution. Doesn't that apply to _all_ mobile grids now? Same goes for the !important classes. Although I'd love to see them go too, they might disrupt the balance in other scenarios.

I feel this is really an isolated bug inside _nested stackable mobile grids_. But I will test these changes out of course, to see what happens.

As a sidenote: the double gutter is not a design decision in SUI. See the comments ;)

https://github.com/fomantic/Fomantic-UI/blob/135cfd1f5f29030143ac36968894a52661013820/src/definitions/collections/grid.less#L1721-L1726

This is the original library with the same HTML:

https://jsfiddle.net/db7xuLgo/1/

lubber-de:
> I'm not sure if the variable you introduced is a safe solution. Doesn't that apply to _all_ mobile grids now? 

tbh, i wasn't really convinced myself, yesm this would affect all mobile grids, if changed (that's why i kept the default value to the original 0)

> Same goes for the !important classes. Although I'd love to see them go too, they might disrupt the balance in other scenarios.

From what i tested, those really were not necessary and removing them just fixed the the issue from #1739
I would appreciate if you could test more deeplyusing your projects 😉 

> I feel this is really an isolated bug inside _nested stackable mobile grids_.
Yes, the more i think about that, i also feel this way. The gutter should only be used if the stackable grid isnt nested. Tehre

> But I will test these changes out of course, to see what happens.
> 
> As a sidenote: the double gutter is not a design decision in SUI. See the comments ;)

Yes, saw them :) But they meant to not pad when nested. The unnested `.ui.stackable.grid` 0 gutter seems to be meant for stackable grids which are not nested and a direct child of body (or at least not inside another grid/container/segment) ,i believe.

> This is the original library with the same HTML:
> 
> [jsfiddle.net/db7xuLgo/1](https://jsfiddle.net/db7xuLgo/1/)

That seems to work exactly like my fix in this PR 🙂

PR #2273, opened 2022-03-22
build(scripts): reduce nightly publishing

hammy2899:
I've made changes to the nightly script to include the latest commit sha in the build metadata of the version so we can compare them. This allows us to only push the latest nightly to npm if there is now commits to push. I have also made changes to the workflow so it only continues and does the publish if the version was updated.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -22,8 +22,10 @@ jobs:
     - name: update nightly version
       run: node ./scripts/nightly-version.js
     - name: fomantic install & build
+      if: ${{ success() }}
       run: npx gulp install
     - name: publish to npm
+      if: ${{ success() }}
       run: |
         npm config set //registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN
         npm publish --tag nightly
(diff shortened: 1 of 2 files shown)


hammy2899:
@lubber-de Next publish we need to make sure the workflow cancels out correctly if there is no new commits just to confirm this line works as intended https://github.com/fomantic/Fomantic-UI/pull/2273/files#diff-3aaf0312cdfcec61163444cd114212c972fa7246adcf2382e07bb69957408333R50

hammy2899:
Ok I've made a change to use a more reliable method https://github.com/fomantic/Fomantic-UI/pull/2273/files#diff-3aaf0312cdfcec61163444cd114212c972fa7246adcf2382e07bb69957408333R51

I will still keep my eyes on it however.

lubber-de (approved):
Very good improvement!
LGTM

lubber-de:
@hammy2899 
Please check this again, because the nightly build seems to ignore the commit hash when publishing to npm
So next time the script cannot determine the last commit hash from npm and will still create a release every night i believe

image

image

lubber-de:
@hammy2899  I think i fixed it by #2279 Please review 😉

PR #2275, opened 2022-03-22
fix(calendar): keep whitespace between date and time

### LINKED ISSUE #2274: [Calendar] wrong datetime parsing
# Bug Report
Actual develop branch seems to wrongly parse datetime in calendar module.

Culprit seems to come from this commit: 8a331f0e62806964cfa00c271b09d54076f28df5.

In `src\definitions\modules\calendar.js` at line 1563.

```js
text = String(text).trim();
```
has been replaced by
```js
text = String(text).replace(/\s/g,'');
```

When this line is reverted, date parsing is correct.

## Steps to reproduce
1. Create a minimal calendar.
2. Set initial date by `data-date="2022-03-17 09:35:00"` or `data-value="2022-03-17 09:35:00"`.
3. Initialize the calendar module.

## Expected result
`March 17, 2022 9:35 AM`

## Actual result
`March 1, 2022 5:35 AM`

## Testcase
[JSFiddle](https://jsfiddle.net/pL8he1ta/)

## Screenshot (if possible)
image

## Version
2.9@nightly

lubber-de:
## Description
This PR corrects the whitespace removement bug from #2200, because the previous fix was not respecting a possible space character between date and time, which was still needed and not working anymore.

I was initially trying to fix it in a single regex but it got too complicated so i went for the easier, reliable and readable solution instead using 2 regexes.

### Broken
The provided date is wrong
https://jsfiddle.net/pL8he1ta/

### Fixed
Date is correct
https://jsfiddle.net/lubber/23y5mc0h/

## Screenshot
image

## Closes 
#2274 
#2200

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1560,7 +1560,7 @@ $.fn.calendar.settings = {
       if (!text) {
         return null;
       }
-      text = String(text).replace(/\s/g,'');
+      text = String(text).trim().replace(/([.:\/\-])\s+/g,'$1').replace(/\s+([.:\/-])/g,'$1').replace(/\s+/g,' ');
       if (text.length === 0) {
         return null;
       }


prudho (approved):
LGTM.

You just made ~100 users happy with this one 👍

PR #2279, opened 2022-03-24
fix(nightly): get commit hash from npm

lubber-de:
## Description

The nightly script is not able to provide the version name + commit hash as version number to npm, as npm always removes the hash from the version number. Thus the nightly is still always created because it can not compare the commit hashes.

This is now fixed, because the script takes the hash information out of the versions details from the npm registry.
npm still does not show the hash in the tags list, but at least the nightly can compare the commit hashes so it only creates a new version now when anything really has changed as @hammy2899 intended 

## Closes
[#2273 ](https://github.com/fomantic/Fomantic-UI/pull/2273#issuecomment-1076019628)

### DIFF
--- a/scripts/nightly-version.js
+++ b/scripts/nightly-version.js
@@ -19,7 +19,7 @@ const currentRev = childProcess // get the current rev from the repo
   .execSync('git rev-parse HEAD')
   .toString()
   .trim()
-  .substr(0, 7)
+  .slice(0, 7)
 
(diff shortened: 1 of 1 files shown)


hammy2899:
Looked into this and npm don't publish build info in the version which is why this was caused which is annoying however this does fix it. Thankfully they do include some extra info like git rev in the API.

hammy2899 (approved):
LGTM

PR #2284, opened 2022-03-28
fix(dropdown): floating shadow and radius on direct child menu only

grandeljay:
It's just a 1 character change so I'll keep it short:

There's an ugly and probably unwanted shadow and radius on the inner floating dropdown `.menu` which this fix removes.

https://jsfiddle.net/0cpwnfkz/

image


lubber-de (approved):
Nice catch!
LGTM

lubber-de:
@all-contributors please add @grandeljay for code

PR #2287, opened 2022-03-28
feat(chore): move from npm to yarn

hammy2899:
Moving the project to yarn instead of npm.

**Why?**
These days yarn is used more often as it is faster then npm at package resolution in turn reducing the time spent setting up the dev environment. It also resolves packages better and because of this it can reduce disk space used.

**To do:**
- [x] Remove package-lock.json and replace with yarn.lock
- [x] Change scripts to use yarn instead of npm in workflows
- [x] Test all is good before merging like build etc. still works.


lubber-de:
Why can't we support both? So leave package.json on addition to yarn.
By the removement of the auto install script yarn install is already working afaik

hammy2899:
This doesn't affect end users only us developers. NPM takes wayyy to long to install nowadays

lubber-de:
> This doesn't affect end users only us developers. NPM takes wayyy to long to install nowadays

But, when removing package.json, projects which have fomantic used as dependency probably dont work anymore when they use npm, isn't it?
So nom install fomantic-ui won#t work properly anymore (?)

And, without a package.json, npm publish does not work either, i believe?

lubber-de:
Forget my comment, i thought about `package.json`.... but you have removed `package-lock.json`

Sorry, my bad, i am sorry, all good, please continue :)

lubber-de (approved):

PR #2289, opened 2022-03-29
build(scripts): correct nightly workflow variable

y0hami:
Correct the variable used in the publish command for the version


lubber-de (approved):
LGTM :)

PR #2293, opened 2022-03-31
build(scripts): fix nightly script

hammy2899:
Fix the nightly script due to yarn publish not adding gitHead to the
registry. Also moved back to npm publish so we can keep functionality of
not publishing when there is no changes.


lubber-de (approved):
LGTM

PR #2297, opened 2022-04-02
build(actions): add condition for shouldPublish

hammy2899:
Add a conditional check to the shouldPublish if expressions


lubber-de (approved):
Unusual, but LGTM :D

PR #2305, opened 2022-04-05
Fix vertical pointing menu

grandeljay:
When you are using a pointing menu and it stacks, due to small screen size, the pointer should disappear. 

Unfortunately I have no experience with CSS preprocessors (yet), so I'll likely need some help for any changes.

image

https://jsfiddle.net/gLhb4um3/

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1793,11 +1793,6 @@ Floated Menu / Item
       border-top: @arrowBorder;
       border-right: @arrowBorder;
     }
-    @media only screen and (max-width: @largestMobileScreen) {
-      .ui.vertical.pointing.menu .item::after {
-        display: none;
-      }
-    }
   }
   .ui.pointing.menu .ui.dropdown .menu .item::after,
   .ui.vertical.pointing.menu .ui.dropdown .menu .item::after {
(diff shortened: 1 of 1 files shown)


grandeljay:
But isn't it always desired for a `fluid` pointing menu?

If it is fluid, there's nothing next to it. So IMO it's save to remove the pointer and your suggested code above would work great.

grandeljay:
And if there was something next to it (like another column) it would be underneath the menu anyway (due to the `.stackable` class)

lubber-de (changes requested):
Ok, convinced 🙂 

Please change it as follows:

- add the selector after the last selector inside the `  & when (@variationMenuVertical) {` scope
- make the selector dependant on enabled stackable grid and enabled fluid menu
- also do the same fix for `stackable pointing menu` see https://jsfiddle.net/lubber/rya73wz2/14/

So it becomes
```less
  & when (@variationMenuVertical) {
    .ui.vertical.pointing.menu .active.item:hover::after {
      background-color: @arrowVerticalHoverColor;
    }
    .ui.vertical.pointing.menu .active.item::after {
      background-color: @arrowVerticalActiveColor;
    }
    .ui.vertical.pointing.menu .menu .active.item::after {
      background-color: @arrowVerticalSubMenuColor;
    }
    // here your addition starts -->
    & when (@variationMenuFluid) and (@variationGridStackable) {
      @media only screen and (max-width: @largestMobileScreen) {
        .ui.stackable.grid .ui.fluid.vertical.pointing.menu .active.item::after {
          display: none;
        }
      }
    }
    // <-- here your addition ends
  }
```

Then at Line 1318
```less
& when (@variationMenuStackable) {
  /*--------------
      Stackable
  ---------------*/

  @media only screen and (max-width: @largestMobileScreen) {
    .ui.stackable.menu {
      flex-direction: column;
	  // here your addition starts -->
	  &.pointing .active.item::after when (@variationMenuPointing) {
			display: none;
	  }
	  // <-- here your addition ends
    }

```

lubber-de (approved):
LGTM

PR #2312, opened 2022-04-09
fix(button): loading buttons still shows labels

### LINKED ISSUE #2288: [dropdown] Item label/icon is still visible for loading dropdown
The dropdown item label/icon is still visible when the dropdown gets the `loading` class

Please see this fiddle for reference: https://jsfiddle.net/5pro4jvc/1/ (just select any item)

lubber-de:
## Decription
Whenever a `loading button` contains any kind of label it was still shown

## Testcase
- Select anything from the dropdown (it will get the loading state then)
- Remove CSS to see issue
https://jsfiddle.net/lubber/mrza0234/9/

## Broken
image

## Closes
#2288

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -169,19 +169,24 @@
 
     box-shadow: 0 0 0 1px transparent;
   }
-}
-& when (@variationButtonLabeledIcon){
-  .ui.labeled.icon.loading.button .icon {
+  .ui.ui.loading.button .label {
     background-color: transparent;
-    box-shadow: none;
+    border-color: transparent;
+    color: transparent;
   }
-}
-& when (@variationButtonBasic){
-  .ui.basic.loading.button:not(.inverted)::before {
-    border-color: @loaderFillColor;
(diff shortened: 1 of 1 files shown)


src/definitions/elements/button.less
```diff
@@ -169,19 +169,24 @@
 
     box-shadow: 0 0 0 1px transparent;
   }
-}
-& when (@variationButtonLabeledIcon){
-  .ui.labeled.icon.loading.button .icon {
+  .ui.ui.loading.button .label {
```
ivangomes, line 172: .ui.ui.loading.button

.ui.ui <-- is this a typo?

exoego: This is a CSS technique to increase specificity of the selector
https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

lubber-de: I even had to increase it again to make it work properly when compiled by 6d249925aa0695b344c5f2e92417f8b76ed12237 :wink:

exoego (approved):
LGTM

PR #2316, opened 2022-04-10
fix(popup): tooltips cannot contain headers

### LINKED ISSUE #2315: [popup] "Unexpected token Delim('.')" when packaging Fomantic-UI css distribution with Parcel.js
# Bug Report
Minifying/packaging Fomantic-UI using the latest version of [Parcel.js](https://parceljs.org/) and the `fomantic-ui-css` package from NPM fails with the following error:

```
@parcel/transformer-css: Unexpected token Delim('.')

  /home/michael/Documents/Freizeit/Programme/python/shc2/node_modules/fomantic-ui-css/semantic.css:61341:36
    61340 | 
  > 61341 | [data-tooltip][data-inverted]:after .header {
  >       |                                    ^
    61342 |   background: none;
    61343 |   color: #FFFFFF;
```

## Steps to reproduce
1. Install the latest Parcel.js and fomantic-ui-css packages from NPM:
   `package.json`:
   ```json
   {
     "dependencies": {
       "parcel": "^2.4.1",
       "fomantic-ui-css": "^2.8.8"
     }
   }
   ```
   ```bash
   npm install
   ```
2. Create a simple `main.js` file as an entry point for Parcel.js:
   ```js
   import 'fomantic-ui-css/semantic.css';
   import 'fomantic-ui-css/semantic.js';
   ```
3. Package project with Parcel.js:
   ```bash
   npx parcel build main.js
   ```

## Expected result
Packaging should complete successful.

## Actual result
Parcel.js fails with the error message shown above.

## Version
2.8.8

## Further Thoughts

The issue seems to be the following CSS/LESS code from [`popup.less`, line 246](https://github.com/fomantic/Fomantic-UI/blob/develop/src/definitions/modules/popup.less#L246):
```css
[data-tooltip][data-inverted]::after .header {
  background: @tooltipInvertedHeaderBackground;
  color: @tooltipInvertedHeaderColor;
}
```
I'm not sure, if this is actually invalid CSS. But it doesn't look sensible either: AFAIK, the `::after` pseudo-elements cannot contain nested elements, so I understand that Parcel.js complains about the `...:after .header` structure.

lubber-de:
## Description

A CSS only tooltip only contains text as it uses the `::after` pseudo element to display the text via the content attribute.
However content cannot contain HTML, thus an additional selector after a ::after pseudo element can never happen and is infact invalid.

This situation not only never happens but also makes some parsers like Parcel complaining about invalid CSS.

Thus, i simply removed the whole selector.

## Closes
#2315

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -243,10 +243,6 @@
       border: @tooltipInvertedBorder;
       box-shadow: @tooltipInvertedBoxShadow;
     }
-    [data-tooltip][data-inverted]::after .header {
-      background: @tooltipInvertedHeaderBackground;
-      color: @tooltipInvertedHeaderColor;
-    }
   }
 
   & when (@variationPopupPosition) {
--- a/src/themes/default/modules/popup.variables
+++ b/src/themes/default/modules/popup.variables
@@ -88,8 +88,6 @@
 @tooltipInvertedColor: @invertedColor;
 @tooltipInvertedBorder: @invertedBorder;
 @tooltipInvertedBoxShadow: @invertedBoxShadow;
-@tooltipInvertedHeaderBackground: @invertedHeaderBackground;
-@tooltipInvertedHeaderColor: @invertedHeaderColor;
 
 /* Arrow */
 @tooltipArrowVerticalOffset: -@2px;


exoego (approved):
LGTM

PR #2319, opened 2022-04-11
fix(toast): support context setting properly

lubber-de:
## Description
The `context` setting was not working properly. While all generated HTML was correctly positioned inside the DOM tree, the toast were still being displayed relative to the body, as they were always being displayed `fixed`.

When choosing a different context, such toasts need to be displayed `absolute` instead.

i also fixed a bug which was creating/searching the same container multiple times

## Testcase
Click on both toast buttons

### Broken
Both Toasts are shown to the top right of the whole screen
https://jsfiddle.net/lubber/c71s8oxj/2/

### Fixed
The teal toast is shown inside the context div
https://jsfiddle.net/lubber/c71s8oxj/4/

### Screenshots
|Broken|Fixed|
|-|-|
|image|image|


exoego (approved):
LGTM

PR #2326, opened 2022-04-16
fix(dropdown): properly support preventchangetrigger

lubber-de:
## Description
The `preventChangeTrigger` option for some behaviors was not optimal and not even working.
Using a behavior needed to also provide some only internally used parameters like `$selectedText` which is confusing also in the docs.
Such functions are now also supporting to be called as behaviors having the second argument used as the `precentChangeTrigger` argument (which usually is only the third or fourth option)

## Testcase
Click on the Button to add 2 values without selecting something from the dropdown

### Broken
The values are selected but the onChange trigger is still fired and you still need to call it with that null (to skip) parameter by
`.dropdown('set selected',['de','us'],null,true)`
https://jsfiddle.net/lubber/3w8jr560/9/

### Fixed
The values are selected and the onChange trigger is not fired and the behavior can be nicely called via
`.dropdown('set selected',['de','us'],true)`
https://jsfiddle.net/lubber/3w8jr560/10/

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1053,7 +1053,7 @@ $.fn.dropdown = function(parameters) {
             menuConfig[fields.values] = values;
             module.setup.menu(menuConfig);
             $.each(values, function(index, item) {
-              if(item.selected == true) {
+              if(item.selected === true) {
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

Issue #1989, opened 2021-06-17
[dropdown] cannot use mouse on iPad


jamessampford:
# Bug Report
Similar to #1988 a number of dropdowns will not work upon clicking with a mouse on iPad (namely a trackpad in my case). This again is using iPadOS 14.6. As per the slider, touch events work as expected

## Steps to reproduce
1. Click on each dropdown found on https://fomantic-ui.com/modules/dropdown.html most of the time there will be no response, ie. nothing happens on click

## Expected result
All dropdowns should open

## Actual result
A number of dropdowns do not open

## Version
2.8.7

lubber-de:
My same guess as for slider: The module checks for a touchdevice (which the ipad still is) and either binds touch or mouse events. As the mouse events are not bound, the mouse is not triggering the necessary events.

jakenuts:
I'm finding that the 'document.activeElement' on mobile lags behind the check for 'focusedOnSearch' so menu's don't appear in search selects on first click. On desktop, 'focusedOnSearch' returns true when you click in the dropdown search but false when you do it on a touch device (tracking focus shows that focus happens after, so too late)

mhthies:
Just as for #1988, the issue is caused by the distinction between touch and non-touch devices (see [dropdown.js, line 702](https://github.com/fomantic/Fomantic-UI/blob/03890cad95af0a3ea8e0ec29a19d6aeeac710736/src/definitions/modules/dropdown.js#L702)): On touch devices, only (!) touch events are handled, on non-touch devices (only) mouse events are handled.

Again, IMHO, the trivial solution would be removing this whole distinction and all the touch event binding code.

However, I learned from the Git history, that the distinction is required to mitigate some missing click events on iOS (#647, #1220), which seem to originate from strange behaviour in the mouse emulation of iOS Safari (see this StackOverflow thread: https://stackoverflow.com/questions/24077725/mobile-safari-sometimes-does-not-trigger-the-click-event). If the iOS issue in #647 could be fixed with one of the other workarounds, mentioned there, it would be possible to remove the special handling of touch events, as suggested above.

Otherwise, I would suggest always binding to both, mouse **and** touch events, and adding `event.preventDefault()` to the event handlers to avoid double-triggering through the browsers' mouse emulation.

PR #2327, opened 2022-04-17
fix(slider): support mouse & touch all the time

### LINKED ISSUE #1988: [slider] cannot move slider using mouse on iPad
# Bug Report
It appears that when using a mouse (in my case a trackpad) on an iPad, the slider cannot not move when trying to either drag the slider along not when clicking at a position along the slider. At least, this appears to be the case when using iPad OS 14.6. When touching the screen, the slider works as expected

## Steps to reproduce
1. Try adjusting any slider on https://fomantic-ui.com/modules/slider.html

## Expected result
Slider slides

## Actual result
Slider does not slide

## Version
2.8.7

mhthies:
## Description
As proposed in #1988, I did a rework of the touch event handling of the slider component. The new touch event code

 - does not prevent mouse input (fixes #1988)
 - tracks Touch identifier to avoid confusion by other fingers on the touch screen
 - handles touchcancel events in a sensible way
 - does no dynamic binding and unbinding to all touch events on the page. In combination with the second point, this enables multitouch sliding of multiple sliders on the same page (I know, nobody asked for this, but it works. ;) ) – Unfortunately, multitouch sliding of the first and second thumb of the same slider is not easy to implement due global state within each slider module.
 - only binds to touchstart events on the `.thumb` element to avoid accidential sliding while scrolling the page on a mobile device (obsoletes #1987)

The last point can be considered a disadvantage/regression compared to the old code, if you consider touch-sliding anywhere on the slider as a feature. However, in my experience, it's really annoying to change a slider position (which has immediate real-world effects in my home automation software) when scrolling the page and accidentially hitting a slider. Using the mouse, one can still slide the slider anywhere on its track.

## Testcase
https://jsfiddle.net/sfz45v0m/3/

## Screenshot (if possible)
--

## Closes
- #1988
- #1987

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -70,6 +70,7 @@ $.fn.slider = function(parameters) {
 
         $module         = $(this),
         $currThumb,
+        touchIdentifier,
         $thumb,
         $secondThumb,
         $track,
@@ -238,7 +239,7 @@ $.fn.slider = function(parameters) {
             $(document).on('keydown' + eventNamespace + documentEventID, module.event.activateFocus);
           },
           mouseEvents: function() {
-            module.verbose('Binding mouse events');
+            module.verbose('Binding mouse and touch events');
             $module.find('.track, .thumb, .inner').on('mousedown' + eventNamespace, function(event) {
               event.stopImmediatePropagation();
               event.preventDefault();
@@ -251,6 +252,14 @@ $.fn.slider = function(parameters) {
             $module.on('mouseleave' + eventNamespace, function(event) {
               isHover = false;
             });
+            // All touch events are invoked on the element where the touch *started*. Thus, we can bind them all
+            // on the thumb(s) and don't need to worry about interference with other components, i.e. no dynamic binding
+            // and unbinding required.
+            $module.find('.thumb')
+              .on('touchstart' + eventNamespace,  module.event.touchDown)
+              .on('touchmove' + eventNamespace, module.event.move)
+              .on('touchend' + eventNamespace, module.event.up)
+              .on('touchcancel' + eventNamespace, module.event.touchCancel);
           },
           slidingEvents: function() {
             // these don't need the identifier because we only ever want one of them to be registered with document
@@ -269,6 +278,11 @@ $.fn.slider = function(parameters) {
             $module.off('mousedown' + eventNamespace);
             $module.off('mouseenter' + eventNamespace);
             $module.off('mouseleave' + eventNamespace);
+            $module.find('.thumb')
+              .off('touchstart' + eventNamespace)
+              .off('touchmove' + eventNamespace)
+              .off('touchend' + eventNamespace)
+              .off('touchcancel' + eventNamespace);
             $module.off('keydown' + eventNamespace);
             $module.off('focusout' + eventNamespace);
             $(document).off('keydown' + eventNamespace + documentEventID, module.event.activateFocus);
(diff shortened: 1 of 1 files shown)


src/definitions/modules/slider.js
```diff
@@ -341,10 +320,30 @@ $.fn.slider = function(parameters) {
               module.bind.slidingEvents();
             }
           },
+          touchDown: function(event) {
+            event.preventDefault();  // disable mouse emulation and touch-scrolling
```
lubber-de, line 324: I wonder, if we probably also need
```
              event.stopImmediatePropagation();
```
because we had this in the previous code for module.bind.touchEvents and it is also used in the mouseevents "mousedown" event.
🤔

lubber-de (changes requested):
Very well done PR so far 👍🏼 
Thank you!

I only found a typo and have a question. Please see my comments.

A little additional request when done/answered:
Please provide two links for jsfiddle testing for this:
- The usual one (jsfiddle.net/abcdef../)
- a link we use for mobile testing using the following syntax
  - https://fomantic-ui.com/jsfiddle/#!/{jsfiddle-code}
  - For example https://fomantic-ui.com/jsfiddle/#!/sfz45v0m/3 (using your jsfiddle above)

Because I would like @jamessampford to test this on his ipad/mouse setup .
I believe you already tested it so far on your Android/bt mouse setup (?)

I have to rely/trust on your experience here, because i can only use Browserstack to emulate a wider range of mobile devices  but none of those have a real mouse attached to the devices, so the only situation i can personally test is if the changes basically still work on "usual" setups.

jamessampford:
@lubber-de This seems to work (both dragging and clicking) on my iPad (iPadOS 15.4.1, latest version) using:
- just my finger
- with Apple Pencil
- using the trackpad on my keyboard (not the Magic Keyboard, not Bluetooth connected)

Also works on my Android phone still but I can only try with finger


It may be a good idea to test in Windows with a touchscreen in all browsers, as well as hoping that someone else has a trackpad keyboard and/or Bluetooth mouse to verify in iOS/iPadOS

As this also relates to #1989 with the dropdown module, the same improvement will also need to be implemented there (and anywhere else similar)

src/definitions/modules/slider.js
mhthies: As far as I know, `stopImmediatePropagation()` prevents following event handlers on the same element and on parent elements from triggering for this event. I don't think skipping other event handlers on the same element is necessary, but it should not hurt either. Skipping possible event handlers on parent elements seems to be a good idea, when the slider is contained in other elements that are listening for touch events. I'll test it and add it if there are no unwanted side-effects.

mhthies:
> I believe you already tested it so far on your Android/bt mouse setup (?)

Yes, I tested with desktop Chrome on Linux (touch + mouse), mobile Chrome on Android (touch + mouse) and mobile Firefox on Android (touch + mouse).

> It may be a good idea to test in Windows with a touchscreen in all browsers, as well as hoping that someone else has a trackpad keyboard and/or Bluetooth mouse to verify in iOS/iPadOS

This is a good idea, although I'm hoping that all modern desktop browsers stick to the Touch API as it is described on MDN. Unfortunately, I don't have Windows touch device at hand.

> As this also relates to https://github.com/fomantic/Fomantic-UI/issues/1989 with the dropdown module, the same improvement will also need to be implemented there (and anywhere else similar)

Yes, I'm already working on this one. Another PR will follow for this, as soon as I can fix some strange behaviour in touch interaction of nested dropdown menus (which is already present in the current code).

mhthies:
Here is the updated JSFiddle:

- https://jsfiddle.net/jyg4wtcq/
- https://fomantic-ui.com/jsfiddle/#!/jyg4wtcq

jamessampford:
@lubber-de Seems to still work the same with the same 3 different ways of testing it as before (with dragging and clicking/tapping) with both the sliders

lubber-de (approved):
LGTM! 👍🏼

lubber-de:
@all-contributors please add @mhthies for code

PR #2330, opened 2022-04-19
feat(slider): customizable event selectors

lubber-de:
## Description
This PR makes the event selector target customizable via settings, so the initial idea of #1987 would be now easier doable based on the new codebase from #2327 

@mhthies Please review, i am interested in your opinion

## Testcase
https://jsfiddle.net/lubber/k3efcLuj/1/
https://fomantic-ui.com/jsfiddle/#!lubber/k3efcLuj/1/

## Replaces
#1987

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -60,6 +60,7 @@ $.fn.slider = function(parameters) {
         className       = settings.className,
         metadata        = settings.metadata,
         namespace       = settings.namespace,
+        selector        = settings.selector,
         error           = settings.error,
         keys            = settings.keys,
         interpretLabel  = settings.interpretLabel,
@@ -240,7 +241,7 @@ $.fn.slider = function(parameters) {
           },
           mouseEvents: function() {
             module.verbose('Binding mouse and touch events');
-            $module.find('.track, .thumb, .inner').on('mousedown' + eventNamespace, function(event) {
+            $module.find(selector.mouseTarget).on('mousedown' + eventNamespace, function(event) {
               event.stopImmediatePropagation();
               event.preventDefault();
               module.event.down(event);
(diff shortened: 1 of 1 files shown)


mhthies:
Unfortunately, this doesn't work as expected. Here is an extended JSFiddle with the code from this PR showing the different options:

- https://jsfiddle.net/bcd23hgs/5/
- https://fomantic-ui.com/jsfiddle/#!/bcd23hgs/5/

The mouse setting does not have an effect due to the additional `mousedown` binding in [line 249](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L249). I think, we can drop one of the two event bindings (but didn't test it yet).

Setting `touchTarget` to anything other than `.thumb` makes (very strange) problems due to my code, which assumes that the event target is always a thumb (see [line 332](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L332). This section would need to be extended to use `module.determine.closestThumb()` just [like the mouse `down` event handler does](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L311).

lubber-de:
Thanks for taking a look!
> The mouse setting does not have an effect due to the additional mousedown binding in [line 249](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L249). I think, we can drop one of the two event bindings (but didn't test it yet).

oh, yes, the second one seems to override the first (which is taking care of event bubbling...)
So i _guess_ the second one can be removed ? Yes, have to test this

> Setting touchTarget to anything other than .thumb makes (very strange) problems due to my code, which assumes that the event target is always a thumb (see [line 332](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L332)). This section would need to be extended to use module.determine.closestThumb() just [like the mouse down event handler does](https://github.com/fomantic/Fomantic-UI/blob/168c1aa8a53b403d72eea2242fe54079465e8772/src/definitions/modules/slider.js#L311).

Makes sense 👍🏼 I'll test this. Thanks for the hint!

exoego (approved):

PR #2331, opened 2022-04-19
fix(dropdown): rework touch input handling

### LINKED ISSUE #1989: [dropdown] cannot use mouse on iPad

mhthies:
## Description
This PR fixes #1989 in a similar way as my previous PR for the slider (#2327): First, I removed all special event handling for touch events and the whole distiction between touch- and non-touch devices, so mouse events will be handled on touch devices as well. Then, I added only a few touch event bindings for the things that don't work via mouse event emulation of the mobile web browsers.

In addition, I fixed two minor bugs with nested submenus being not correctly hidden in some cases (see commit messages for more details).

I tested everything on
- Chrome mobile on Android (with touch and Bluetooth mouse)
- Firefox mobile on Android (with touch and Bluetooth mouse)
- Chromium on Linux/X11 (with multitouch screen and mouse)
- Firefox on Linux/X11 (with multitouch screen and mouse)

## Testcase
- https://jsfiddle.net/17ouw0gd/18/
- (result only:) https://fomantic-ui.com/jsfiddle/#!/17ouw0gd/18/

## Screenshot (if possible)
(looks the same as before)

## Closes
- #1989


lubber-de (approved):
LGTM, very good work. Thanks! and i learned a lot 😄

PR #2340, opened 2022-04-25
feat(build): require node 12 as minimum

lubber-de:
## Description

Upgrade node dependency to 12, which is the least supported version by now (yes, runs out of service by end of April, but i believe v12 is still widely spread and other dependencies are ugrading as well


exoego (approved):
LGTM

PR #2357, opened 2022-04-29
fix(emoji): emojies/flags were not properly themeable

lubber-de:
## Description

Emojis+new Flags were not properly themeable. Means, if one would have changed the theme for them inside theme.config, Those were still (also) included form the default theme, because their whole definition was done inside the overrides files.

I now moved the pure emoji list mixin into the .variables file instead of overrides and moved the generator code inside the less file. This way any modified emoji or flag list can be changed/reduced/modified in a separate theme by redeclarin the mixins as expected.

I tag this as breaking change because it could happen that somebody needs to adjust their own themes variables/overrides file for emojies.

> BTW: Similar issue (still) for icons. Separate PR follows :)

## Testcase
- Change the theme for emoji to `joypixels` in theme.config
- Change the theme for flags to `famfamfam` in theme.config
- Build FUI

### Broken
Both themes are included in semantic.css (tweemoji default and joypixels for emoji, tweemoji and famfamfam png sprite for flag)

### Fixed
Only the selected theme of each element is included


exoego (approved):

src/definitions/elements/flag.less
```diff
@@ -39,4 +39,49 @@ i.flag:not(.icon)::before {
...
+each(@size-map, {
+  i.flag.@{key} {
+    font-size: 1.5em * @value;
+    vertical-align: middle;
+  }
+});
+
+/*rtl:begin:ignore*/
+
+each(@flags, {
```
mvorisek, line 51: Somethings generates 2 same selectors:

```
i.flag.rainbow::before,
i.flag.rainbow::before,
i.flag.gay.pride::before {
```
```
i.flag.transgender::before

i.flag.rainbow::before,
i.flag.rainbow::before,
i.flag.gay.pride::before

i.flag.pirate::before
```

selectors should be not generated at all, they are not country flags

lubber-de: i'll fix that. However they are designed as flags, even if not a country

mvorisek: FYI I added `i.flag.crosseds`

lubber-de: > FYI I added `i.flag.crosseds`

I see. Will remove crosseds, this is indeed not even a "flag" as meant to be used as a fui flag

lubber-de: Fixed by #2466

Issue #2355, opened 2022-04-29
Use native font stack, remove Lato font download


mvorisek:
# Feature Request

![image](https://user-images.githubusercontent.com/2228672/84135218-023bd980-aa4a-11ea-9cee-686b5e1fc241.png)

`non-nglish` characters like `ř` are displayed very differently than the others without diacritics

word like `Řeřicha` should be exactly/pixelly the same as `Rericha` (ignoring the accents)

quite important for European languages where accents are very common

it seems like a Lato does not have a support for these characters and also the Lato v1 is still used (Lato v2 is not however available thru Google Fonts)

see https://getbootstrap.com/docs/5.1/content/reboot/#native-font-stack

I propose to use the native font stack also for Fomantic UI, it will fix issues with non-English characters and also remove HTTP requests to Google Fonts

mvorisek:
related to https://github.com/fomantic/Fomantic-UI/issues/367, same solution (native font stack) proposed

mvorisek:
this is what Bootstrap use: https://github.com/twbs/bootstrap/blob/v5.1.3/scss/_variables.scss#L577

lubber-de:
Fixed and implemented by #2359 
- Google Import removed (they used Lato 1.0 anyway)
- Bundled Lato 2.0 for offline usage which also covers all missing diacritic letters

PR #2359, opened 2022-04-30
feat(site): offline bundled lato 2.0 and systemfonts, no googleimport by default

### LINKED ISSUE #367: Offline fonts
# Feature Request
I absolutely love Semantic-UI/Fomantic-UI. However, one of my biggest gripes with Semantic-UI was it could only be used online. It would be wonderful if there was a way to include the fonts with the repository, npm or build process so they didn't have to be pulled from the internet. 

## Description
I work for a financial organization and we use Semantic-UI/Fomantic-UI in a few or our internal applications. On the rare occasion that we have internet issues, the fonts can't be loaded, and it essentially makes our applications unusable.

Also, we use another application called PrinceXML to create PDFs from these applications. Because we block all internet traffic on servers, PrinceXML can't pull the fonts/icons when it generates the PDFs.

## Example
We currently use the semantic-ui-offline project to get this to work properly: https://github.com/arthurlacoste/semantic-ui-offline

This isn't a deal breaker for us, but would just be nice if the fonts could be saved and served locally without relying on Google for them. Thanks!

lubber-de:
## Description
This PR bundles the recent Lato 2.0 Font. It uses the same styles as the previous googleFonts Import does.

This fixes possible GDPR issues as the default theme now disables googlefont import.

I implemented a new variable `@fonts` which allows for a totally free description of font faces to be imported if needed. Each declared property will be used, so one can customize them as needed by that single variable.

Disabling the whole import can be done via  `@importFont: false` or setting `@font: {}` or setting `@fontName: ''` just as one likes.
The new variables for unicode or font properties are done to simplify the default import and of course if one just wants to adjust them without redeclaring the whole `@font` variable mixin.

I also added `@supportIE` which allows to disable the woff-font import, as all other supported browsers support woff2 already.
> This new variable and also the icon font-face will be adjusted accordingly in a separate PR... in case you might ask/wonder 😉 

The default theme uses Lato-Latin 2.0  now which is smaller in size but includes the most important glyphs for European Latin-based languages which were missing in Lato v1 (via googlefonts) like in "Řeřicha"
It also uses Lato with Greek+Cyrillic+Extended 2.0 via `unicode-range` property which makes sure the font is only loaded when related characters are used inside the page.
If no character matches at all, the system fonts will be used as fallback.

I created the unicode ranges by analyzing the fonts character ranges using https://github.com/fontforge/fontforge 

I also added an optional theme to adjust the font via a change in theme.config:

- systemfont (which completely disables any font import and uses the platform system fonts only instead AND adjusts the bold setting so the overall visual appearance is very similar) See comments below for more info.

The used Lato fonts are available free of charge under the SIL Open Font License from http://www.latofonts.com/

## Testcase
Build FUI :)

### To test System Fonts
- Adjust `theme.config`:
```less
@site       : 'systemfont';
```
- Build FUI
- Use some latin as well as cyrillic and greek characters on a webpage
- inspect the character sections to see which font is used

## Screenshots
lato2

|Lato 2.0|Lato 1.0 (until FUI 2.8.8)| systemfont|
|-|-|-|
image|image|image|

## Closes
#367 
#2355 
https://github.com/Semantic-Org/Semantic-UI/issues/7075


exoego (dismissed):
LGTM

src/themes/default/globals/site.variables
```diff
@@ -9,12 +9,40 @@
 --------------------*/
 
 @fontName          : 'Lato';
+@fontFileName      : 'LatoLatin';
```
mvorisek, line 12: "Lato v1" has very bad font for Cyrillic. With "LatoLatin", this issue remains.

"System native font" (as defined in Bootstrap v4+) however renders Cyrillic glyphs correctly, can it be defined as a fallback font?

I understand the "LatoLatin" font is smaller in size, but a lot of websites use the compiled dist semantic.css style. Is there any easy way how to use full "Lato" font /wo recompile?

lubber-de: > "System native font" (as defined in Bootstrap v4+) however renders Cyrillic glyphs correctly, can it be defined as a fallback font?

Not without recompile by using the systemfont theme for site or changing the `@pageFont` accordingly.

However, you may do a manual overwrite after importing the prebuild semantic.css like this :
```css
*:not(i) {
	font-family: system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", "Liberation Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !important;
}
```
Providing the System fonts as fallback src setting for the font family does not make sense i IMHO, because the Lato font will still be applied/loaded.
I decided against making Systemfonts the default as this will definately break many projects and the main concern was to disable the googlefonts connection.

> Is there any easy way how to use full "Lato" font /wo recompile?

Not without recompile. My intention was to have the same font as before (LatoLatin 2.0 in comparision to 1.0 is at least fixing the missing european diacritic letters which issued by #2355)

But we can of course discuss this 🙂 

@fomantic/maintainers @fomantic/admins 

Should i
- remove the extra Lato Full theme and use the use that as default instead? (Only downside: The font files are 5 times as big as LatoLatin is
- revert the logic: Use Lato Full as default and LatoLatin as optional theme?
- leave the PR as it is

?

mvorisek: This is how Google define fonts:

https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400%3B0,700%3B1,400%3B1,700&subset=latin&display=swap

The individual font files are downloaded on the first use, test:

```
<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400%3B0,700%3B1,400%3B1,700&subset=latin&display=swap">

<style>
* { font-family: 'Lato'; font-weight: 400; }
</style>

<p>some text</p>
```

given this behaviour, we can define both LatoLatin and Lato fonts together with `unicode-range` CSS property. This way, full Lato will be supported by default, but downloaded only if needed by the webpage. `unicode-range` property can be also used to define fallback to native system font stack for the glyphs not present even in the full Lato font.

lubber-de: I'll check and test that and will adjust the PR accordingly. Thanks for the hint 👍

prudho (approved):
I think it looks good as it 👍

src/themes/default/globals/site.variables
mvorisek: I compared the Lato font with native font stack visually. The native font stack (as defined by Bootstrap) is almost the same as Lato font, only bold font (/w style `font-weight: bold`) is much bolder. `bold` `font-weight` equals to 700 weight. When I changed it to 600, the visual difference is very low. I also checked other popular websites like Github :) and it uses `font-weight` set to 600. So I propose to change `font-weight: bold` in Fomantic-UI to `font-weight: 600` and drop any bundling of Lato font completely. Using native font stack is definitely a trend, the biggest styling framework Bootstrap has choosen it for a reasons. For the user it renders instantly (no download needed) and it covers also the biggest Unicode range consistently (like the `Řeřicha` issue).

lubber-de: 🤔 Well, at least for windows, i can see a very clear difference using systemfonts vs Lato

|Lato|System|
|-|-|
|image|image|

### Lato
image
### System
image

Here are two jsfiddles to compare almost all components
Lato https://jsfiddle.net/hpetaxy0/show
System fonts https://jsfiddle.net/lubber/9c3hf7eq/1/show

The minimum immediate different is the bold weight. Yes, if this is changed from `bold` to `600` the thickness is nearly the same as when using Lato. 
However, the font-weight variable `@bold` is unfortunately not used everywhere. font-weight is changed by  individial variables at several places (calendar, button, table...) where this would have to be changed as well everywhere, which, in turn, _could_ have impact on existing projects, even while at the first look, it looks similar/identical. 
Also the line height changes for the systems fonts which in turn also changes the padding which _could_ have undesired impact on existing projects.

So, we will bundle Lato 2.0 with 2.9.0 (using unicode-range as suggested by you) for most possible backward compatibility to existing projects. Other people might have a different opinion about using only system fonts.

Don't get me wrong, i fully understand and mostly agree with your arguments 🙂 

I am currently rewriting the PR to have the full freedom to import as many custom fontfaces with custom font-face properties as desired by a single variable (or none and only stay with native fontstack, which will become the fallback default as suggested by you (instead of an optional theme)).

We can discuss using pure Native fontstack as default for 3.0, which will be as simple as setting `@font: ''` and optional `@importFonts: false` or `@fonts: {}` and the LESS script takes care of that.

lubber-de: 🤔 Seems i am wrong with my statement about the `@bold` variable, which indeed seems to be used consistenly everywhere. So i will consider changing this from `bold` to 600 in this PR as you suggested (bold/600 look identical using lato and 600 looks nearly equal using systemfont as you explained.

lubber-de: 🤔 There is one issue when using the systemfonts and bold: The default tag `<b></b>` defaults the font-weight to `bold` which would still show them noticable different when not also set to `600` manually and i believe this is still often used technique to show some inline text as bold.
Yes, we could override this for `strong,b` to 600 in general via `reset.overrides`, but i fear side effects for existing projects as this will also override something outside the `ui` scope.

mvorisek: I would not object decision in any direction for now. On one side, we override some global tags like `h1` and set font style globally (on `body`), so overriding `font-weight` of tag `<b>` to `600` outside the `ui` scope is not that problematic.

Nevertheless I slightly prefer to keep `<b>` `font-weight` set to default (`bold`/`700`) as it should be consistent with setting CSS font weight to `bold`/`700`, eg. `<b>test</b>` and `<span style="font-weight: bold;">test</span>` should be strictly equivalent. Also I find the visually thicker bold much more readable in places where "desired for semantic purposes".

Currently, `b` (and `strong`) is actually overridden/set globally to `font-weight: bolder` which does imply font-weight `700` when used with `normal`/`400` font [1]. So I think we do not need any change for `b`/`strong`.

[1] https://css-tricks.com/firefoxs-bolder-default-is-a-problem-for-variable-fonts/ (inconsistent across Chrome/Firefox, but consistent when the base font has `normal`/`400` font weight 😄)

lubber-de: I updated the PR and description accordingly. :)

lubber-de:
@fomantic/maintainers  I updated the PR, made it more flexible and used the advises by @mvorisek (thanks again)
Please review again

src/themes/default/globals/site.variables
```diff
@@ -3,18 +3,112 @@
...
        Fonts
 --------------------*/
 
 @fontName          : 'Lato';
-
-@headerFont        : @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
-@pageFont          : @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
+@fontFileNameLatin : 'LatoLatin';
+@fontFileNameSupplement: 'Lato';
+
```
mvorisek, line 15:
```suggestion

// based on https://github.com/twbs/bootstrap/blob/v5.1.3/scss/_variables.scss#L577 list
```

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/383

PR #2362, opened 2022-04-30
fix(chore): eol of text files CRLF to LF

mvorisek:
pure CS fix


FAQ.md
```diff
@@ -1,38 +1,38 @@
-# FAQ
```
lubber-de, line 1: Why is that file changed? I cannot spot the change (and the PR was about EOL fixes)

mvorisek: I belive CRLF (`\r\n`) or CR (`\r`) EOL was used instead of pure NL (`\n`).

lubber-de (changes requested):
Please dont change the files inside the dist/folder. 
That folder ie being generated upon release.
Adjust the relevant less/overrides/variables file instead

mvorisek:
> Please dont change the files inside the dist-folder. That folder is being generated upon release. Adjust the relevant less/overrides/variables file instead

Fixed.

`git add . -N && git diff --exit-code` should be added to CI after compilation so commited files are verified if they are in-sync.

tasks/collections/README.md
```diff
@@ -13,4 +13,4 @@ gulp = install(gulp);
 
 // tasks are now injected and ready to be used
 gulp.start('install');
-```
\ No newline at end of file
+```
```
lubber-de, line 16: There is no change at all it seems?

lubber-de (changes requested):
Please remove the files which purely changed the whole Line Separator. (See comments)
I am fine with all the other changes which only add a new line at the end of each file.

FAQ.md
mvorisek: Are you worried about possible merge conflicts or blame results?

These files should be normalized. A lot of people, including me, have environment setup to normalize text files on git stage/commit and this is causing a real problem to me. If this PR is complicated to review, please compare the files using case insensitive diff tool, there should be no changes.

tasks/collections/README.md
mvorisek: image

there is missing EOL too, for some reasons, GH does not render 
image
symbol on the left side of the diff

FAQ.md
lubber-de: No, i am not worried about merge conflicts :)

I am wondering why only a few files are "normalized" if their content does not even change. It's just inconsistent, that's why 😉 
Some Files are changed in your PR for LineFeed settings, but some/many are not, although they have the same Line Ending settings.

I just don't understand the reason for the change. Just because _your_ environment setup automagically does/needs such changes, others might have a different setup and in the worst case other contributors probably change it it again in another PR just because of their setup.

If your setup really needs LF instead of CRLF, why only for a few files and leave all other files using CRLF as they are?
That's confusing to me, so i dont get the technical reason for that. I know that Linux/Mac/Win have different default settings for LF Endings, but this does not conflict with building FUI and _IF_ we would change it, it should be done for _all_ files to be consistent, isn't it?

However, i dont feel, IDEs should auto change the Line endings at all.

Don't get me wrong, i appreciate your effort to support FUI, but i feel this is somewhat just a cosmetic or personal preference change, which others might not share.

@fomantic/maintainers  Opinion?

mvorisek: > and leave all other files using CRLF as they are?

This is probably the misunderstanding. Only 2 files were using CRLF and I changed them to LF. All other files are thankfully with LF EOLs already.

lubber-de: Ok, found it: git checkout in windows is converting from LF to CRLF for editing, that's why i did not see that the other less files were still LF inside git origin.... Sorry, my fault using Windows 😆 

So, i will merge your PR, all fine, sorry for any inconvenience

lubber-de (approved):
LGTM after all :) Thanks for your patience

PR #2377, opened 2022-05-29
fix(table): colored marked in colored cells

### LINKED ISSUE #2168: [table] Marked cells cannot be colored
# Bug Report
A table cell cannot be marked and colored at the same time.
This happens because the selector `.right.marked.blue` (for example) takes precedence over the simpler `.orange` (for example) selector. Maybe switching to something like `orange colored` as the classes required to get a colored cell and styling using selectors like `[class *= "orang colored"]` and `[class *= "marked blue"]` could solve the issue.

## Steps to reproduce
Set the classes `yellow right marked red` for a `<td>`

## Expected result
The cell gains a yellow background and text and a red marker on the right

## Actual result
Only the red marker is visible, background and text coloring are discarded

## Testcase
[https://jsfiddle.net/e24q89c7/](https://jsfiddle.net/e24q89c7/)

## Version
2.8.8

lubber-de:
## Decription
Colored cells could not also get marked.
This is now fixed the way that you have to explicitely declare marked cell color classnames in an ordered way like `right blue marked` (first color, then `marked`), in combination with `red colored` if the cell should have a red background color as well.

Usual colored cells (without the suffix `colored`) are still supported as i believe this is a more used feature than `marked`.

So the following class orders are now supported

|Classname order|Appearance|
|-|-|
|red|will display a red backgrounded cell|
|red colored|will display a red backgrounded cell|
|right orange marked|will display a transparent cell having a right orange marker|
|red colored right orange marked|will display a red backgrounded cell having a right orange marker|

## Testcase
https://jsfiddle.net/lubber/z8yx3Lbe/5/

## Screenshot
image

## Closes
#2168

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -243,7 +243,7 @@
       @color: @value;
       @c: @colors[@@color][color];
       @l: @colors[@@color][light];
-      .ui.ui.ui.ui.table:not(.unstackable) tr.marked.@{color} {
+      .ui.ui.ui.ui.table:not(.unstackable) tr[class*="@{color} marked"] {
         &.left {
           box-shadow: @responsiveRowBoxShadow, @coloredBorderSize 0 0 0 @c inset;
         }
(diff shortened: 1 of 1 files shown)


prudho (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/352

PR #2378, opened 2022-05-30
feat(chore): optional compile for ie specific code

lubber-de:
## Description
This PR makes all IE specific LESS code optional. By default the new variable `@supportIE` (introduced by #2359 ) is true, so the code compiles as before

Early Edge Legacy versions still use the IE11 engine / -ms prefix syntax, so setting `supportIE: false;` will also disable support for those browsers.


exoego (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/383

PR #2409, opened 2022-07-21
feat(dropdown): improve accessibility in relation to search dropdown

Inselhopper:
## Description
in case of using the &lt;label&gt; element before a search dropdown to improove accessibility
- id of module is used to add an unique id to this very label 
- a reference to this label is made via the attribut "labelledby"
- see https://www.w3.org/WAI/tutorials/forms/labels/

## Testcase
https://jsfiddle.net/inselhopper/ot5zqvb2/38/

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -110,8 +110,8 @@ $.fn.dropdown = function(parameters) {
               settings.ignoreDiacritics = false;
               module.error(error.noNormalize, element);
             }
-            module.create.id();
-            module.setup.layout(id);
+
+            module.setup.layout();
 
             if(settings.values) {
               module.set.initialLoad();
@@ -124,6 +124,7 @@ $.fn.dropdown = function(parameters) {
             module.save.defaults();
             module.restore.selected();
 
+            module.create.id();
             module.bind.events();
 
             module.observeChanges();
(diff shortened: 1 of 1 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -366,11 +365,22 @@ $.fn.dropdown = function(parameters) {
...
-              $search = $('<input />')
+              if($module.prev('label').length) {
+                $search = $('<input />')
                 .addClass(className.search)
                 .prop('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
+                .attr('aria-labelledby',id+'_formLabel')
                 .insertBefore($text)
-              ;
+                ;
+                $module.prev('label').attr('id', id + '_formLabel');
```
hdogan, line 375: What if I already set `id` attribute to that label?

src/definitions/modules/dropdown.js
```diff
@@ -366,11 +365,22 @@ $.fn.dropdown = function(parameters) {
...
+              }
+              else {
```
lubber-de, line 377: Let's not duplicate the code here.
I suggest to change your PR into something like this (which will also support an already existing id as [questioned by @hdogan](https://github.com/fomantic/Fomantic-UI/pull/2409/files#r928965797) )
```javascript
var labelNode = $module.prev('label');
$search = $('<input />')
  .addClass(className.search)
  .prop('autocomplete', module.is.chrome() ? 'fomantic-search' : 'off')
;
if (labelNode.length) {
  if(!labelNode.attr('id')) {
    labelNode.attr('id', module.get.id() + '_formLabel');
  }
  $search.attr('aria-labelledby',labelNode.attr('id'));
}
$search.insertBefore($text);
```

lubber-de (changes requested):
Thanks for the PR. Please see my comment for a possible change.

src/definitions/modules/dropdown.js
```diff
@@ -110,8 +110,8 @@ $.fn.dropdown = function(parameters) {
               settings.ignoreDiacritics = false;
               module.error(error.noNormalize, element);
             }
-
-            module.setup.layout();
+            module.create.id();
+            module.setup.layout(id);
```
lubber-de, line 114: The (id) parameter is not consumed inside the layout function.
As i suggested in https://github.com/fomantic/Fomantic-UI/pull/2409/files#r932673809 the "module.get.id()" function should be used instead, so please remove the provided id as a function call parameter.

src/definitions/modules/dropdown.js
Inselhopper: thanks a lot for your objection, it is fixed

src/definitions/modules/dropdown.js
Inselhopper: thank you so much, it is a lot better now

lubber-de (approved):
LGTM

lubber-de:
@all-contributors please add @Inselhopper for code and a11y

PR #2416, opened 2022-08-04
feat(dropdown): optional stuck addition item

xamey:
## Description

When using a search dropdown, having `allowAdditions` activated and `hideAdditions` disabled, every time the user types something in the input, it is prepended in the menu.
Perhaps, in some cases, the user may not be aware that he has an addition option in the dropdown to click on : 
- there is no value selected yet in the dropdown. The user clicks on it, scrolls down, begins to type something, but there is so much results shown in the menu that the first line with the addition message displayed isn't shown
- there is a value selected in the dropdown. The user clicks on it, and the value is displayed within the menu, but it is like the 50th element on 100. The scrollbar is half way. The user begins to type something else, and it doesn't show the first line with the addition message

It is even more remarkable if `fullTextSearch` equals `true`, a fuzzy search could display a lof of results while typing an addition.
If the user isn't used to that option, he may miss it or don't fully understand this option.

An example, where a value is already selected, which is far from the 1rst entry, and I begin to type a custom entry : 
Aug-04-2022 18-26-35

That's why I wanted to introduce a new option : `scrollTopWhenTypingAdditions`.
If `allowAdditions` equals `true`, `hideAdditions` equals `false` and `scrollTopWhenTypingAdditions` equals `true`, every time the user types something in the search input, the menu scrolls to the top, so the custom addition entry is always displayed.

## Screenshot

With `scrollTopWhenTypingAdditions` as `false`
Aug-04-2022 18-26-35

With `allowAdditions` as `true`, `hideAdditions` as `false` and `scrollTopWhenTypingAdditions` as `true`
Aug-04-2022 18-46-10


xamey:
> Could you please provide a working jsfiddle where your addition gets reproduced? (like the example you took your screenshots from) From a quick setup, without your addition, the "add:" Line is already always displayed at the top of the dropdown. For example at [jsfiddle.net/lubber/948ts7oa/25](https://jsfiddle.net/lubber/948ts7oa/25/) This would also deserve a proper example in the docs

It can be reproduced through the jsfiddle you made : 
- step1; scroll down to Javascript 
image
- step2; type the letter "a", the "add: " line isn't displayed
image
- step3; you have to scroll up to find it  
image

My addition aims to prevent the user from scrolling up to find this line, i.e. removing the step 3.

I'll add an example in the docs PR I made as soon as possible :)

lubber-de (commented):
@xamey What do you think about solving it via pure CSS and make the additon item sticky instead?
```css
.ui.selection.dropdown > .menu > .addition.item:first-child {
  position: sticky;
  top:0;
  box-shadow: 0 2px 3px 0 rgb(34 36 38 / 15%);
  background: #f8f8f8;
  z-index:1;
}
```
See https://jsfiddle.net/dy52z9ek/2/
stickyaddition

xamey:
> @xamey What do you think about solving it via pure CSS and make the additon item sticky instead?
> 
> ```css
> .ui.dropdown > .menu > .addition.item:first-child {
>   position: sticky;
>   top:0;
>   box-shadow: 0 2px 3px 0 rgb(34 36 38 / 15%);
>   background: #f8f8f8;
>   z-index:1;
> }
> ```
> 
> See [jsfiddle.net/dy52z9ek/2](https://jsfiddle.net/dy52z9ek/2/) stickyaddition [ stickyaddition ](https://user-images.githubusercontent.com/18379884/185113601-ebf1a4b1-68c7-4d0a-93cf-9616f784212a.gif) [ ](https://user-images.githubusercontent.com/18379884/185113601-ebf1a4b1-68c7-4d0a-93cf-9616f784212a.gif)

Excellent! That's indeed way better. Thanks for the suggestion, I'll modify the PR with that.

xamey:
> @xamey What do you think about solving it via pure CSS and make the additon item sticky instead?
> 
> ```css
> .ui.selection.dropdown > .menu > .addition.item:first-child {
>   position: sticky;
>   top:0;
>   box-shadow: 0 2px 3px 0 rgb(34 36 38 / 15%);
>   background: #f8f8f8;
>   z-index:1;
> }
> ```
> 
> See [jsfiddle.net/dy52z9ek/2](https://jsfiddle.net/dy52z9ek/2/) stickyaddition [ stickyaddition ](https://user-images.githubusercontent.com/18379884/185113601-ebf1a4b1-68c7-4d0a-93cf-9616f784212a.gif) [ ](https://user-images.githubusercontent.com/18379884/185113601-ebf1a4b1-68c7-4d0a-93cf-9616f784212a.gif)

I would just add a `z-index: 0;` to other items, otherwise, when hovering an item, it induces that display :
image

```

.ui.selection.dropdown .menu > .item {
  border-top: 1px solid #FAFAFA;
  padding: 0.78571429rem 1.14285714rem !important;
  white-space: normal;
  word-wrap: normal;
  z-index: 0;
}

.ui.dropdown > .menu > .addition.item:first-child {
  position: sticky;
  top: 0;
  box-shadow: 0 2px 3px 0 rgba(34, 36, 38, 0.15);
  background: #FFFFFF;
  z-index: 1;
}
````

See the result in [this jsfiddle](https://jsfiddle.net/ow2v7Lts/7/).
image

File: src/definitions/modules/dropdown.less [hunk 1]
```diff
@@ -565,13 +565,23 @@ select.ui.dropdown {
...
     word-wrap: normal;
+    z-index: 0;
   }
 
   /* User Item */
   .ui.selection.dropdown .menu > .hidden.addition.item {
     display: none;
   }
 
+  /* User addition item */
```
lubber-de, line 576: To "support" IE11/old Edge, we should wrap this by `@supports(position: sticky)`

src/definitions/modules/dropdown.less
```diff
@@ -565,13 +565,23 @@ select.ui.dropdown {
     padding: @selectionItemPadding;
     white-space: normal;
     word-wrap: normal;
+    z-index: 0;
```
lubber-de, line 568: In relation to my below `@supports` suggestion, we should also put this into the extra wrapper and add it to the `:hover` selector only. This way, there won't be any possible sideeffects

src/definitions/modules/dropdown.less
```diff
@@ -565,13 +565,23 @@ select.ui.dropdown {
...
   }
 
   /* User Item */
   .ui.selection.dropdown .menu > .hidden.addition.item {
     display: none;
   }
 
+  /* User addition item */
+  .ui.dropdown > .menu > .addition.item:first-child {
```
lubber-de, line 577: we may make this feature optional anyway by using an necessary extra classname like "sticky" (save to use as long as "ui" is not also used inside the item, otherwise let's use "stuck" instead
```css
.ui.dropdown > .menu > .sticky.addition.item:first-child {
or
.ui.dropdown > .menu > .stuck.addition.item:first-child {
```

lubber-de (changes requested):
i added my comments here as an official review ;)

src/definitions/modules/dropdown.less
xamey: alright, thx for the review!
I defined a new variable, `@additionZIndex`, which is basically `@hoveredZIndex + 1`. We don't need to override hover z-index but we need the addition z-index to be gt than it. And I used it :)

src/definitions/modules/dropdown.less
```diff
@@ -572,6 +572,17 @@ select.ui.dropdown {
     display: none;
   }
 
+  /* User addition item */
+  @supports (position: sticky) {
```
lubber-de, line 576: I thought about this again: We should remove the @supports declaration again and instead add `position:-webkit-sticky `in addition to `position: sticky` . Why? To [support Safari versions up to 12](https://caniuse.com/?search=sticky) which is from 2018 and most probably still widely used (at least more than ie11), so resulting in

I also would prefer `stuck` instead of sticky to make sure it does not interfere with the sticky component and because the addition is definately stuck at the top of the menu, not just sticky (which basically means is dynamic and not always stuck)

I hope you are ok with that (?) 🙂 

```css
    .ui.dropdown > .menu > .stuck.addition.item:first-child {
      position: -webkit-sticky;
      position: sticky;
      top: 0;
      box-shadow: @selectionVisibleBoxShadow;
      background: @menuBackground;
      z-index: @additionZIndex;
    }
  ```
We did it the same way for the new `stuck table` variant and added a hint in the docs that that feature needs a modern browser.

lubber-de (changes requested):
Please see my comment about removing supports and add -webkit-sticky

src/definitions/modules/dropdown.less
xamey: Yeah I'm completely OK, this is for sure more interesting to support Safari rather than ie11!
Modified to `stuck` too, understandable.

I'll add another hint in the doc as you made for the table this afternoon :)

lubber-de (approved):
LGTM

lubber-de:
@all-contributors please add @xamey for code

PR #2417, opened 2022-08-04
feat(transition): firefox needs reattached node to retrieve new anima…

### LINKED ISSUE #2412: [Accordion] Firefox 103 throws: Transition: Element is no longer attached to DOM
# Bug Report
After latest Firefox upgrade on 103 version Accordion feature throws error: Transition: Element is no longer attached to DOM. Unable to animate.  Use silent setting to surpress this warning in production. fade in .

Same behavior is on the Fomantic-UI site https://fomantic-ui.com/modules/accordion.html, and on similar jsfiddle code snippets

## Steps to reproduce
Go on the https://fomantic-ui.com/modules/accordion.html site, with latest Firefox, Firefox dev, and just use the accordion and open the inspection tool.

https://jsfiddle.net/d2s8nhw3/

## Version

2.7.0, 2.8.
full
Capture

lubber-de:
## Description
Latest Firefox 103 does not seem to provide the correct changed animationname in all cases.
This fix simply reattaches the temporary node to make sure firefox freshly computes the animationname.

## Testcase
- Use Firefox 103.x (latest)
- Open dev console
- close the first accordion
- reopen the accordion or any other closed accordion

### Broken
- Console says `Transition: Element is no longer attached to DOM. Unable to animate.  Use silent setting to surpress this warning in production. fade in <p class="transition hidden">`
- accordion does not show opened content

https://jsfiddle.net/lubber/3ybfs51n/2/

### Fixed
All fine 
https://jsfiddle.net/lubber/3ybfs51n/4/

## Closes
#2412

### DIFF
--- a/src/definitions/modules/transition.js
+++ b/src/definitions/modules/transition.js
@@ -698,6 +698,7 @@ $.fn.transition = function() {
                 .addClass(className.transition)
                 .css('animationName')
               ;
+              $clone.detach().insertAfter($module);
               inAnimation = $clone
                 .addClass(className.inward)
                 .css('animationName')
--- a/src/definitions/modules/transition.js
+++ b/src/definitions/modules/transition.js
@@ -704,6 +704,7 @@ $.fn.transition = function() {
                 .css('animationName')
               ;
               if(!displayType) {
+                $clone.detach().insertAfter($module);
                 displayType = $clone
                   .attr('class', elementClass)
                   .removeAttr('style')


IwanBurg:
Nice fix, it seems to work, but I've found an edge case with a problem. Lines 702-705 also use the css 'animationName', which gives us the same issue, but more rare. Adding another `$clone.detach().insertAfter($module);` at line 706 would fix this.

I didn't run into more errors, but missed some custom styling. I encountered this when using an 'accordion' with 'ui multiple dropdown' within, which already had an initial selection.

Visuals for the edge case:
Issue:
image

Should be:
image

lubber-de:
@IwanBurg Thanks for reviewing! 👍🏼 I will check and add your suggestion to the PR this evening!

lubber-de:
@IwanBurg Could you quickly provide a jsfiddle testcase so i quickly check your case as well?

IwanBurg:
@lubber-de Here is a jsfiddle: https://jsfiddle.net/9qko2zh8/
Issue is seen when selecting the other values. The initial value does not have the custom styling, while the newly selected values do. My proposed solution should be able to fix this.

lubber-de:
> @lubber-de Here is a jsfiddle: [jsfiddle.net/9qko2zh8](https://jsfiddle.net/9qko2zh8/) Issue is seen when selecting the other values. The initial value does not have the custom styling, while the newly selected values do. My proposed solution should be able to fix this.

Thanks for the jsfiddle, @IwanBurg !
Indeed your proposed solution fixes it. I updated the PR accordingly
here is your jsfiddle using the latest code
https://jsfiddle.net/lubber/a6estmg2/

exoego (approved):

PR #2422, opened 2022-08-09
fix(form): inverted form labels were black

### LINKED ISSUE #2421: [form] inverted label is not inverted
# Bug Report
When inverting a form, the label color does not seem to be inverted anymore.

## Steps to reproduce
1. Add an inverted form
2. The label color is black instead of white

## Expected result
The label color should be white.

## Actual result
The label color is still black.

## Testcase
https://jsfiddle.net/rz5u34gv/

## Screenshot (if possible)
image

## Version
2.9.0-beta.301+42e68bc

lubber-de:
## Description
form labels in inverted forms were not shown white but still black

## Screenshot
invertedformlabel

## Closes
#2421

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -69,11 +69,10 @@
   font-size: @labelFontSize;
   font-weight: @labelFontWeight;
   text-transform: @labelTextTransform;
-  &:not(.button) {
-    color: @labelColor;
-  }
 }
-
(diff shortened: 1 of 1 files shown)


exoego (approved):
LGTM

exoego:
Note) Possible XSS are fixed in https://github.com/fomantic/Fomantic-UI/pull/2423

PR #2423, opened 2022-08-09
fix(chore): prevent possible xss

lubber-de:
## Description
In some edge cases a class or iconname, given via settings, might get such values out of uncontrolled input.
For such cases we need to dequote the values.

## Fixes
https://github.com/fomantic/Fomantic-UI/security/code-scanning/56
https://github.com/fomantic/Fomantic-UI/security/code-scanning/57

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -4248,7 +4248,8 @@ $.fn.dropdown.settings.templates = {
     var
       placeholder = select.placeholder || false,
       html        = '',
-      escape = $.fn.dropdown.settings.templates.escape
+      escape = $.fn.dropdown.settings.templates.escape,
+      deQuote = $.fn.dropdown.settings.templates.deQuote
     ;
     html +=  '<i class="dropdown icon"></i>';
     if(placeholder) {
@@ -4257,7 +4258,7 @@ $.fn.dropdown.settings.templates = {
     else {
       html += '<div class="text"></div>';
     }
-    html += '<div class="'+className.menu+'">';
+    html += '<div class="'+deQuote(className.menu)+'">';
     html += $.fn.dropdown.settings.templates.menu(select, fields, preserveHTML,className);
     html += '</div>';
     return html;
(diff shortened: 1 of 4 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -4257,7 +4258,7 @@ $.fn.dropdown.settings.templates = {
     else {
       html += '<div class="text"></div>';
     }
-    html += '<div class="'+className.menu+'">';
+    html += '<div class="'+deQuote(className.menu)+'">';
```
exoego, line 4261: Note) `deQuote` converts `"` into `&quot;` which might break styling.
But I assume it is OK if class name contains `"` it is invalid and users should be aware of before production.

src/definitions/modules/embed.js
```diff
@@ -677,7 +677,7 @@ $.fn.embed.settings = {
           src += '?' + parameters;
       }
       return ''
-        + '<iframe src="' + src + '"'
+        + '<iframe src="' + src.replace(/"/g,"") + '"'
```
exoego, line 680: How about defining a helper?

src/definitions/modules/rating.js
```diff
@@ -542,7 +542,7 @@ $.fn.rating.settings = {
         html = ''
       ;
       while(icon <= maxRating) {
-        html += '<i class="'+iconClass+' icon"></i>';
+        html += '<i class="'+String(iconClass).replace(/"/g,"")+' icon"></i>';
```
exoego, line 545: How about defining a helper?

exoego (approved):
LGTM
Added some nits

src/definitions/modules/dropdown.js
lubber-de: No, it doesn't, because that only happens when a second parameter "encode" is provided. Default is false/undefined, so if only `deQuote()` is used the double quotes are simply getting removed. Only `deQuote('bla',true)` converts the double quotes into an htmlentity
https://github.com/fomantic/Fomantic-UI/blob/77bb2f281d60113cb684071d66d308670e4ff770/src/definitions/modules/dropdown.js#L4218-L4221

src/definitions/modules/rating.js
lubber-de: Implemented by fb905d2ec7f615e1c71f986691e6ab1208659669 (identical to dropdown)

src/definitions/modules/embed.js
lubber-de: Implemented by fb905d2ec7f615e1c71f986691e6ab1208659669 (identical to dropdown)

Issue #2428, opened 2022-08-15
[modal] Unable to dismiss dropdown / calendar / search using escape key without closing the modal


mat128:
# Bug Report
It is not possible to close a calendar picker via the escape key when it is embedded in a modal.
It appears the issue is not limited to the calendar picker. The dropdown menu, the checkbox and the search components are also affected. The components behave normally when used directly inside a page, outside of a modal.

It should be possible to use the escape key to dismiss an open dropdown, or a date picker.
Such behavior is natural when using native components.

**Edit**: It should be possible to close the modal when interaction is not on a dropdown / calendar / search

## Steps to reproduce
1. Add a dropdown/calendar/search to a modal
2. Open the modal
3. Open the dropdown
4. Hit the escape button

## Expected result
The dropdown/calendar/search popup is closed. The modal stays visible.

## Actual result
The modal is closed.

## Testcase
https://jsfiddle.net/mat128/e5of9mdg/83/

## Screenshot (if possible)
![fomantic-escape](https://user-images.githubusercontent.com/234168/184693176-2dc3b0a2-c1eb-4099-a3c2-a4726adfa3bc.gif)

## Version
2.8.8

lubber-de:
Use `closable: false` for the modal, which prevents that. However it will also prevent closing the modal by clicking on the dimmer.
See your adjusted jsfiddle here https://jsfiddle.net/lubber/khdqz97p/

PR #2429, opened 2022-08-15
fix(modal): avoid propagation of escape to modal

mat128:
## Description
Allow using escape to dismiss calendars, dropdown, checkbox or search popups.
Previously, the escape key events (keydown, keyup) bubbled through the whole component tree.
This closed the modal when the escape key was used to dismiss a popup.

Fixing the issue requires all components to use the same event (keydown or keyup).
Standardize on keydown. This way propagation can be stopped prior to the modal closing.

This approach allows different components to "eat" the keystroke, while leaving the event
bubble up when required.

https://www.mutuallyhuman.com/blog/keydown-is-the-only-keyboard-event-we-need/

## Testcase
https://jsfiddle.net/lubber/hvbq6ke0/1/

## Closes
- #2428

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -593,8 +593,8 @@ $.fn.calendar = function(parameters) {
           },
           keydown: function (event) {
             var keyCode = event.which;
-            if (keyCode === 27 || keyCode === 9) {
-              //esc || tab
+            if (keyCode === 9) {
+              //tab
               module.popup('hide');
             }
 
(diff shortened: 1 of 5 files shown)


lubber-de:
Wow, that was a quick fix. 👍🏼  I added a testcase based on your jsfiddle from #2428 but using your PR code

lubber-de (approved):
LGTM

lubber-de:
@all-contributors please add @mat128  for code

PR #2436, opened 2022-08-23
fix(sidebar): scrolllock and returnscroll

lubber-de:
## Description
The `scrollLock` and `returnScroll` setting was not working which this PR fixes
It also adopts the same scrollbar vanish code from modal to avoid moving page content when the scrollbar vanishes

>  When this is merged, remember to also adjust the new flyout component, as this uses the same selectors for .pushable and needs to be equal

#### ScrollLock
- open sidebar
- dimmed pusher area is still scrollable as scrollbar remains visible
https://jsfiddle.net/lubber/dpu8gy65/1/

#### returnScroll
- open sidebar
- scroll pusher content
- close sidebar
- scroll position does not change
https://jsfiddle.net/lubber/dpu8gy65/18/

#### ScrollLock
- dimmed pusher area is not scrollable anymore and scrollbar is hidden just as in modal
https://jsfiddle.net/lubber/dpu8gy65/23/

#### returnScroll
- scrollbar returns to the position where it was when the sidebar was opened
https://jsfiddle.net/lubber/dpu8gy65/19/

## Closes
#2406

PR #2437, opened 2022-08-23
fix(popup): detailed rtl check

lubber-de:
## Description
I added  more detailed check for rtl. Currently only the module is checked. However, if body or the surrounding context would be defined as rtl then the module itself is also displayed as rtl, so this PR enhances the check (i already did the same inside the [sidebar PR](#2436 ) and [modal PR](#2433 ))


exoego (approved):
LGTM

PR #2438, opened 2022-08-23
fix(chore): forward debug settings to transition on internal use

lubber-de:
## Description
Some modules use the transition module internall to do animations. But some missed to provide all debug settings from the parent module.
Especially the `silent` setting was not forwarded.
So whenever someone tried this
```javascript
$(el).dropdown({
  silent: true
});
```
This was ignored by the transition module (in case it logs errors), because the default setting of silent is false in all modules.


exoego (approved):
LGTM

PR #2439, opened 2022-08-28
fix(chore): variable declaration and usage fixes and optimization

lubber-de:
## Description
Many variable usages were wrong/unused/buggy which this PR fixes or optimizes

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -1015,7 +1015,7 @@ $.api = $.fn.api = function(parameters) {
             response
           ;
           passedArguments = passedArguments || queryArguments;
-          context         = element         || context;
+          context         = context         || element;
           if(typeof query == 'string' && object !== undefined) {
             query    = query.split(/[\. ]/);
             maxDepth = query.length - 1;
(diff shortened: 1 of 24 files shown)


exoego:
How about introducing linter or static analyzer again?
It detects these kind of issues and prevents.

lubber-de:
> How about introducing linter or static analyzer again? It detects these kind of issues and prevents. I am happy to work on that.

An analyzer might be a good start. I used SonarLint locally to detect (some of) these changes. However many of the findings were false posisitves, so i would not auto adjust code by some bot but _always_ look at what a bot/linter would suggest first.

Regarding (automatic) linters: Again, i currently would not necessarly autoadjust things like codeformat/spacings/ curly braces etc. as there is alwas room for (in my pov mostly unnessary discusssions (for example i personally dont like putting curly braces into the next line of if/functions, but other might think different), but for existing (working, "nice enough") code i would not care much. On the other hand i dont want to mess with people just because there contributions have a space missing or something. 
And a PR which would completely rearrange the whole codebase just because the code styling does not match the linters setting would behard to review (for the first time ;)

I wonder, if we can use the already existing codeql bot ?

exoego:
Linters can be configured only to check unused variables and ignore styling rules.

PR #2444, opened 2022-08-30
fix(modal): Hiding dimmer with multiple modals

### LINKED ISSUE #2441: [Modal] Dimmer not hiding if multiple modals animating
# Bug Report
When a modal is hidden, it checks for other modals and if there are no other active or animating modals it hides the modal dimmer. This happens on the hiding animation start. If two modals start hiding at the same time, neither of them will hide the dimmer. This leaves you with a dimmer with no modals

## Steps to reproduce
1. Create a modal.
2. Create another modal.
3. Hide the second modal.
4. While the second modal is animating with its hide, hide the first modal.

## Expected result
Modal dimmer should be hidden

## Actual result
The modal dimmer stay visible and active with no modals in it.

## Testcase
[https://jsfiddle.net/xmd9eht6/](https://jsfiddle.net/xmd9eht6/)

## Version
2.8.8

## Potential Solution
* Don't start hiding first modal until second modal is hidden. This is the current workaround I'm doing.
* Use the modal's `hide all` or `hide dimmer` functions.
* (Code change) When hiding a modal, instead of doing the check for other modals and hiding the dimmer on hide animation `onStart`, do all this on hide animation `onComplete`.

I can make a pull request if others think this would be an appropriate fix

mareeo:
## Description
This PR moves the check for other active/animating modals to hide the dimmer to the hide animation complete instead of the hide animation start.

## Testcase
[https://jsfiddle.net/xmd9eht6/](https://jsfiddle.net/xmd9eht6/)

## Screenshot
Original
Original

After change
Fix

## Closes
#2441

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -614,6 +614,9 @@ $.fn.modal = function(parameters) {
                     if( settings.keyboardShortcuts && !module.others.active() ) {
                       module.remove.keyboardShortcuts();
                     }
+                    if(!module.others.active() && !module.others.animating() && !keepDimmed) {
+                      module.hideDimmer();
+                    }
                   },
                   onComplete : function() {
                     module.unbind.scrollLock();
@@ -627,9 +630,9 @@ $.fn.modal = function(parameters) {
                       else {
                         $previousModal.find(selector.dimmer).removeClass('active');
                       }
-                    }
-                    if(!module.others.active() && !module.others.animating() && !keepDimmed) {
-                      module.hideDimmer();
+                      if(!module.others.active() && !module.others.animating() && !keepDimmed) {
+                        module.hideDimmer();
+                      }
                     }
                     if($.isFunction(settings.onHidden)) {
                       settings.onHidden.call(element);
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -611,9 +611,6 @@ $.fn.modal = function(parameters) {
                   duration    : settings.transition.hideDuration || settings.duration,
                   useFailSafe : true,
                   onStart     : function() {
-                    if(!module.others.active() && !module.others.animating() && !keepDimmed) {
-                      module.hideDimmer();
-                    }
                     if( settings.keyboardShortcuts && !module.others.active() ) {
                       module.remove.keyboardShortcuts();
                     }
@@ -630,6 +627,9 @@ $.fn.modal = function(parameters) {
                       else {
                         $previousModal.find(selector.dimmer).removeClass('active');
                       }
+                    }
+                    if(!module.others.active() && !module.others.animating() && !keepDimmed) {
+                      module.hideDimmer();
                     }
                     if($.isFunction(settings.onHidden)) {
                       settings.onHidden.call(element);


lubber-de:
Thanks for the PR, it basically works. However, hiding the dimmer after (even if just one) modal is hidden, is somehow disturbing the design/flow of the modal. 
What if we just add the same code to onComplete (as you did but inside the `if ( settings.allowMultiple ) {` condition because this issue only happens on multiple modals, isn't it?) , but keep the code at onStart as well? Doesn't this also fix your issue?

mareeo:
Good catch, the my change caused the modal hide animation to start only once the modal was hidden. I've made the change you suggested and it does appear to work. If this edge case is reached, the dimmer will still start to hide once the modal is hidden, but I guess that's better than previous behavior of the dimmer not hiding at all.

One potential catch with this change, if you display a single modal with `allowMultiple` and then hide it, `hideDimmer()` is now called on the modal hide animation start and complete.  This doesn't break anything, but it does result in the debug message "Animation is already occurring, will not execute repeated animation" if you have debug mode enabled.

I'd like to add a condition to prevent this, but I'd need some way to know if the dimmer is animating out as opposed to just simply animating (which is what it's doing now by calling the dimmer's `is animating`). I could check if the dimmer has the `out` class to determine if it's animating out and prevent duplicate calls in modal's `hideDimmer()`.

lubber-de (changes requested):
Please see my comment.
Regarding "avoid the debug info by checking for the 'out' class"
Well, the hideDimmer function does also support _not_ using css classes but JS animation, so if someone possibly uses `dimmerSettings: {useCss:false}` checking for the 'out' class won't work.

The issue you mention actually comes from the "hideAll" function rather than your new change in the onComplete method.

So, instead, i suggest to check if "hide dimmer" is still in progress.
You can accomplish this by  centrally adjusting the `hideDimmer` method as follows :

### First: create new module wide variable `isHiding`
Snippet:
```javascript
    time           = new Date().getTime(),
    performance    = [],
    isHiding       = false,
```

### Second: Let the dimmers hide method return false in case the dimmer cannot be hidden

https://github.com/fomantic/Fomantic-UI/blob/86121efcc4508824f1f3ec21fffd1c8b3131ad44/src/definitions/modules/dimmer.js#L218

-> change to
```javascript
return false;
```

https://github.com/fomantic/Fomantic-UI/blob/86121efcc4508824f1f3ec21fffd1c8b3131ad44/src/definitions/modules/dimmer.js#L225-L226

-> change to
```javascript
  module.debug('Dimmer is not visible'); 
  return false;
 }
```

### Third: Adjust the hideDimmer method to check for that variable and take care of falsy return like this

Snippet
```javascript
        hideDimmer: function() {
          if(!isHiding && ($dimmable.dimmer('is animating') || ($dimmable.dimmer('is active'))) ) {
            isHiding = true;
            module.unbind.scrollLock();
            if($dimmable.dimmer('hide', function() {
              isHiding = false;
              module.restore.bodyMargin();
              module.remove.clickaway();
              module.remove.screenHeight();
            }) === false) { isHiding = false;};
          }
          else {
             module.debug('Dimmer is not visible cannot hide');
            return;
          }
        },
```

lubber-de:
@mareeo Will you continue to work on this?

mareeo:
I plan to, just limited on time currently. Thanks for the detailed reply and suggestion, I'll give it a go once I have time

PR #2451, opened 2022-09-10
fix(sidebar): dont resize on ios touchmove

### LINKED ISSUE #2449: [iOS] Sidebar touch event breaks layout
# Bug Report
Hi,

I've been making an app and noticed, only on iOS apparently (Safari works), the sidebar on my project when I touch/drag it back brings the pusher div over and breaks the layout.

I went over and noticed it also happens on the official website, so it's not any personal coding interference.

## Steps to reproduce

1. Go to fomantic-ui.com on iOS Safari
2. Open sidebar on the hamburger menu
3. Drag laterally on the sidebar

## Expected result
Expect it to bring back the pushed div over to where the sidebar is, default behaviour elsewhere.

## Actual result
GIF:

Sep-09-2022 17-44-10

Latest version of fomantic-ui or semantic-ui

lubber-de:
## Description
Safari on iOS 13.1+ seem to drag html containers on touchmove by default. At least this is happening for the sidebar.

I am not quite sure if this could break existing projects as the `pan-y` default limits the sidebars content to only that gesture. Thinking of only default menu items, this might be ok, but other content might need other gestures(?). 
The `set.ios()` function was already present but not used, just in case you wonder. 

## Testcase
- Run this on Safari 13+ on iOS only
- Open the sidebar
- Inside the open sidebar, drag to the left

### Broken
The sidebar gets shrinked
https://fomantic-ui.com/jsfiddle/#!lubber/7d2hxouf/44/

### Fixed
Nothing happens as expected. Sidebar keeps it's width
https://fomantic-ui.com/jsfiddle/#!lubber/7d2hxouf/43/

## Closes
#2449

### DIFF
--- a/src/definitions/modules/sidebar.js
+++ b/src/definitions/modules/sidebar.js
@@ -95,7 +95,9 @@ $.fn.sidebar = function(parameters) {
           module.debug('Initializing sidebar', parameters);
 
           module.create.id();
-
+          if(module.is.ios()) {
+            module.set.ios();
+          }
           transitionEvent = module.get.transitionEvent();
 
           // avoids locking rendering if initialized in onReady
--- a/src/definitions/modules/sidebar.less
+++ b/src/definitions/modules/sidebar.less
@@ -270,6 +270,9 @@ body.pushable > .pusher {
        iOS
 ---------------*/
 
+html.ios .ui.sidebar {
+  touch-action: @iOStouchAction;
+}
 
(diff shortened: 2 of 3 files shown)


lubber-de:
Set on hold/switch to draft because it does not seem to fix the whole isse as mentioned in https://github.com/fomantic/Fomantic-UI/issues/2449#issuecomment-1250084215

PR #2453, opened 2022-09-13
refactor(sidebar): remove un-used legacy code

ColinFrick:
This removes the duration property from the default settings.

It was used for legacy animation mode (jquery instead of css), but that feature was removed a long time ago (22293069205d2e68dbbe926481a99caa2682d1a0)


lubber-de (approved):
LGTM, welcome back 😃

PR #2442, opened 2022-08-30
feat(toast,modal): support short notation $.

lubber-de:
## Description

This PR adds jquery short notation support (`$.`) to toast and modal which are made dynamically out of js properties and thus are not bound to an existing dom node.

Previously you always had to select body (which still works) when no existing domnode template exists. Infact the body selector was completely ignored.

### Before notation (still supported)
```javascript
  $('body').toast({
    message: 'hello'
  });

  $('body').modal({
    autoShow: true,
    title: 'hello'
  });
```

### After notation
```javascript
  $.toast({
    message: 'hello'
  });

  $.modal({
    autoShow: true,
    title: 'hello'
  });
```

## Testcase
https://jsfiddle.net/lubber/7s9ouapq/2/

PR #2455, opened 2022-09-16
feat(flyout): support short notation $.

prudho:
## Description

This PR is an adaptation from @lubber-de's #2442.

It adds jquery short notation support (`$.`) to flyout which is made dynamically out of js properties and thus are not bound to an existing dom node.

Previously you always had to select body (which still works) when no existing domnode template exists. Infact the body selector was completely ignored.

### Before notation (still supported)
```javascript
  $('body').flyout({
    title: 'Important Notice',
    content: 'You will be logged out in 5 Minutes',
    actions: [{
      text: 'Alright, got it',
      class: 'green',
      icon: 'save'
    }]
  });
```

### After notation
```javascript
  $.flyout({
    title: 'Important Notice',
    content: 'You will be logged out in 5 Minutes',
    actions: [{
      text: 'Alright, got it',
      class: 'green',
      icon: 'save'
    }]
  });
```

### DIFF
--- a/src/definitions/modules/flyout.js
+++ b/src/definitions/modules/flyout.js
@@ -23,7 +23,7 @@ window = (typeof window != 'undefined' && window.Math == Math)
     : Function('return this')()
 ;
 
-$.fn.flyout = function(parameters) {
+$.flyout = $.fn.flyout = function(parameters) {
   var
     $allModules     = $(this),
     $window         = $(window),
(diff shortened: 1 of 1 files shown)


prudho:
Also please note, the removed `repaint()` function was an artifact from the `sidebar` module which was useless in this module.

prudho:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/369.

src/definitions/modules/flyout.js
```diff
@@ -75,8 +75,10 @@ $.fn.flyout = function(parameters) {
         $pusher              = $context.children(selector.pusher),
         $style,
 
+        isFlyoutComponent    = $module.hasClass(namespace),
```
lubber-de, line 78: The namespace variable is customizable and not necesarily identical to the default css classname. In case one changes that to 
```javascript
settings: {
	namespace: 'flyoutCustom'
}
```
the module is not detected as a flyout component anymore as a `flyoutCustom` css class does not exists.

lubber-de:
```suggestion
        isFlyoutComponent    = $module.hasClass('flyout'),
```

lubber-de (changes requested):
Please see my little comment 😉

lubber-de (approved):
I commited my suggested little change myself, i hope you don't mind

PR #2463, opened 2022-09-25
Remove dist/ files from git

mvorisek:
Currently, `dist/` files are comming from the latest stable release and storing them in versioned git is useless and it complicates development. Local changes followed by a local rebuild should not imply additional (git tracked) changed files,

This PR removes `dist/` directory from git and adds `.gitignore` exclude pattern for it. In release scripts (nightly/stable), the exclude pattern is removed so `dist/` files are pushed to npm as before.


mvorisek:
When building the `dist/` data using https://fomantic-ui.com/introduction/getting-started.html#installing-via-npm instructions, `semantic.css` is not built (only the minified file is built).

I would expect build result 1:1 same with the official release, ie. incl. `semantic.css`.

Also, the `node_modules/fomantic-ui/package.json` version is `"version": "2.8.8"`. When I require `"fomantic-ui": "^2.9.0-beta.319"` nightly package, I would expect that version everywhere. Nightly tags should be build exactly the same as the stable releases are. Develop branch version should be something like `2.9.0-dev`.

jamessampford:
It was mentioned/suggested on Discord that generating beta docs would be a good idea (which IMO would happen when there are updates to either this repo for nightly or the docs repo for nightly) - I would have thought updating the `dist/` dir would also be a good idea as part of the nightly, seeing that the beta docs would need the updated files, as well as everything else being up to date

lubber-de:
We will keep the /dist folder. Especially if we would update it via the nightly script and create a nightly beta tag. So, the /dist folder would be present / recreated every night and that would mean to also remove the dist folder again in a separate commit , which is a mess of git history changes.

Not to commit files inside the/dist folder is mentioned in the contributing.md already
https://github.com/fomantic/Fomantic-UI/blob/0740de88774bb3171b0badb2481aea5bfe07461a/CONTRIBUTING.md?plain=1#L47

If you would like to help out for that, you may try to adjust the [nightly.js](https://github.com/fomantic/Fomantic-UI/blob/develop/scripts/nightly-version.js) script instead
I did not reeach yet, but maybe we can setup some bot which will remove any changes from the /dist folder from any incoming PR (it it's not meant for a release merge..)

> When building the dist/ data using [fomantic-ui.com/introduction/getting-started.html#installing-via-npm](https://fomantic-ui.com/introduction/getting-started.html#installing-via-npm) instructions, semantic.css is not built (only the minified file is built).

Cannot reproduce. Using `npx gulp build` or `yarn build` creates both files semantic.css/js and semantic.min.css/js for me

lubber-de:
> When the dist/ content is built for a (stable & nightly) release, the /dist/ line in .gitignore should be removed and files added to git.

This is something i would kind of agree with (instead of deleting the /dist folder) but the above mentioned nightly script should be adjusted accordingly nevertheless

mvorisek:
> So, the /dist folder would be present / recreated every night and that would mean to also remove the dist folder again in a separate commit , which is a mess of git history changes.

Given the nightly tags are not commited to the develop branch, what is the problem with history changes?

I would expect dist/ directory to be always in sync with the develop/ changes (and asserted by CI) or not present at all. If dist needs to be commited on stable release, just build, remove dist/ from .gitignore, add all dist/ files and tag. When history needs to be compared, just compare one release vs. another.

> Cannot reproduce. Using `npx gulp build` or `yarn build` creates both files semantic.css/js and semantic.min.css/js for me

ignore this comment, probably some misconfiguration on my site when debugging another issue

> It was mentioned/suggested on Discord that generating beta docs would be a good idea (which IMO would happen when there are updates to either this repo for nightly or the docs repo for nightly) - I would have thought updating the `dist/` dir would also be a good idea as part of the nightly, seeing that the beta docs would need the updated files, as well as everything else being up to date

Yes, docs should be moved in this repo and versioned together with the function changes/commits.

mvorisek:
@lubber-de I still belive development/contributions would be much easier if the build can be run locally /wo affecting any (already added) git file. I propose the following git workflow:

image

as the `dist/` is outdated in the `develop` branch anyway, I do not see any advantage of having `dist/` files in the develop branch

git is also smart enough to dedup the files between all commits/tag and also show the changes correctly (if you compare tag vs. another tag /w `dist/` files present)

mvorisek:
> > When the dist/ content is built for a (stable & nightly) release, the /dist/ line in .gitignore should be removed and files added to git.
> 
> This is something i would kind of agree with (instead of deleting the /dist folder) but the above mentioned nightly script should be adjusted accordingly nevertheless

Fixed. The `/dist/` line in `.gitignore` is removed before the package is pushed to the registry, so the built `dist/` is pushed correctly.

lubber-de:
Your PR still deletes the /dist folder which i told you we won't do. Instead what you wanted is to have the dist folder being updated by the nightly script _in the git repo_. But this is not what your changes do. Your changes just make sure that the npm package includes the generated dist folder (which is fine) and that /dist is ignored for commits (which is also fine) but it actually does not update the git repo itself.

mvorisek:
> Your PR still deletes the /dist folder which i told you we won't do.

I know, but why? What is the purpose of having stable/nightly `dist/` files in develop branch (and in git in general)?

I have described the reason (in updated) PR description, I think we should really not have these files in develop branch, local rebuild should really not imply additional git tracked changes files locally.

lubber-de:
1. We are following gitflow principle, so  the master branch will _always_ contain the same content as the develop branch (at point of merge), because master is always a merge of develop and should not have dedicated commits (which would be needed if only the master branch should contain /dist). 
2. There are still a lot of people not using cdn/npm or do/can not build locally but just want to download the repo as it is ready for usage. Those people also want to have the /dist folder ready for usage without the need to compile themselves. Yes, that's exactly the point where we would need improvement for the develop branch, as the /dist inside the develop branch currently does not reflect the changes in /src as you initially pointed out.

mvorisek:
What about introducing a `nightly` branch?

- in `develop`, there will be no `dist/` (thus a local rebuild will not imply additional (git tracked) changed files), point I solve by this PR now
- in `nightly` there will be merged in `develop` + unignored `dist/` files rebuilt by CI, point you want to address
- in `master` the same but stable release instead of nightly release, point we support currently by updating `dist/` once per stable release

lubber-de:
Your main concern was that you accidently had /dist updated files in another PR some weeks ago, where i asked you to remove them.
You solved that situation in this PR here by adding /dist to .gitignore but also making sure let the nightly/release script temporary  removes it from gitignore right before publishing.

Why don't we start with that first? 🙂

jamessampford:
> What about introducing a `nightly` branch?
> 
> * in `develop`, there will be no `dist/` (thus a local rebuild will not imply additional (git tracked) changed files), point I solve by this PR now
> * in `nightly` there will be merged in `develop` + unignored `dist/` files rebuilt by CI, point you want to address
> * in `master` the same but stable release instead of nightly release, point we support currently by updating `dist/` once per stable release

I’d stick with just a develop` and `master`, where `develop` is always updated, even the `dist/` folder, keeps everything consistent

mvorisek:
> Your main concern was that you accidently had /dist updated files in another PR some weeks ago, where i asked you to remove them.

It may sounds so, but the main reason is: compile should not change any git (non-gitignored) file. This is industry standard practice.

Also there is no rational reason, as done currently, to store stable/compiled release together with develop changes.

This PR removes the `dist/` gitignore line before the files are pushed to npm, so npm nightly/stable releases continue to work as before.

With this reasoning I belive this PR can be merged safely.

PR #2464, opened 2022-09-25
fix(emoji): removed unavailable twemoji resources

mvorisek:
## Description

some defined twemoji codes has no resources, these codes must be removed

- `1f441-fe0f-200d-1f5e8-fe0f` twemoji - related with https://github.com/twitter/twemoji/issues/253
- `digit_xxx` - just use `xxx`, 23, 2a, 30 - 39 codes have no twemoji resources

discovered by our atk4/ui build script, where we resolve all CSS resources in CI

if these variables are generated via some script, please update the script accordingly


lubber-de (changes requested):
Thanks for bringing this to our attention in time before 2.9.0 gets released 😄 
Please see my little comment about the changes inside the joypixels file.
Everything regarding tweemoji / default theme LGTM 🙂

lubber-de:
in 2.8.8 digit_{digit} resolved to the same value as only {digit} does.
Completely removing the digit_{digit} from the tweemoji /default theme would make this a breaking change.
I suggest to keep the digit_{digit} entries but use the same values (3x-20e3) as for the non digit_ values

src/themes/default/elements/emoji.variables
```diff
@@ -3269,18 +3269,6 @@
...
-  33: digit_three;
-  34: digit_four;
-  35: digit_five;
-  36: digit_six;
-  37: digit_seven;
-  38: digit_eight;
-  39: digit_nine;
-  23: pound_symbol;
-  2a: asterisk_symbol;
   26a7: transgender_symbol;
```
lubber-de, line 3272: Instead of removing i suggest to reuse the values from the non digit_ entries as it was before 2.9.0

src/themes/twitter/elements/emoji.variables
```diff
@@ -3259,18 +3259,6 @@
...
-  32: digit_two;
-  33: digit_three;
-  34: digit_four;
-  35: digit_five;
-  36: digit_six;
-  37: digit_seven;
-  38: digit_eight;
-  39: digit_nine;
-  23: pound_symbol;
-  2a: asterisk_symbol;
```
lubber-de, line 3273: Instead of removing i suggest to reuse the values from the non digit_ entries as it was before 2.9.0

```less
  30-20e3: digit_zero;
  31-20e3: digit_one;
  32-20e3: digit_two;
  33-20e3: digit_three;
  34-20e3: digit_four;
  35-20e3: digit_five;
  36-20e3: digit_six;
  37-20e3: digit_seven;
  38-20e3: digit_eight;
  39-20e3: digit_nine;
  23-20e3: hash_symbol;
  2a-20e3: asterisk_symbol;

```

I also saw, that "pound_symbol" is actually a "hash_symbol"  (23-20e3)

src/themes/joypixels/elements/emoji.variables
```diff
@@ -3267,18 +3267,6 @@
...
-  0031: digit_one;
-  0032: digit_two;
-  0033: digit_three;
-  0034: digit_four;
-  0035: digit_five;
-  0036: digit_six;
-  0037: digit_seven;
-  0038: digit_eight;
-  0039: digit_nine;
-  0023: pound_symbol;
```
lubber-de, line 3280: This is actually a 'hash_symbol', please rename

src/themes/joypixels/elements/emoji.variables
```diff
@@ -3267,18 +3267,6 @@
...
-  002a: asterisk_symbol;
```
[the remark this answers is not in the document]
mvorisek: I will revert joypixels changes. I removed it to offer same set of names.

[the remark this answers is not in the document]
mvorisek: @lubber-de I diffed the available names in twemoji and joypixels and all of them (almost to my suprise) are 1:1 (same ordered set). Is it really wanted to keep more joypixels icons available?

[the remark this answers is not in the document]
mvorisek: UPDATE: resolved by https://github.com/fomantic/Fomantic-UI/pull/2464#discussion_r979421077

src/themes/default/elements/emoji.variables
mvorisek: for joypixels, should we keep the 2 versions (digit_zero vs. zero), or map digit_zero to zero?

src/themes/twitter/elements/emoji.variables
mvorisek: it is a map, how can I add 2 names for one key? is the small BC worth it?

src/themes/default/elements/emoji.variables
lubber-de: both for joypixels as joypixels really offers two different emojis
for tweemoji/default map digit_zero to value of zero

src/themes/twitter/elements/emoji.variables
lubber-de: uhm...you've got a valid point here, my bad.
Ok then lets use the PR as you intentially did it

src/themes/default/elements/emoji.variables
lubber-de: forget my comment, you are right, lets skip that

lubber-de (approved):
LGTM

PR #2465, opened 2022-09-25
fix(dropdown): generated id should not start with a number

mvorisek:
introduced in https://github.com/fomantic/Fomantic-UI/pull/2409

this change was failing our strict tests as we require ID to not start with a number

no BC break as no stable release shipped yet

ping @Inselhopper, you are the original author, can the ID be anything or is some special format required?


lubber-de (approved):
LGTM, all other modules generating ids (modal, flyout, toast) also have an underscore prefix.

PR #2466, opened 2022-09-25
fix(flag): ensure unique flag and a-z class names

lubber-de:
## Description

- flag class generation sometimes generated identical selectors.
- classnames had diacritic letters
- crosseds is not a flag in the manner of country/landscape/grouping flags

## Closes
https://github.com/fomantic/Fomantic-UI/pull/2357#discussion_r979423076


src/themes/default/elements/flag.variables
```diff
@@ -17,19 +17,19 @@
...
-    class: transgender;
+    class: false;
     aliasClass: false;
     aliasClass2: false;
   };
   @1f3f3-fe0f-200d-1f308: {
     countrycode: rainbow;
-    class: rainbow;
-    aliasClass: gay_pride;
+    class: gay_pride;
```
mvorisek, line 26: any idea why this generated `i.flag.gay.pride::before` selector? (not tested this exact PR)

lubber-de: Because the code assumes an underscrore to be a word/classname split. (Same for "cayman_islands") following the original idea of SUI to have "spoken sentences words as classnames"
However, i just named it "pride" and added another alias LGBT instead by bb3281735f72040b030e068a4d61ab208a51ee6c according to https://emojiterra.com/rainbow-flag/ or  https://emojipedia.org/rainbow-flag/

mvorisek (approved):
LGTM

PR #2470, opened 2022-09-28
feat(icon): customizable icons by using maps

lubber-de:
## Description
This PR moves all icon declarations out of the overrides file and into LESS maps, so the icon selectors are not hardcoded anymore but rendered at compile time.

This now offers easy adjustment of (un)-wanted icons by simpy adjusting the related `@icon-map` maplist in a possible custom themes icon.variables

Previously that wasn't really possible, as the default theme overrides were always included, so even if your custom theme only needed a handful of icons you always got the whole set anyway.

I did the same for emoji and flag already.

The related script for self creation is adjusted in https://github.com/fomantic/create-fomantic-icons/pull/184


prudho (approved):
LGTM

lubber-de:
Doc hint added by https://github.com/fomantic/Fomantic-UI-Docs/pull/377

PR #2474, opened 2022-09-30
fix(dropdown): keep original values, also on remove and more fixes

### LINKED ISSUE #2246: [dropdown] is corrupting data (/w multiple enabled & allowAddition)

lubber-de:
## Description
Fixed the following issues:
- preset values inside an already existing hidden dropdown input field containing characters like < > " ' were encoded during setup on non-select dropdowns and thus added twice inside the hidden input
- removing selected multiple entries encoded the whole selected string on non-select dropdowns
- entering the same value in multiple search dropdowns as already provided as "values"  on initialize time were ignored/not removed on enter/delimiter
- pressing enter on such already existing values forced a form submit
- placeholder text was encoded twice when preserveHTML was set to false
- when every item was filtered and reopened again via backspace deletion some css was missing for the menu
- hiding menu when all items are filtered left the visible class which prevents another dropdown to be shown with higher z-index
- reshow menu when all non selected items were filtered and an item was selected

## Testcase
- Watch console and see initial form input field
- remove one field and watch the hidden input
- enter "a" and press comma
- enter "a" and press enter

### Broken
https://jsfiddle.net/lubber/61u0xpL8/13/
- the username email is encoded and added to the existing value
- removing one entry will entitiy encode all other values
- pressing comma does nothing, it does not even remove the value 
- pressing enter triggers a form submit

### Fixed
https://jsfiddle.net/lubber/61u0xpL8/20/
- the username email is kept original
- after removing one item the hidden input still contains the untouched values as before
- pressing comma removes the search entry as the value is already selected
- pressing enter also removes the search entry and does not trigger a form submit

## Closes
#2246


mvorisek:
> fixed / https://jsfiddle.net/lubber/61u0xpL8/14/

when `allowAdditions: true` is set, there are no dropdown options (when some value from the dropdown is removed and the input is clicked), the following is shown:

image

when `allowAdditions: true` is removed (and when some value from the dropdown is removed and the input is clicked), the following is shown:

image

I would expect the options present even with `allowAdditions: true`. Or is this difference intentional?

lubber-de:
> when `allowAdditions: true` is set, there are no dropdown options (when some value from the dropdown is removed and the input is clicked), the following is shown:
> I would expect the options present even with `allowAdditions: true`. Or is this difference intentional?

Fixed in the latest commit. See https://jsfiddle.net/lubber/61u0xpL8/19/

> also how/when is the `preserveHTML: true` useful? I tried [jsfiddle.net/2kg60xqb](https://jsfiddle.net/2kg60xqb/) (forked from [jsfiddle.net/lubber/61u0xpL8/14](https://jsfiddle.net/lubber/61u0xpL8/14/)) but the `<b>` is ignored.

It is not ignored. But the label class already has a bold value set for the whole label text, so `<b></b>` has no effect.
To see the difference i added a `` instead (not the way FUI expect it, but for educational reasons only here 😉 )

### preserveHTML: true
https://jsfiddle.net/lubber/y9td3car/2/
image

### preserveHTML: false
https://jsfiddle.net/lubber/y9td3car/3/
image

> also when there are multiple dropdown inputs and a custom value is added to the dropdown below, when the dropdown above is opened and the options are listed, the dropdown below is displayed over the list from the 1st/above dropdown:

Fixed in latest commit. See https://jsfiddle.net/lubber/2s6y87kf/

mvorisek:
another small issue - when a custom value is added (write some text and press enter), then when the dropdown field is clicked, the options are not opened - currently user must unfocus (click somewhere else) and then click the field again (then the options are opened)

lubber-de:
> another small issue - when a custom value is added (write some text and press enter), then when the dropdown field is clicked, the options are not opened - currently user must unfocus (click somewhere else) and then click the field again (then the options are opened)

Fixed in latest commit, see https://jsfiddle.net/lubber/61u0xpL8/20/

> `preserveHML: true` - [jsfiddle.net/aguch8p4](https://jsfiddle.net/aguch8p4/) - `Username` cannot be removed

Yes, because the <> creates invalid html making the close icon not an direct child of the label anymore, thus it is not found by the internal selector for binding the close event. If you provide <> in the hidden input field and those are not part of existing items and you also use allowAdditions, you really should use preserveHTML: false as such values are understood as custom additions. 
Not entity encoding use cases values was the initial fix of this PR, anything else deserves a separate PR for such use case as i really want to get 2.9.0  finished now :)

I think i now covered everything else you mentioned

mvorisek:
yes and thank you ❤️

PR #2475, opened 2022-10-02
fix(dropdown): hide menu on menu search blur

### LINKED ISSUE #2101: Floating dropdown’s menu can be escaped by keyboard without it closing
https://fomantic-ui.com/, the Dropdown section, open the filter one (the second one), press Tab twice, and you’re out of the popup and onto the next field, but the popup is still open until you click outside it, or perform some keyboard action that *does* close it (e.g. return to it and select an item with Enter).

lubber-de:
## Description
Whenever a dropdown has a search input inside the menu itself (for example a button dropdown), blurring this search field did not close the dropdown. 
Some docs examples (mainly the one on the homepage) has to be slightly adjusted for a proper fix as those did not have the `search` class or a `search input` wrapper being added. Will do that in a separate docs PR 

## Testcase
- open the dropdown by clicking on the button
- click into the searchfield (should be focussed already, so most probably not needed)
- press tab

### Broken
The next item in the dom gets highlighted, but the dropdown stays open
https://jsfiddle.net/lubber/xom7fy4z/2/

### Fixed
The dropdown gets closed
https://jsfiddle.net/lubber/xom7fy4z/6/

## Closes
#2101

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1156,7 +1156,7 @@ $.fn.dropdown = function(parameters) {
             },
             blur: function(event) {
               pageLostFocus = (document.activeElement === this);
-              if(module.is.searchSelection() && !willRefocus) {
+              if(module.is.searchSelection(true) && !willRefocus) {
                 if(!itemActivated && !pageLostFocus) {
                   if(settings.forceSelection) {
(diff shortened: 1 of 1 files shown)


lubber-de:
some more edge case fixes are needed, setting to draft

Issue #2481, opened 2022-10-03
[transition] "Element is no longer attached to DOM" animate error


mvorisek:
# Bug Report

<details>
  <summary>original report for fixed issue, see post below for a report of related but still present issue</summary>

## Steps to reproduce
1. open https://jsfiddle.net/65rgakbm/3/ (repro reply can be downloaded from https://github.com/atk4/ui/blob/292e18208b2b8f32551cc0ad037c93c6d87e9533/repro.json)
2. click "Save" button
3. click "Section 1" accordion section title (to expand it)
4. notice
   ![image](https://user-images.githubusercontent.com/2228672/193636513-a6c3f8bd-6b2a-4c8b-b6c0-f20136782e0b.png)
   in the browser console

Observations:

When accordion section is opened before "Save", the problem is not present.

When AJAX return non-toast like `alert`, example response:
```
{
    "success": true,
    "message": null,
    "atkjs": "alert(\"x\")",
    "portals": []
}
```
the issue is not present neither.

There is some atk4/ui js code involved - https://github.com/atk4/ui/tree/develop/js/src/services - but based on the stack trace, the issue seems to be in Fomantic-UI.

I did my best to be minimize the code needed to reproduce, but once the issue is identified, it can be probably simplified even more, ie. /wo AJAX toast reply needed.

Browser: Firefox

## Expected result

No error in browser console, accordion section must always close/open.

## Actual result

Browser console error, when accordion section is reopened, it is blank.

## Version

2.8.8
</details>

mvorisek:
fixed in latest 2.9.0 beta, also a duplicate of https://github.com/fomantic/Fomantic-UI/issues/2412

mvorisek:
reopening as #2412 fixed https://github.com/atk4/ui/issues/1821 but not https://github.com/atk4/ui/issues/1817, tested with `Fomantic UI - 2.9.0-beta.320+c4b3a74` version.

how to reproduce:

1. see https://github.com/atk4/ui/issues/1817 issue and hosted https://dev.agiletoolkit.org/demos/form-control/lookup.php public demo
2. copy these loaded
   ![image](https://user-images.githubusercontent.com/2228672/193902970-cba6e06d-45bf-4b5e-8f7c-b9fcdbaaa660.png)
   resources locally
3. copy the latest Fomantic UI 2.9.0 beta resources (semantic.js/css files)
4. in the public demo, fire the Save and save the AJAX response locally too
[...]

PR #2484, opened 2022-10-05
feat(form): debug a warning if field identifier is not found

lubber-de:
## Description
In #2481 it was tried to call `add prompt` to a form given an identifier that does not exist. Because of that a prompt cannot be found/used and eventually the transition which tries to fade in the prompt throws a warning, that the node is not attached to the DOM anymore.

This PR adds a debug log warnin in case the identifier is not found in the form, so developers can figure out whats wrong.


src/definitions/behaviors/form.js
```diff
@@ -642,6 +642,7 @@ $.fn.form = function(parameters) {
             if((t=$field.filter('[data-' + metadata.validate + '="'+ identifier +'"]')).length > 0 ) {
               return t;
             }
+            module.debug('Field identifier not found.', identifier);
```
mvorisek, line 645: this error seems to me much more severe

What is the purpose of returning an unattached `$('<input/>')`?

`module.debug` is silent when debug is not enabled - would a change to `module.error` be ok?

the console error would be like:

image

and guide the developer to the problem much faster

lubber-de: I changed it to an error by #2489 
I believe the purpose of returning the empty input was to not break all the following code, so we leave that as it is for now.

PR #1838, opened 2021-01-10
fix(dropdown): fix menu item text breaking issue

ko2in:
## Description
When the menu item contains too long text, the text doesn't break into the new line when there's no enough space to display in
single line and the text overflows the container.

Even though the text can be forced to break using `break-word: break-all`, it doesn't look good enough when there is an image representing next to the menu text.

menu

This PR fixes the menu item text breaking problem and also changes the container to flexbox to render the image, text, and the dropdown icon as flexbox items to align better.

Since the menu item has changed to the flexbox model and the dropdown icon and direct children elements of a menu item are appearing as flex items and the float property has no effect on flexbox, and so the float rules and variables for the dropdown icon are removed by this PR.

## Testcase
**Bug:** https://jsfiddle.net/ko2in/3sw01eLz/1/

**Fixed:** https://jsfiddle.net/ko2in/3sw01eLz/9/

## Screenshot
**Before:**
menu-bug

**After:**
menu-fix

## Closes
#1820

PR #2493, opened 2022-10-11
feat(menu): support descriptions in vertical menu

### LINKED ISSUE #2232: [menu] support descriptions in vertical menu
# Feature Request
There may be a need to use a description in a vertical menu; for example, potentially showing an extra item that is new (without using a label or an icon)

I believe having a similar display/working as dropdown will keep things consistant (allow both horizontal and vertical) - https://fomantic-ui.com/modules/dropdown.html#description

prudho:
## Description
This PR implements a description in the menu items, which can be right floated or under a text.

Under description can also be displayed in not vertical menu, but right floated no as it looks bloated right now (see screenshots).

## Screenshot
image

Inverted works too:
image

Description under text works as intended, but not the right floated one. Help is welcome 😃 
image

## Closes
#2232

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -97,6 +97,11 @@
   border-radius: @borderRadius 0 0 @borderRadius;
 }
 
+.ui.menu > .item.vertical {
+  display: flex;
+  flex-direction: column-reverse;
+}
+
 /* Border */
 .ui.menu .item::before {
   position: absolute;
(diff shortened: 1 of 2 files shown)


lubber-de:
@prudho What's the status of this PR?
Regarding the issue of a description inside a non vertical item: It seems to be solved when the description node is simply declared before the  text node 🙂 , so need to adjust your PR (missing related docs should provide an example to make this clear)

https://jsfiddle.net/lubber/8o5x3syq/

lubber-de:
If we yould make this the standard config (first text node, then description) also for `vertical item`, you just need to change "flex-direction" to "column" instead of "column-reverse".
Would be more consistent

```
.ui.menu > .item.vertical {
  display: flex;
  flex-direction: column;
}

See
https://jsfiddle.net/lubber/8o5x3syq/2/

lubber-de:
Ok i think you made it the same way it is done in dropdown menu items (first description, then text). This could be solvable there as well by setting `display: flex;` to dropdown menu item....BUT that would be a breaking change and people have to adjust the dropdown code, not worth the hassle
So we might keep the PR as it is and provide a docs info as suggested in my previous comment

ko2in:
I suggest to hold this PR for a while. I've already made a change for vertical menu as a flex item in my current branch based on my last PR #1838.

I'm still in progress of fixing merge conflicts since there were too many commits ahead of my PR.

PR #2512, opened 2022-10-18
fix(chore): support node12 on add

### LINKED ISSUE #2511: [scope] node does not support 12
# Bug Report

## Steps to reproduce
image

image

## Expected result
node support 12, because node 12 support windows7

## Actual result
node does not support 12

## Node version
```v12.22.10```

lubber-de:
## Description

The dependency to octokit was not bound to a major version making `yarn add fomantic-ui` break on node 12

## Closes
#2511

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@actions/core": "^1.6.0",
-    "@octokit/core": ">=3",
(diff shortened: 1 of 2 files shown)


mvorisek:
the CI job should detect such problem

also node v18 should be added to CI

PR #2527, opened 2022-10-30
feat(build): improve error message for when semantic.json is not found

dmp1ce:
## Description
Improve error message for when `semantic.json` cannot be found.

I was having an issue with my setup where `semantic.json` couldn't be found. Looking at require-dot-file, it seems `false` is returned when `semantic.json` isn't found. An error is printed only when the `MODULE_NOT_FOUND` error is thrown which wasn't happening in my case.


lubber-de (approved):
LGTM

PR #2534, opened 2022-11-01
feat(input): make invalid pseudo class optional

### LINKED ISSUE #2533: [scope] How to disable email input validation in newest version (2.9.0)
In previous versions this problem didnt happen, but in the latest one it happens. The minimum code to reproduce this is at:

https://jsfiddle.net/pksvnr0e/

It has a `email` input field and if I fill it with anything that is not a valid email, it becomes red. It's a nice feature but I would like to disable this, even better if it was possible to be disabled using the `build` in order to minimize the `semantic.js` and `semantic.css` files.

This "problem" started happening only in the latest version. 

Thank you in advance for all the great work!

lubber-de:
## Description
This PR makes automatic browser invalidation via `:invalid` pseudo class optional via new variables

```
@variationInputInvalid
@variationFormInvalid
```

## Closes
#2533

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -591,7 +591,7 @@
       border-radius: @formStates[@@state][borderRadius];
       box-shadow: @formStates[@@state][boxShadow];
     }
-    & when (@state=error) {
+    & when (@state=error) and (@variationFormInvalid) {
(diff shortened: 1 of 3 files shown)


exoego (approved):
LGTM

PR #2548, opened 2022-11-08
feat(calendar): support context setting

### LINKED ISSUE #2542: Context setting not working on Calendar popup
# Bug Report
The calendar popup is created inside the .input DIV what cause an issue if the popup overflows the form frame that has overflow settings different than "visible". I tried using the "context" setting to create the popup outside the form frame into the page body (similar to jquery ui datepicker), but this does not have any impact.

```
<div class="ui calendar">
      <div class="ui input">
	      <input type="text" name="test">
      </div>
</div>
```

image

## Steps to reproduce
https://jsfiddle.net/rf0puq4m/9/

## Expected result
I would like the popup be created into the page body

## Actual result
The popup is created inside the .input DIV

## Screenshot (if possible)
image

## Version
V2.9.0

lubber-de:
## Description
A calendar popup inside an overflowing container does not allow the popup to be shown completely, because it's still inside the overflowing container of the related calendar input field
You can now use a `context` setting which will move the calendar popup outside of the overflowing container (if the given context isnt an overflowing container itself...)

### Before
image

### After
image

### Before
https://jsfiddle.net/lubber/6evhd45q/

### After
https://jsfiddle.net/lubber/6evhd45q/6/

## Closes
#2542

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -1556,8 +1556,8 @@
     border-color: @invertedCellBorderColor;
   }
   & when (@variationTableDisabled) {
-    .ui.inverted.table tr.disabled td,
-    .ui.inverted.table tr td.disabled,
+    .ui.ui.ui.inverted.table tr.disabled td,
+    .ui.ui.ui.inverted.table tr td.disabled,
     .ui.inverted.table tr.disabled:hover td,
     .ui.inverted.table tr:hover td.disabled {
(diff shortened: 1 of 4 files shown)


exoego (approved):
LGTM

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/415

PR #2552, opened 2022-11-11
fix(nag): overlay nags were always shown

lubber-de:
## Description

overlay nag were always shown, although a nag should be hidden by default.
While we are at it i slightly adjusted the default background for better readability

## Closes
https://github.com/fomantic/Fomantic-UI-Docs/issues/400

### DIFF
--- a/src/themes/default/modules/nag.variables
+++ b/src/themes/default/modules/nag.variables
@@ -11,7 +11,7 @@
 @zIndex: 999;
 @margin: 0;
 
-@background: #777777;
+@background: #909090;
 @opacity: 0.95;
 @minHeight: 0;
 @padding: 0.75em 1em;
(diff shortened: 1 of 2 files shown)


src/themes/default/modules/nag.variables
```diff
@@ -11,7 +11,7 @@
 @zIndex: 999;
 @margin: 0;
 
-@background: #777777;
+@background: #909090;
```
exoego, line 14: Both bright (white) and dark (black) texts has enough contrast on new background??
image

lubber-de: > Do both bright (white) and dark (black) texts have enough contrast on the new bg-color?
> 
> Below is a current screenshot taken from [fomantic/Fomantic-UI-Docs#400](https://github.com/fomantic/Fomantic-UI-Docs/issues/400) and dark text is a bit low-contrast to me. image

That's why i wanted to get this a bit better. It's still not perfect, but the whole default theme would need a groundup refactoring regarding a11y contrast colors (not only nag)

|Old|New|
|-|-|
|image|image|

exoego (approved):

PR #2559, opened 2022-11-18
[dropdown] Fix rounded corners on menu

grandeljay:
## Description

On a filtered dropdown menu the bottom corners are not round. This PR makes all corners of the dropdown menu round.

## Testcase

When opening the filtered dropdown, the bottom two corners of the menu are not rounded.

https://jsfiddle.net/e0zmt376/

## Screenshot (if possible)

image

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1550,9 +1550,6 @@ select.ui.dropdown {
     box-shadow: @floatingMenuBoxShadow !important;
     border-radius: @floatingMenuBorderRadius !important;
   }
-  .ui.floating.dropdown > .menu {
-    border-radius: @floatingMenuBorderRadius !important;
-  }
   .ui:not(.upward).floating.dropdown > .menu {
     margin-top: @floatingMenuDistance;
   }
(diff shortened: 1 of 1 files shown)


lubber-de (changes requested):
This would break submenus in floating dropdowns
See https://jsfiddle.net/lubber/umyns6zw/7/
image

Your use case is very special, as your change is valid only when the menu itself contains a `scrolling menu`. Unfortunately forward lookups in CSS (`:has` )is [not implemented in every browser](https://caniuse.com/?search=%3Ahas) at the moment.

Otherwise the following selector would work for your use case (it's barely visible anyway)
```css
.ui.floating.dropdown > .menu:has(> .scrolling.menu) {
  overflow: hidden;
}
```
See https://jsfiddle.net/lubber/umyns6zw/9/

grandeljay:
Alright and what if we just set the border to 0 for only the top corners in the scrolling menu (like I did now)?

lubber-de:
> Alright and what if we just set the border to 0 for only the top corners in the scrolling menu (like I did now)?

Well, your change now applies to `border-radius` and "0" is the same as "0 0", so there is no visual change in your PR ... 🤔

lubber-de (approved):
LGTM

PR #2561, opened 2022-11-18
fix(dropdown): keep overflow visible for non simple selection

### LINKED ISSUE #2500: Fixed menu dropdown sub-menu not working
# Bug Report
Sub-menu in Dropdown in Menu is not working as expected. The sub-menu does not pop-out but stays inside the "area" and you must scroll to the side or up and down to see the sub-menu item.

## Steps to reproduce
1. Go to https://fomantic-ui.com/examples/fixed.html
2. Go to the menu and open Dropdown
3. Open Sub-Menu

## Expected result
In Fomantic-UI 2.8.8 the sub-menu pop-outs as if it is a normal dropdown with sub-menu

## Actual result
The sub-menu does not pop-out but stays inside the "area" and you must scroll to the side or up and down to see the sub-menu item

## Testcase
Go to https://fomantic-ui.com/examples/fixed.html

## Screenshot (if possible)
fixedmenu

## Version
2.9.0

lubber-de:
## Description

This partly reverts #2219 and only adds the overflow setting in case of a `simple selection` dropdown where a possible scrollbar is only needed. Since 2.9.0 a simple menu item dropdown unfortunately also applied a scrollbar, which is wrong.

This PR also fixes upward menu positioning in case a `floating simple upward dropdown` is used

## Testcase
https://jsfiddle.net/lubber/cz09kaue/12/

## Closes
#2500

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1479,7 +1479,7 @@ select.ui.dropdown {
 
   .ui.simple.active.dropdown > .menu,
   .ui.simple.dropdown:hover > .menu {
-    overflow: auto;
+    overflow: visible;
     width: auto;
     height: auto;
     top: 100%;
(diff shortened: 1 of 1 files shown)


jamessampford:
Not 100% sure if related to simple/this but the overflow on compact goes outside the border

IMG_1442

lubber-de:
> Not 100% sure if related to simple/this but the overflow on compact goes outside the border when somewhat narrow
> 
> IMG_1442

This isn't related to the change of this PR.
The menu just isnt wide enought not fitting content in one line anymore.
A quick fix might be

```css
.ui.compact.menu > .item {
  flex:auto
}
```
but this has to be tested and possibly fixed separately

exoego (approved):

PR #2183, opened 2021-12-18
fix(input): support icon and corner labeled dropdown variant

lubber-de:
## Description

This PR fixes support for `icon input` and `corner labeled input` when used with a `dropdown`

## Testcase
https://jsfiddle.net/lubber/szgydxh9/6/

## Screenshots
|Before|After|
|-|-|
|image|image|

## Closes
#2171 
#2172

PR #2567, opened 2022-11-21
fix(input): adjust padding in labeled dropdown icon inputs

### LINKED ISSUE #2547: [input dropdown] Extraneous padding in dropdown when input has left icon
# Bug Report
Since 2.9.0, there is extraneous padding to the left of the dropdown label.
This seems to have been introduced in PR #2183

## Steps to reproduce
Use the right action left icon example from the documentation:
```
<div class="ui right action left icon input">
  <i class="[search icon](https://fomantic-ui.com/elements/icon.html)"></i>
  <input type="text" placeholder="Search">
  <div class="[ui basic floating dropdown button](https://fomantic-ui.com/elements/button.html)">
    <div class="text">This Page</div>
    <i class="[dropdown icon](https://fomantic-ui.com/elements/icon.html)"></i>
    <div class="menu">
      <div class="item">This Organization</div>
      <div class="item">Entire Site</div>
    </div>
  </div>
</div>
```

## Expected result
Padding similar to screenshot from the [2.8 release using archive.org](https://web.archive.org/web/20211220015310/https://fomantic-ui.com/elements/input.html#action).
Screen Shot 2022-11-08 at 2 07 57 PM

## Actual result
Extraneous padding, see screenshot.
Screen Shot 2022-11-08 at 2 08 07 PM

## Screenshot (if possible)
Screen Shot 2022-11-08 at 1 52 58 PM

## Version
2.9.0

lubber-de:
## Description
in some situations the padding inside labeled icon inputs is not correct
This change uses `@supports selector` which in turn needs a fix in autoprefixer. That's why i updated the autoprefixer dependency as well to avoid a build error.

## Testcase
The testcase is the same as from #2183 to make sure those examples stay intact
What matters is the first labeled dropdown having left/right icons leaving a gap between the dropdown icon and the input
https://jsfiddle.net/lubber/37xuoygt/5/

### Before
image
### After
image

## Closes
#2547

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "del": "^6.1.1",
     "extend": "^3.0.2",
     "gulp": "^4.0.0",
-    "gulp-autoprefixer": "^6.0.0",
+    "gulp-autoprefixer": "^8.0.0",
     "gulp-chmod": "^2.0.0",
     "gulp-clean-css": "^4.3.0",
     "gulp-clone": "^2.0.1",
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -428,7 +428,8 @@
   }
 
   /* Left Labeled */
-  .ui[class*="left icon"].input > .ui.dropdown,
+  .ui[class*="left icon"].input > .ui.dropdown:first-child,
+  .ui[class*="left icon"].input > i.icon + .ui.dropdown,
   .ui[class*="left corner labeled"].input > .ui.dropdown,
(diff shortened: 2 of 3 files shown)


exoego (approved):
LGTM
Hmm I hope we have screenshot-diffs on CI so that we ensure nothing breaks

lubber-de:
> LGTM Hmm I hope we have screenshot-diffs on CI so that we ensure nothing breaks

Yeah, we definately should somehow implement that

PR #2568, opened 2022-11-21
feat(statistic): fluid variant

### LINKED ISSUE #2536: [statistic] fluid statistic
# Feature Request
At present, if one just wants to show one statistic and have it centred taking up the available width (as opposed to it just appearing) one needs to wrap it with `ui one statistics`

I propose that optionally the same can be done by adding the `fluid` class

## Example

So instead of:
```
<div class="ui segment">
  <div class="ui one statistics">
    <div class="statistic">
      <div class="value">
        5,550
      </div>
      <div class="label">
        Downloads
      </div>
    </div>
  </div>
</div>
```

One could do:
```
<div class="ui segment">
  <div class="ui fluid statistic">
    <div class="value">
      5,550
    </div>
    <div class="label">
      Downloads
    </div>
  </div>
</div>
```

lubber-de:
## Description

Added a `fluid statistic` variant which is a quicker and Dom node saving method than using a `one statistic` group to occupy the whole width.

## Testcase
https://jsfiddle.net/lubber/p9nj5gdy/1/

## Closes
#2536

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "del": "^6.1.1",
     "extend": "^3.0.2",
     "gulp": "^4.0.0",
-    "gulp-autoprefixer": "^6.0.0",
+    "gulp-autoprefixer": "^8.0.0",
     "gulp-chmod": "^2.0.0",
     "gulp-clean-css": "^4.3.0",
     "gulp-clone": "^2.0.1",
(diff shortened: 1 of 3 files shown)


exoego (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/418

PR #2574, opened 2022-11-26
feat(chore): prevent overscrolling

### LINKED ISSUE #2573: Possible undesired pull-to-refresh look on iPadOS when using scrolling
# Bug Report
I spotted that by trying the JSFiddle within #2571 on iPad the whole table pulls down (like pull down to refresh). This similarly also happens when using `.scrolling` on a `<table>` (with the `<tbody>` pulling down to refresh). This also happens with a dropdown menu that uses scrolling as well.

## Steps to reproduce
Either table scenario can only be spotted if the table appears without needing to scroll down the page, or if scrolling slightly down the table then back up too far. The same happens on drop-down menus too.

## Expected result
In a way don’t expect it to pull down, unless of course the data will update, but then it may be ‘useful’ to know that one is at the top. Not sure if one can listen to this with JS, or would be useful to implement as a feature if it were to do some kind of refresh . . .

## Possible Solution
For the first (JSFiddle) scenario, the below seems to work in preventing the ‘pull down to refresh’ look:

```
.scrolling {
  overscroll-behavior: none;
}
```

The second scenario seems to like the below:

```
.scrolling tbody {
  overscroll-behavior: none;
}
```

lubber-de:
## Description
On mobile devices when scrolling inside a dropdown menu, scrolling table, segment, container, etc. reaching the top of the scrolling area, the browsers will reload the page "pull to refresh".

This PR prevents such behavior on selected components, where an accidental refresh shouldnt be wanted.

## Testcase
- Run on a mobile device
- touch scroll inside the table 

#### :( Reloads the page
https://fomantic-ui.com/jsfiddle/#!lubber/fas7wbey/20/

#### Ignores it :)
https://fomantic-ui.com/jsfiddle/#!lubber/fas7wbey/19/

## Closes
#2573

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -498,6 +498,7 @@
   .ui.overflowing.table {
     display: block;
     overflow: auto;
+    overscroll-behavior: @overscrollBehavior;
   }
   @media only screen and (min-width: @tabletBreakpoint) {
     & when (@variationTableOverflowingShort) {
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -399,6 +399,7 @@
   .ui.scrolling.table > tbody {
     display: block;
     overflow-y: scroll;
+    overscroll-behavior: @overscrollBehavior;
     scrollbar-width: thin;  /* Firefox */
     & > tr {
       display: table;
--- a/src/definitions/elements/container.less
+++ b/src/definitions/elements/container.less
@@ -189,6 +189,7 @@
   ---------------*/
   .ui.scrolling.container {
     overflow: auto;
+    overscroll-behavior: @overscrollBehavior;
   }
   @media only screen and (max-width : @largestMobileScreen) {
     & when (@variationContainerScrollingShort) {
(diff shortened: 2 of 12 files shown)


exoego:
Hmm, it seems that behavior on both example page is identical for me...?

- iPhone
- iOS 16.1.1

lubber-de:
Try Chrome, This feature needs Safari 16 according to https://caniuse.com/?search=overscroll-behavior

exoego:
Safari is 16.1

lubber-de:
mmh, using Browserstack it works on iphone 14 using safari 16. 
iphone16

lubber-de:
I now got my hands on a real iphone 14 and indeed the bound effect inside the container still exists (however, it does not reload the page). This seems to be a webkit bug, because as by specification the css property is correct but Safari 16 is quite new.

As it works/fixes on chrome/android, i think we could merge it anyway. But i am also fine to set it on hold for a while and watch Safari devlopment

exoego (approved):
Ok then 👍

PR #2575, opened 2022-11-27
fix(modal): do autofocus on content change

### LINKED ISSUE #2490: [modal] tab key is not trapped on modal open
# Bug Report

## Steps to reproduce

1. have a page with several links (`<a...` tags) and a button which can open a modal (with blured background) on click with 2 input fields
2. click the button to open the modal
3. press `tab` key on keyboard and notice the links can/are focused (browser usually shows the focused link URL in the bottom left browser window corner)
4. notice, once you cycle thru all the links and move the tab/focus into the modal input, the tab is then trapped correctly (cycles in modal input fields only), given this fact/functionality, I belive the tab key must be trapped immediatelly on open

env: FIrefox, modal `observeChanges` enabled

## Expected result

tab key is trapped immediatelly once the modal is open

## Actual result

tab key is NOT trapped until it hit some focusable element in the modal for the first time

## Version

2.9.0

lubber-de:
## Description
The autofocus feature does not work when the content of a modal or flyout is loaded/changed after the modal is shown (and contains an input after the change)

## Testcase
The content of the modal changes after 3 seconds (and then has got an input

### Broken
- new input field is not focussed
https://jsfiddle.net/lubber/Lzkw5dyq/4/

### Fixed
- new input field is focussed
https://jsfiddle.net/lubber/Lzkw5dyq/18/

## Closes
#2490


mvorisek:
maybe it can be part of `refreshInputs`, but LGTM

exoego (approved):

exoego (approved):

mvorisek:
@lubber-de please see my updated comment, I do not think the current fix is good

there are 2 separate problems:

1. tab key must be trapped in modal even if there is no input (on show immediately, once only) - #2490
2. autoficus on observed change - something that this PR tries to solve - see my previous comment

lubber-de:
> 1. tab key must be trapped in modal even if there is no input (on show immediately, once only)

I'll do that if `autoFocus` is true (which it is by default already) otherwise it's a breaking change if we also focus trap when eombody explicitely set autoFocus: false (by whatever reason....)

> 2. autofocus on observed change - something that this PR tries to solve - see my previous comment
Yes, understood, i am on it

lubber-de:
I think i covered all mentioned situations now.
@mvorisek If you have time this weekend, i would appreciate, if you can test this

mvorisek:
I upgraded atk4/ui to Fomantic-UI 2.9.1 and the autofocus is still not fully working, steps to reproduce:

1. open https://dev.agiletoolkit.org/demos/interactive/modal.php
2. click `Show` (first) button
3. tab key seems to be trapped correctly
4. hide the modal by clicking on the dimmer
5. repeat step 2 - open modal again
6. tab key is not trapped, pressing tab key many times cycle thru the links below the modal

lubber-de:
> I upgraded atk4/ui to Fomantic-UI 2.9.1 and the autofocus is still not fully working, steps to reproduce:
> 
> 1. open [dev.agiletoolkit.org/demos/interactive/modal.php](https://dev.agiletoolkit.org/demos/interactive/modal.php)
> 2. click `Show` (first) button
> 3. tab key seems to be trapped correctly
> 4. hide the modal by clicking on the dimmer
> 5. repeat step 2 - open modal again
> 6. tab key is not trapped, pressing tab key many times cycle thru the links below the modal

Should finally be fixed by https://github.com/fomantic/Fomantic-UI/pull/2674

PR #2580, opened 2022-12-01
fix(dropdown): reuse select tabindex

lubber-de:
## Description
Reuse existing tabindex from select tag when transforming into dropdown. Even worse, an existing tabindex for a select would have inpact on tabbing even though the original select is hidden when transformed into a dropdown.

@inspector71 This will fix your issue

### Before
image

### After
image

### Before
https://jsfiddle.net/lubber/qkyou3c1/2/
### After
https://jsfiddle.net/lubber/qkyou3c1/3/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/7100


exoego (approved):
LGTM

PR #2582, opened 2022-12-02
fix(api): explicite url setting should have priority over data url

### LINKED ISSUE #2581: [api] Explicitly passed URL should honored over $el.data.url
# Bug Report

It seems like a bug, explicitly passed URL to `$el.api()` should be always preffered.

I have discovered this issue in atk4/ui where we set `$modal.url` for our purposes - https://github.com/atk4/ui/blob/4.0.0/js/src/plugins/create-modal.plugin.js#L18 - but then query API using the same element like `$modal.api(...)` with different URL for another purpose.

(for me - https://dev.agiletoolkit.org/demos/interactive/modal.php `Multi Step Modal` is broken for this reason)

## Steps to reproduce

see https://jsfiddle.net/xj02cbog/

## Expected result

Explicitly passed URL is preffered over URL from  jQuery element data.

## Actual result

URL from jQuery element data is preffered.

## Version
2.9.0

lubber-de:
## Description

A given url as api setting should always have higher priority than a previously existing data value of the same element

## Testcase
Watch console

### Broken
url from data is used
https://jsfiddle.net/xj02cbog/

### Fixed
url from api is used
https://jsfiddle.net/lubber/t5jokpqd/2/

## Closes
#2581


src/definitions/behaviors/api.js
```diff
@@ -868,7 +868,7 @@ $.api = $.fn.api = function(parameters) {
           },
           templatedURL: function(action) {
             action = action || $module.data(metadata.action) || settings.action || false;
-            url    = $module.data(metadata.url) || settings.url || false;
+            url    = settings.url || $module.data(metadata.url) || false;
```
mvorisek, line 871: what about the config one line above, shouldn't it be fixed too (and maybe there are also more places)

lubber-de: 👍🏼 You've got a point here (i was so concentrated on your issue with the url 😆 
Will check the reamining code and ajust accordingly. Thanks for the hint

lubber-de: Done by [1b75224](https://github.com/fomantic/Fomantic-UI/pull/2582/commits/1b752241e4d26c5123da3fcf711d1fa7a09edc91) Seems to be the only other similiar code in the whole api

mvorisek: `(?<!meta)data.*setti` regex returns 27 matches

- I am not sure about https://github.com/fomantic/Fomantic-UI/blob/7071e5977705394994b87df42855435f4b357920/src/definitions/behaviors/api.js#L205
- in calendar and popup module there is quite simillar pattern like in API

lubber-de: > 

In API only? When just searching for module.data and also same settings attribute (in API.js) i cannot find any similar statement 

> - I am not sure about https://github.com/fomantic/Fomantic-UI/blob/7071e5977705394994b87df42855435f4b357920/src/definitions/behaviors/api.js#L205

This is correct, as it was implemented some versions ago to respect already sent data

> - in calendar and popup module there is quite simillar pattern like in API

Well, your issue was initially about API..;)... i take a look at the other modules as well

exoego (approved):

src/definitions/behaviors/api.js
lubber-de: > `(?<!meta)data.*setti` regex returns 27 matches
> * in calendar and popup module there is quite simillar pattern like in API

i checked the source given your search regex. I could not identify the same issues in the other modules because the found targets were either 
- get functions where is is intended to get a possibly existing previously created metadata before
- currently not getting any data at all for metadate as the necessary code for setting the date is missing (so it always fetches the setting)
- having  a non falsy default setting (like a string) which would mean we _always_ would provide the setting and never catch an existing .data setting
- did not apply at all 

That said, i am merging this PR as is. If you think we are still have similiar issues in other modules, please prepare a jsfiddle testcase 🙂

mvorisek: > If you think we are still have similiar issues in other modules, please prepare a jsfiddle testcase 🙂

No, I am not aware of the same problem anywhere else, I am just used to check every similar pattern (often using regex) everywhere else.

PR #2587, opened 2022-12-05
feat(chore): use modern globalThis

mvorisek:
`globalThis` is quite new feature, but all supported browsers except IE11 should support it - https://caniuse.com/?search=globalThis

I do not consider this a problem, as IE11 is kept supported when `window` is defined which normally is, only there is no longer a fallback when undefined/shadowed.

For documentation purposes and easier revert/git bisect I propose this change as a separate PR.

### DIFF
--- a/src/definitions/globals/site.js
+++ b/src/definitions/globals/site.js
@@ -9,9 +9,15 @@
  */
 
 (function ($, window, document, undefined) {
+    'use strict';
+
     function isFunction(obj) {
         return typeof obj === 'function' && typeof obj.nodeType !== 'number';
     }
+
+    window = (window !== undefined && window.Math == Math)
+        ? window
+        : globalThis;
 
     $.site = $.fn.site = function (parameters) {
         var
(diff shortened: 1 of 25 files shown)


src/definitions/globals/site.js
```diff
@@ -9,10 +9,16 @@
  */
 
 (function ($, window, document, undefined) {
+    'use strict';
```
mvorisek, line 12: no strict nor window fallback was here, is `site.js` somehow special?

lubber-de:
> I do not consider this a problem, as IE11 is kept supported when `window` is defined which normally is, only there is no longer a fallback when undefined/shadowed.

I basically agree. However, IE11 supports web workers where this change won't work anymore.
But, i doubt any FUI module can be successfully used inside webworkers anyway, because of not being able to access the DOM, which every module needs.

lubber-de:
🤔 globalThis does not work in Safari < 12.1 and/or Android 4.4.4.
As the fallback is the same as for IE11 (window object) this should be safe for 2.9.x

PR #2584, opened 2022-12-04
Replace "var" declarations with "let"

mvorisek:
`let` vs `var` declaration is safer, see https://stackoverflow.com/questions/762011/what-is-the-difference-between-let-and-var.

No functional change.

PR #2588, opened 2022-12-05
feat(lint): fix global-require eslint rule

mvorisek:


mvorisek:
PR is complete and it should not contain any breaking change, so this can go into the develop branch without having to wait for branch cut for 2.10.x.

Node 12 support is kept, Node 14+ is however required for eslint as described in the PR description.

src/definitions/modules/sidebar.js
```diff
@@ -309,7 +309,7 @@
                     module.verbose('Forcing repaint event');
                     element.style.display = 'none';
                     var ignored = element.offsetHeight;
-                    element.scrollTop = element.scrollTop;
+                    element.scrollTop = element.scrollTop; // eslint-disable-line no-self-assign
```
mvorisek, line 312: is this line really useful? can we unify the (forced) repaint across modules?

src/definitions/modules/calendar.js
```diff
@@ -758,13 +777,13 @@
                         var ms1d = 864e5, // milliseconds in a day
                             ms7d = 7 * ms1d; // milliseconds in a week
 
-                        return function () { // return a closure so constants get calculated only once
+                        return (function () { // return a closure so constants get calculated only once
```
mvorisek, line 780: the closure is invoked immediately and not returned, is it really useful?

src/definitions/modules/dropdown.js
```diff
@@ -2391,7 +2397,7 @@
                         $nextSelectedItem
                             .addClass(className.selected)
                         ;
-                        if (settings.selectOnKeydown && module.is.single() && !$nextItem.hasClass(className.actionable)) {
+                        if (settings.selectOnKeydown && module.is.single() && !$nextSelectedItem.hasClass(className.actionable)) {
```
mvorisek, line 2400: did I fixed the name to an intended variable?

src/definitions/behaviors/api.js
```diff
@@ -157,11 +157,6 @@
                 },
                 write: {
                     cachedResponse: function (url, response) {
-                        if (response && response === '') {
```
mvorisek, line 160: never true condition, not sure what to do

if response can be a plain text, caching empty response might be expected, as the condition is never true, no BC break

lubber-de (changes requested):
I havent reviewed all, but those 2 changes i wanted to report beforehand

.github/workflows/release.yml
```diff
@@ -16,7 +16,6 @@ jobs:
         ref: master
     - uses: actions/setup-node@v3
       with:
-        node-version: 14
```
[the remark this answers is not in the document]
mvorisek: this is a CI script that use `ubuntu-latest` Docker image, so this way, it will use the latest stable Node which should be always supported by Fomantic-UI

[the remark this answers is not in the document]
lubber-de: The latest stable node does not necessarily mean that the nightly script will still work in any future version! If this isnt tested for node 22 (for example) and the docker images gets updated and, by whatever (unlikely) incompatibility the scripts does not work (or throws any unexpected error) it will break
Yes, unlikely it breaks, but who knows. As said, even setup-node suggests to state an explicit version 

Remember, we had the situation where Node 15/16 was previouly not working with fomantic anymore because of an incompatibility with inquirere not supporting yarn with interactive scripts . We had this before, that's why :wink:

Yes, Fomantic _should_ support the latest node version, but only until tested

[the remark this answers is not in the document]
mvorisek: I prefer no fixed numbers anywhere so I added `latest` to the CI.

- the release script (outside the push to registry) is the same as our CI test
- the nighly script is the same as well plus one file to compute the next version

so release script should never fail and nighly script can fail, but I would say nighly failure is good, as we can detect the problems :)

src/definitions/modules/sidebar.js
lubber-de: > is this line really useful?

From the git history, is is not clear, why this line was necessary (clear commit message ftw). I strongly believe it was needed by IE8-11 (?) back then.

>  can we unify the (forced) repaint across modules?

Otherwise, there are a lot of functions i would remove from all modules and put into a separate "core.js" or similar  (like the debug functions). maybe enhance the existing site.js. 

Perhaps repaint is also a candidate.

But this would mean each module is not working on its own anymore (at least when following the ES5 syntax as it it now).
I am already thinking about such way to reduce the code overhead. Even if that means people have to make sure to include core.js...but this is another story

src/definitions/modules/calendar.js
lubber-de: > the closure is not returned (invoked immediately)

I think it is returned (wouldn't work totherwise)
```javascript
return function () { // return a closure so constants get calculated only once
...
}();
```
>  is it really useful?

tbh, i just took that calculation function from the credited website back then 😆

src/definitions/modules/dropdown.js
lubber-de: You actually found a bug and fixed it. 👍🏼 Deserves a separate PR, please.

src/definitions/behaviors/api.js
lubber-de: Agreed, lets remove this

tasks/config/npm/gulpfile.js
```diff
@@ -2,33 +2,33 @@
  *           Set-up
  *******************************/
 
-var
+let
     gulp   = require('gulp'),
 
     // read user config to know what task to load
-    config = require('./tasks/config/user')
+    config = require('./tasks/config/user') // eslint-disable-line import/extensions
```
mvorisek, line 9: this file is copied in two places using `gulp.src(source.userGulpFile)` - do we really need that instead of the `/gulpfile.js`

please notice `tasks/config/npm/gulpfile.js` vs. `/gulpfile.js` which are almost identical

src/definitions/modules/calendar.js
mvorisek: `let x = (function () {})( return 5; );` (function evaluated immediatelly) is basically a `let x = 5;`

maybe it has something to do with https://eslint.org/docs/latest/rules/wrap-iife

still, I think "function evaluated immediatelly" is not needed here

src/definitions/modules/dropdown.js
mvorisek: fixed in #2591

src/definitions/behaviors/api.js
mvorisek: fixed in #2591

tasks/admin/components/init.js
```diff
@@ -38,6 +37,8 @@ const
 ;
 
 module.exports = function (callback) {
+    const github = require('../../config/admin/github'); // eslint-disable-line global-require
```
lubber-de, line 40: Why was this necessary? If you leave the github const where it was before the eslint rule is already fulfilled, isn't it?
Now, you moved it here and have to disable the rule... i don't get the reason 🤔

tasks/admin/components/update.js
```diff
@@ -36,6 +35,8 @@ const
 ;
 
 module.exports = function (callback) {
+    const github = require('../../config/admin/github'); // eslint-disable-line global-require
```
lubber-de, line 38: same here. If you havent moved it, it was alreay a global const

tasks/admin/distributions/init.js
```diff
@@ -38,6 +37,8 @@ const
 ;
 
 module.exports = function (callback) {
+    const github = require('../../config/admin/github'); // eslint-disable-line global-require
```
lubber-de, line 40: and again here... 🙂

tasks/admin/distributions/update.js
```diff
@@ -36,6 +35,8 @@ const
 ;
 
 module.exports = function (callback) {
+    const github = require('../../config/admin/github'); // eslint-disable-line global-require
```
lubber-de, line 38: and again...

tasks/collections/build.js
```diff
@@ -2,17 +2,15 @@
...
-        watch       = require('../watch'),
+const watch = require('../watch');
 
-        build       = require('../build'),
-        buildJS     = require('../build/javascript'),
-        buildCSS    = require('../build/css'),
-        buildAssets = require('../build/assets')
-    ;
+// build sub-tasks
+const build = require('../build');
```
lubber-de, line 8: Why did you put each const into a separate `const` statement instead of a comma separted list? in admin.js above, you did it as a comma separated list (most probably because of copy/paste form the origin)
I like the comma separated list instead (which seems to be done all the other times in the scripts already.
Let's be consistent and also have a comma separated list here 🙂

tasks/admin/components/init.js
mvorisek: Try moving this require up, local build will fail, as the require tries to connect to GitHub. It is good practise in libs to only define things, but no execute them immediately. Can you rewrite the `config/admin/github.js` to a singleton pattern?

tasks/collections/build.js
mvorisek: I would say single declaration per variable is better for diff, readability, ... But definitely not to be changed here nor now, probably to be changed in #2584. Here fixed.

tasks/admin/components/init.js
lubber-de: Ok, understood. The whole github API integration seems to be broken anyway and using deprecated functions, so i approve your changes as is for now as it needs a refactoring separately and is not a task for local building but for repo-admin purposes only

lubber-de (approved):
LGTM

PR #2589, opened 2022-12-07
feat(lint): remove unneeded parentheses around expressions

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on fixing `no-extra-parens` rule with a help or a custom rule output filter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "yamljs": "^0.3.0"
   },
   "devDependencies": {
+    "@internal/eslint-plugin": "file:.eslint",
     "all-contributors-cli": "^6.7.0",
     "auto-changelog": "^2.4.0",
     "eslint": "^8.29.0",
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -362,11 +362,11 @@
                                         variable = templatedString.indexOf('$') !== -1
                                             ? templatedString.slice(2, -1)
                                             : templatedString.slice(1, -1),
-                                        value   = ($.isPlainObject(urlData) && urlData[variable] !== undefined)
+                                        value   = $.isPlainObject(urlData) && urlData[variable] !== undefined
                                             ? urlData[variable]
-                                            : (($module.data(variable) !== undefined)
+                                            : ($module.data(variable) !== undefined
                                                 ? $module.data(variable)
-                                                : (($context.data(variable) !== undefined) // eslint-disable-line unicorn/no-nested-ternary
+                                                : ($context.data(variable) !== undefined // eslint-disable-line unicorn/no-nested-ternary
                                                     ? $context.data(variable)
                                                     : urlData[variable]))
                                     ;
@@ -393,11 +393,11 @@
                                         variable = templatedString.indexOf('$') !== -1
                                             ? templatedString.slice(3, -1)
                                             : templatedString.slice(2, -1),
-                                        value   = ($.isPlainObject(urlData) && urlData[variable] !== undefined)
+                                        value   = $.isPlainObject(urlData) && urlData[variable] !== undefined
                                             ? urlData[variable]
-                                            : (($module.data(variable) !== undefined)
+                                            : ($module.data(variable) !== undefined
                                                 ? $module.data(variable)
-                                                : (($context.data(variable) !== undefined) // eslint-disable-line unicorn/no-nested-ternary
+                                                : ($context.data(variable) !== undefined // eslint-disable-line unicorn/no-nested-ternary
                                                     ? $context.data(variable)
                                                     : urlData[variable]))
                                     ;
@@ -748,15 +748,15 @@
                     },
                     responseFromXHR: function (xhr) {
                         return $.isPlainObject(xhr)
-                            ? ((module.is.expectingJSON())
+                            ? (module.is.expectingJSON()
                                 ? module.decode.json(xhr.responseText)
                                 : xhr.responseText)
                             : false;
                     },
                     errorFromRequest: function (response, status, httpMessage) {
-                        return ($.isPlainObject(response) && response.error !== undefined)
+                        return $.isPlainObject(response) && response.error !== undefined
                             ? response.error // use json error message
-                            : ((settings.error[status] !== undefined) // use server error message
+                            : (settings.error[status] !== undefined // use server error message
                                 ? settings.error[status]
                                 : httpMessage);
                     },
@@ -836,9 +836,9 @@
                         }
                         if (settings.on == 'auto') {
                             if ($module.is('input')) {
-                                return (element.oninput !== undefined)
+                                return element.oninput !== undefined
                                     ? 'input'
-                                    : ((element.onpropertychange !== undefined)
+                                    : (element.onpropertychange !== undefined
                                         ? 'propertychange'
                                         : 'keyup');
                             }
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -136,7 +136,10 @@
                 },
 
                 attachEvents: function (selector, action) {
-                    action = action || 'submit';
+                    if (!action) {
+                        action = 'submit';
+                    }
+
                     $(selector).on('click' + eventNamespace, function (event) {
                         module[action]();
                         event.preventDefault();
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -492,9 +492,9 @@
                         return module.get.inputEvent();
                     },
                     inputEvent: function () {
-                        return (document.createElement('input').oninput !== undefined)
+                        return document.createElement('input').oninput !== undefined
                             ? 'input'
-                            : ((document.createElement('input').onpropertychange !== undefined)
+                            : (document.createElement('input').onpropertychange !== undefined
                                 ? 'propertychange'
                                 : 'keyup');
                     },
@@ -539,13 +539,11 @@
                         if (ancillary && ['integer', 'decimal', 'number'].indexOf(ruleName) >= 0 && ancillary.indexOf('..') >= 0) {
                             parts = ancillary.split('..', 2);
                             if (!rule.prompt) {
-                                suffixPrompt = (
-                                    parts[0] === ''
-                                        ? settings.prompt.maxValue.replace(/{ruleValue}/g, '{max}')
-                                        : (parts[1] === ''
-                                            ? settings.prompt.minValue.replace(/{ruleValue}/g, '{min}')
-                                            : settings.prompt.range)
-                                );
+                                suffixPrompt = parts[0] === ''
+                                    ? settings.prompt.maxValue.replace(/{ruleValue}/g, '{max}')
+                                    : (parts[1] === ''
+                                        ? settings.prompt.minValue.replace(/{ruleValue}/g, '{min}')
+                                        : settings.prompt.range);
                                 prompt += suffixPrompt.replace(/{name}/g, ' ' + settings.text.and);
                             }
                             prompt = prompt.replace(/{min}/g, parts[0]);
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -136,10 +136,7 @@
                 },
 
                 attachEvents: function (selector, action) {
-                    if (!action) {
-                        action = 'submit';
-                    }
-
+                    action = action || 'submit';
                     $(selector).on('click' + eventNamespace, function (event) {
                         module[action]();
                         event.preventDefault();
--- a/src/definitions/modules/flyout.js
+++ b/src/definitions/modules/flyout.js
@@ -1015,8 +1015,8 @@
                     ie: function () {
                         if (module.cache.isIE === undefined) {
                             var
-                                isIE11 = (!(window.ActiveXObject) && 'ActiveXObject' in window),
-                                isIE = ('ActiveXObject' in window)
+                                isIE11 = !window.ActiveXObject && 'ActiveXObject' in window,
+                                isIE = 'ActiveXObject' in window
                             ;
                             module.cache.isIE = isIE11 || isIE;
                         }
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -941,8 +941,8 @@
                     ie: function () {
                         if (module.cache.isIE === undefined) {
                             var
-                                isIE11 = (!(window.ActiveXObject) && 'ActiveXObject' in window),
-                                isIE = ('ActiveXObject' in window)
+                                isIE11 = !window.ActiveXObject && 'ActiveXObject' in window,
+                                isIE = 'ActiveXObject' in window
                             ;
                             module.cache.isIE = isIE11 || isIE;
                         }
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -260,16 +260,13 @@
 
                 get: {
                     text: function (templateText, index) {
-                        if (!index) {
-                            index = 0;
-                        }
-
                         var
-                            value   = module.get.value(index),
+                            index_  = index || 0,
+                            value   = module.get.value(index_),
                             total   = module.get.total(),
                             percent = animating
-                                ? module.get.displayPercent(index)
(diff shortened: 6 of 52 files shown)


mvorisek:
I know how hard is to review large PRs, so I basically redone #2588 (without some fixes to be fixed later). This PR should be reviewed commit by commit, (large) commits starting with `strictly autofix` are done purely by eslint.

All changes are compatible with Node 12 and IE11. No functional change.

.github/workflows/nightly.yml
```diff
@@ -18,7 +18,6 @@ jobs:
         ref: develop
     - uses: actions/setup-node@v3
       with:
-        node-version: 16
```
lubber-de, line 21: I still dont want this to be removed. It is not recommended  setup-node and the nightly script has to work all the time. 
if it breaks, i dont want to stress us to investigate.

.github/workflows/release.yml
```diff
@@ -16,7 +16,6 @@ jobs:
         ref: master
     - uses: actions/setup-node@v3
       with:
-        node-version: 14
```
lubber-de, line 19: Same here. The release action has to work, no matter what. So a node dedicated node version has to be set.
However, i am fine if you upgrade to node 16 here

.github/workflows/nightly.yml
mvorisek: I restored the fixed version to the latest LTS - https://endoflife.date/nodejs

src/definitions/behaviors/state.js
```diff
@@ -8,14 +8,14 @@
  *
  */
 
-(function ($, window, document, undefined) {
```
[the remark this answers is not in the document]
mvorisek: 👍 I did not know, I have reverted this and added eslint ignore comments in the files directly, so the rule is still active otherwise.

src/definitions/behaviors/form.js
```diff
@@ -136,7 +137,10 @@
                 },
 
                 attachEvents: function (selector, action) {
-                    action = action || 'submit';
```
lubber-de, line 139: I actually dont like this change. It bloats the code, although it's clear what it does.
It is also inconsistent as many other places in the code still have " xxx || yyy"

mvorisek: see https://github.com/fomantic/Fomantic-UI/commit/3b79c1cee555bf22ee10cf4be2e4978a97de73d6

this is to fix `unicorn/prefer-default-parameters` rule, only a few lines were required to be updated

IMHO the `if` is more readable than reassigning the same value, but that just my opinion, I did not fixed the code in general nor plan do to so

lubber-de: Wouldn't it be better to remove that rule from eslint until we finally dropped IE11 support? The desired fix according to that rule would be
```
attachEvents: function (selector, action = 'submit') {
```

src/definitions/modules/toast.js
```diff
@@ -8,33 +8,34 @@
...
     function isFunction(obj) {
         return typeof obj === 'function' && typeof obj.nodeType !== 'number';
     }
 
-    window = (typeof window != 'undefined' && window.Math == Math)
+    window = window !== undefined && window.Math === Math
         ? window
         : globalThis;
 
-    $.toast = $.fn.toast = function (parameters) {
```
lubber-de, line 22: Dont remove the short cut call without an attached element. It was explcitely added to 2.9.0

src/definitions/modules/flyout.js
```diff
@@ -8,18 +8,19 @@
...
     function isFunction(obj) {
         return typeof obj === 'function' && typeof obj.nodeType !== 'number';
     }
 
-    window = (typeof window != 'undefined' && window.Math == Math)
+    window = window !== undefined && window.Math === Math
         ? window
         : globalThis;
 
-    $.flyout = $.fn.flyout = function (parameters) {
```
lubber-de, line 22: Dont remove the short cut call without an attached element. It was explcitely added to 2.9.0

src/definitions/behaviors/api.js
```diff
@@ -19,30 +20,30 @@
         return typeof obj === 'function' && typeof obj.nodeType !== 'number';
     }
 
-    window = (typeof window != 'undefined' && window.Math == Math)
+    window = window !== undefined && window.Math === Math
         ? window
         : globalThis;
 
-    $.api = $.fn.api = function (parameters) {
```
lubber-de, line 26: Dont remove the short cut call without an attached element. It is referenced later at settings, used inside tab.js and would be a breaking change

src/definitions/modules/modal.js
```diff
@@ -8,18 +8,19 @@
...
     function isFunction(obj) {
         return typeof obj === 'function' && typeof obj.nodeType !== 'number';
     }
 
-    window = (typeof window != 'undefined' && window.Math == Math)
+    window = window !== undefined && window.Math === Math
         ? window
         : globalThis;
 
-    $.modal = $.fn.modal = function (parameters) {
```
lubber-de, line 22: Dont remove the short cut call without an attached element. It was explcitely added to 2.9.0

src/definitions/modules/flyout.js
```diff
@@ -1303,10 +1310,11 @@
             }
         });
 
-        return (returnedValue !== undefined)
+        return returnedValue !== undefined
             ? returnedValue
             : this;
     };
+    $.flyout = $.fn.flyout;
```
lubber-de, line 1317: Ok, i see this is instead added here... what's the benefit?

src/definitions/modules/toast.js
```diff
@@ -769,10 +789,11 @@
             }
         });
 
-        return (returnedValue !== undefined)
+        return returnedValue !== undefined
             ? returnedValue
             : this;
     };
+    $.toast = $.fn.toast;
```
lubber-de, line 796: Ok, i see this is instead added here... what's the benefit?

src/definitions/behaviors/api.js
```diff
@@ -1062,10 +1063,11 @@
             }
         });
 
-        return (returnedValue !== undefined)
+        return returnedValue !== undefined
             ? returnedValue
             : this;
     };
+    $.api = $.fn.api;
```
lubber-de, line 1070: Ok, i see this is instead added here... what's the benefit?

src/definitions/modules/flyout.js
mvorisek: yes this change is intended, see https://github.com/fomantic/Fomantic-UI/pull/2589/commits/e4333cfff8c6fd1cf856a0b5b83658c3327785d9, it is a recommended practice to never assign more than one variable in one line to improve readability

src/definitions/behaviors/form.js
mvorisek: In this PR, I focus on no functional change. The fix with default argument is changing the "method prototype", please confirm we can use it safely in all https://github.com/fomantic/Fomantic-UI/commit/3b79c1cee555bf22ee10cf4be2e4978a97de73d6 changes, if yes, I will change it, if not, I prefer to keep the changes as they are.

src/definitions/behaviors/state.js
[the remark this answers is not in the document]
mvorisek: 👍 I was not aware of the minimizer/uglify impact, ~I have reverted it.~

UPDATE: I did more research and I prefer cleaner code.

File size of `semantic.min.js` /w `undefined` defined (as before) was/is 399 699 bytes, after the change it is 402 684 bytes.

IMHO the total file size difference of about 0.7% is minimal (and the transfer size can be even the same if served with transport compression).

There should be no BC break (as `undefined` is not redefined anywhere).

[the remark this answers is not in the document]
lubber-de: Well we already had complaints from people to reduce the min files as much as possible and now it gets increased...just saying😉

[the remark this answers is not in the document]
mvorisek: Of course we should aim for minimal size, but 0.7% percent is not significant to worry about.

mvorisek:
PR is complete

package.json
```diff
@@ -73,9 +73,12 @@
     "yamljs": "^0.3.0"
   },
   "devDependencies": {
+    "@internal/eslint-plugin": "link:.eslint",
     "all-contributors-cli": "^6.7.0",
     "auto-changelog": "^2.4.0",
-    "eslint": "^8.28.0",
+    "eslint": "^8.29.0",
+    "eslint-plugin-import": "^2.26.0",
+    "eslint-plugin-jest": "26.9.0",
```
lubber-de, line 81: yarn moans about typescript being an unmet peer dependency.
```
warning "eslint-plugin-jest > @typescript-eslint/utils > @typescript-eslint/typescript-estree > tsutils@3.21.0" has unmet peer dependency "typescript@>=2.8.0 || >= 3.2.0-dev || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >= 3.6.0-beta || >= 3.7.0-dev || >= 3.7.0-beta".
```
i wonder why typescript is needed at all when we dont even have any typescript nor typescript-eslint ?

mvorisek: fixed in https://github.com/fomantic/Fomantic-UI/pull/2589/commits/b5a02002458e6626d232d15a3c2a465369b789e6, `eslint-plugin-jest` needs several peer dependencies, but because of Node 12 support we cannot add Airbnb and Unicorn linter packages directly anyway, so I removed `eslint-plugin-jest` from `package.json` too and fixed the peer deps in the CI install workflow step only (the tests that need `eslint-plugin-jest` can eventually be dropped in the future, as they do not test much)

tasks/docs/serve.js
```diff
@@ -25,6 +26,8 @@ var
     assets      = require('../build/assets')
 ;
 
+let config = configConst;
```
lubber-de, line 29: What's the purpose of this change?
I understand config cant be a constant because of later changes, so why not remove the declaration from the const group and instead? Would be more clean, as otherwise "configSetup" is never used anywhere else
```
let config      = require('../config/docs');
```

tasks/docs/build.js
```diff
@@ -33,6 +33,8 @@ var
     log         = tasks.log
 ;
 
+let config = configConst;
```
lubber-de, line 36: Same here. Remove the config declaration form the const group and add as separate let statement
```javascript
let  config      = require('../config/docs');
```

tasks/docs/serve.js
mvorisek: in https://github.com/fomantic/Fomantic-UI/pull/2589/commits/f9cd84c64e2e751f7bcb1bed72da84cedd413670, in 2 files, `config` is assigned later in the script, so I fixed the problem quickly like I fixed it...

If I did not overlook something, `require('tasks/config/project/config').addDerivedValues()` can mutate the passed object, so now I fixed it properly in https://github.com/fomantic/Fomantic-UI/pull/2589/commits/bacb5cfedbb66f5f52875eaa2d67da0174e5a309, please review carefully

tasks/docs/build.js
mvorisek: fixed/discussed in https://github.com/fomantic/Fomantic-UI/pull/2589#discussion_r1045226791

tasks/docs/build.js
```diff
@@ -35,7 +36,8 @@ var
 
 module.exports = function (callback) {
     // use a different config
-    config = configSetup.addDerivedValues(config);
+    let config = extend(true, {}, configDocs);
```
lubber-de, line 39: Why so complicated rewrite? If config cannot be a const, please make it a let instead.

tasks/config/project/config.js
```diff
@@ -137,11 +133,11 @@ module.exports = {
         config.globs.components = componentsExceptIndividuals.length === 1 ? componentsExceptIndividuals[0] : '{' + componentsExceptIndividuals.join(',') + '}';
 
         // components that should be built, but excluded from main .css/.js files
-        config.globs.individuals = individuals.length === 1 ? individuals[0] : (individuals.length > 1)
-            ? '{' + individuals.join(',') + '}'
-            : undefined;
-
-        return config;
```
lubber-de, line 144: done remove the return, it's needed at 2 calls in other files

lubber-de:
Three things i dont like (yet) while reviewin all of th changes:
- There is too much changed into one single PR
- switching _properly_ from var to const/let needs much more love and patience and testing and, as i commented some times, leads into code rewrites which are unneccesarily complex than before.
- That many eslint comment hints again: If the linter is not doing what is expected then we should disable it (for now).
  Contributors which perhaps are not familiar with that eslint config could potentionally write "bad" code and have now idea what the linter is probably doing wrong. So when there seems to be a need of telling the linter to ignore things too often: Don't add the rule at all

tasks/config/project/config.js
mvorisek: in the problematic/fixed ones :)

mvorisek:
Even this PR is/was RTM, I marked it as a draft and will split it up in several separate PRs that will address only a few rules at once.

mvorisek:
> TODO: unify hex escape case - see [#2607 (comment)](https://github.com/fomantic/Fomantic-UI/pull/2607#discussion_r1046148597)

@lubber-de fixing stylelint for upper case colors can be done easily with `'color-hex-case': 'upper'` configuration, but I have checked the most commonly used coding standards and it seems lower case is used much more. Also advised by CS guides like: https://google.github.io/styleguide/htmlcssguide.html and https://docs.ckan.org/en/2.9/contributing/css.html

so I prefer to stay with the default LC for LESS/CSS colors and UC for escapes in JS.

Unified LC is not (easily) possible, as Unicorn has no configuration for it. Also UC for escapes in the only option with PHP CS Fixer, so it seems to be well accepted casing.

So if ok with you, I prefer too keep the casing rules as they are (and as default by the recommended CS AirBnb/Unicorn/Stylelint configs).

src/definitions/behaviors/state.js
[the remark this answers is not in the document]
mvorisek: marking as resolved here - moved to separate PR https://github.com/fomantic/Fomantic-UI/pull/2604

src/definitions/modules/progress.js
```diff
@@ -260,16 +260,13 @@
...
-
                         var
-                            value   = module.get.value(index),
+                            index_  = index || 0,
+                            value   = module.get.value(index_),
                             total   = module.get.total(),
                             percent = animating
-                                ? module.get.displayPercent(index)
-                                : module.get.percent(index),
+                                ? module.get.displayPercent(index_)
```
lubber-de, line 268: We have reduced that in the previous PRs...now you put in again?

src/definitions/modules/progress.js
```diff
@@ -260,16 +260,13 @@
 
                 get: {
                     text: function (templateText, index) {
-                        if (!index) {
```
lubber-de, line 263: We have explicitely added this in the previous PRs...now you put in again? Can we please find a "final standard" :)

src/definitions/behaviors/form.js
```diff
@@ -136,10 +136,7 @@
                 },
 
                 attachEvents: function (selector, action) {
-                    if (!action) {
```
lubber-de, line 139: We have explicitely added this in the previous PRs...now you remove in again? Can we please find a "final standard" :)

src/definitions/modules/progress.js
mvorisek: see https://github.com/fomantic/Fomantic-UI/pull/2589#discussion_r1043860140, when I was splitting PRs, I reverted this on top of the commits, thus the change was merged...

I am fine to keep this few lines for 2.10.x as agreed before, so reverted here.

lubber-de: I actually liked the changes now, especially for this codeblock (the mix using index_ is ugly)
By dropping IE11 we can simply remove the "if (!index)" block in the beginning.

Same for the other code (if !action), i dont want to change code again and again. So please dont change them agai. They are working fine as they are and satify the linter :)

mvorisek: perfect, revert reverted :)

lubber-de (approved):
LGTM

PR #2590, opened 2022-12-07
feat(lint): type-safe equality operator when comparing literals

mvorisek:
For comparing (non empty) strings, we can use `===` reliably (instead of `==`).

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -117,7 +117,7 @@
                             $module
                                 .on(triggerEvent + eventNamespace, module.event.trigger)
                             ;
-                        } else if (settings.on == 'now') {
+                        } else if (settings.on === 'now') {
                             module.debug('Querying API endpoint immediately');
                             module.query();
                         }
@@ -296,7 +296,7 @@
                     },
                     loading: function () {
                         return module.request
-                            ? module.request.state() == 'pending'
+                            ? module.request.state() === 'pending'
                             : false;
                     },
                     abortedRequest: function (xhr) {
@@ -334,13 +334,13 @@
                         return module.cancelled || false;
                     },
                     successful: function () {
-                        return module.request && module.request.state() == 'resolved';
+                        return module.request && module.request.state() === 'resolved';
                     },
                     failure: function () {
-                        return module.request && module.request.state() == 'rejected';
+                        return module.request && module.request.state() === 'rejected';
                     },
                     complete: function () {
-                        return module.request && (module.request.state() == 'resolved' || module.request.state() == 'rejected');
+                        return module.request && (module.request.state() === 'resolved' || module.request.state() === 'rejected');
                     },
                 },
 
@@ -525,7 +525,7 @@
                 event: {
                     trigger: function (event) {
                         module.query();
-                        if (event.type == 'submit' || event.type == 'click') {
+                        if (event.type === 'submit' || event.type === 'click') {
                             event.preventDefault();
                         }
                     },
@@ -614,15 +614,15 @@
                                 response     = module.get.responseFromXHR(xhr),
                                 errorMessage = module.get.errorFromRequest(response, status, httpMessage)
                             ;
-                            if (status == 'aborted') {
+                            if (status === 'aborted') {
                                 module.debug('XHR Aborted (Most likely caused by page navigation or CORS Policy)', status, httpMessage);
                                 settings.onAbort.call(context, status, $module, xhr);
 
                                 return true;
                             }
-                            if (status == 'invalid') {
+                            if (status === 'invalid') {
                                 module.debug('JSON did not pass success test. A server-side error has most likely occurred', response);
-                            } else if (status == 'error') {
+                            } else if (status === 'error') {
                                 if (xhr !== undefined) {
                                     module.debug('XHR produced a server error', status, httpMessage);
                                     // make sure we have an error to display to console
@@ -829,12 +829,12 @@
                         return data;
                     },
                     event: function () {
-                        if (isWindow(element) || settings.on == 'now') {
+                        if (isWindow(element) || settings.on === 'now') {
                             module.debug('API called without element, no events attached');
 
                             return false;
                         }
-                        if (settings.on == 'auto') {
+                        if (settings.on === 'auto') {
                             if ($module.is('input')) {
                                 return element.oninput !== undefined
                                     ? 'input'
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -478,7 +478,7 @@
                                 while (nameKeys.length > 0) {
                                     var k = nameKeys.pop();
 
-                                    if (k == '' && !Array.isArray(value)) { // foo[]
+                                    if (k === '' && !Array.isArray(value)) { // foo[]
                                         value = build([], pushes[pushKey]++, value);
                                     } else if (settings.regExp.fixed.test(k)) { // foo[n]
                                         value = build([], k, value);
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -12,7 +12,7 @@
     'use strict';
 
     function isWindow(obj) {
-        return obj != null && obj === obj.window;
+        return obj !== null && obj === obj.window;
     }
 
     function isFunction(obj) {
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -1004,17 +1004,17 @@
                         query = query.split(/[ .]/);
                         maxDepth = query.length - 1;
                         $.each(query, function (depth, value) {
-                            var camelCaseValue = depth != maxDepth
+                            var camelCaseValue = depth !== maxDepth
                                 ? value + query[depth + 1].charAt(0).toUpperCase() + query[depth + 1].slice(1)
                                 : query
                             ;
-                            if ($.isPlainObject(object[camelCaseValue]) && (depth != maxDepth)) {
+                            if ($.isPlainObject(object[camelCaseValue]) && (depth !== maxDepth)) {
                                 object = object[camelCaseValue];
                             } else if (object[camelCaseValue] !== undefined) {
                                 found = object[camelCaseValue];
 
                                 return false;
-                            } else if ($.isPlainObject(object[value]) && (depth != maxDepth)) {
+                            } else if ($.isPlainObject(object[value]) && (depth !== maxDepth)) {
                                 object = object[value];
                             } else if (object[value] !== undefined) {
                                 found = object[value];
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -865,7 +865,7 @@
                         // For each new rule, check if there's not already one with the same type
                         $.each(newValidation.rules, function (_index, rule) {
                             if ($.grep(validation[name].rules, function (item) {
-                                return item.type == rule.type;
+                                return item.type === rule.type;
                             }).length === 0) {
                                 validation[name].rules.push(rule);
                             }
@@ -1169,7 +1169,7 @@
                     optional: function (identifier, bool) {
                         bool = bool !== false;
                         $.each(validation, function (fieldName, field) {
-                            if (identifier == fieldName || identifier == field.identifier) {
+                            if (identifier === fieldName || identifier === field.identifier) {
                                 field.optional = bool;
                             }
                         });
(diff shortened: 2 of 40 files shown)


src/definitions/behaviors/api.js
```diff
@@ -117,7 +117,7 @@
                             $module
                                 .on(triggerEvent + eventNamespace, module.event.trigger)
                             ;
-                        } else if (settings.on == 'now') {
+                        } else if (settings.on === 'now') {
```
mvorisek, line 120: @lubber-de just checking, am I right here the change is functionally equivalent?

lubber-de: String comparision should always be safe using === or am i not getting the question? :)

mvorisek: see the actual/old code, there is `==` in many places

mvorisek: please see the last commit - `==` must be reviewed very carefully case by case

these are true:
- `0 == '0'`
- `0 == ''`
- `null == undefined`

these are false:
- `null == ''`
- `null == 0`
- `undefined == ''`
- `undefined == 0`

thus we can autofix only `==` to `===` only for non-numerical strings

with this in mind, that is how I crafted this PR

lubber-de: Yes, thats what i meant by string comparison. Your research is (again) perfectly summarized 🙂

src/definitions/behaviors/form.js
```diff
@@ -345,14 +345,14 @@
...
-                        if (initialValue == null) {
+                        // Explicitly check for undefined/null here as value may be `false`, so ($el.data(dataInitialValue) || '') would not work
+                        if (initialValue === undefined || initialValue === null) {
                             initialValue = '';
                         } else if (Array.isArray(initialValue)) {
                             initialValue = initialValue.toString();
                         }
                         var currentValue = $el.val();
-                        if (currentValue == null) {
+                        if (currentValue === undefined) {
```
mvorisek, line 355: https://api.jquery.com/val/

jQuery 3.0.0 (or higher) is required, it was released in 2016, is a minimal version of jQuery documented somewhere in the docs?

mvorisek:
I have updated more `==` comparisons in my best intentions, as there are no tests, please review/test carefully.

If all fine, I have nothing to add to this PR.

src/definitions/behaviors/form.js
lubber-de: 🤔 The docs examples always show jquery 3.x , but infact it seems we havent explicitely told anywhere that jquery 3+ is needed.
only the package.json "says" you need jquery 3.4+
Regarding this LOC, the check for null would still fetch jquery 2, as jq3 will always return an array, so the next else branch would fetch it. Checking for undefined will never be reached it seems

lubber-de: I think we should replace this by 
```suggestion
                        if (currentValue === null || currentValue === undefined) {
```
Doesn't hurt, does it? What do you think?

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
             },
 
             minCount: function (value, minCount) {
-                if (minCount == 0) {
+                if (minCount === 0) {
```
lubber-de, line 2053: We should cast minCount to Integer as the minCount value is provided as String from the `get. ancillaryValue` function

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
...
                     return true;
                 }
-                if (minCount == 1) {
+                if (minCount === 1) {
```
lubber-de, line 2056: Same here. mincount is provided as string, thus would not match anymore

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
...
                     return value !== '';
                 }
 
                 return value.split(',').length >= minCount;
             },
 
             exactCount: function (value, exactCount) {
-                if (exactCount == 0) {
+                if (exactCount === 0) {
```
lubber-de, line 2064: See avove. exactcount is possibly provided as string

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
...
                     return value === '';
                 }
-                if (exactCount == 1) {
+                if (exactCount === 1) {
```
lubber-de, line 2067: Again. exactCount could be String

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
...
                     return value !== '' && value.search(',') === -1;
                 }
 
                 return value.split(',').length == exactCount;
             },
 
             maxCount: function (value, maxCount) {
-                if (maxCount == 0) {
+                if (maxCount === 0) {
```
lubber-de, line 2075: And again :) maxCount could be provided as string

src/definitions/behaviors/form.js
```diff
@@ -2039,32 +2050,32 @@
...
                     return false;
                 }
-                if (maxCount == 1) {
+                if (maxCount === 1) {
```
lubber-de, line 2078: annnd again :) maxCount could be String

src/definitions/modules/dropdown.js
```diff
@@ -1602,7 +1614,7 @@
                                 }
 
                                 // up arrow (traverse menu up)
-                                if (pressedKey == keys.upArrow) {
+                                if (pressedKey === keys.upArrow) {
```
lubber-de, line 1617: above, such if structure was converted into a switch/case construct. Why not here?

mvorisek: because eslint rule did not fix it :)

I am not huge fan of switch/case constructs because it adds 1 indentation level and `break`s, so I do not want to pursue this to be fixed if not needed

src/definitions/modules/slider.js
```diff
@@ -411,39 +411,43 @@
                         }
                         if (first || module.is.focused()) {
                             var step = module.determine.keyMovement(event);
-                            if (step != NO_STEP) {
+                            if (step !== NO_STEP) {
                                 event.preventDefault();
                                 switch (step) {
-                                    case SINGLE_STEP:
+                                    case SINGLE_STEP: {
```
lubber-de, line 417: Why is that extra brace encapsulation necessary? (same for all following cases)

src/definitions/modules/slider.js
```diff
@@ -616,12 +624,15 @@
                         }
 
                         switch (settings.labelType) {
-                            case settings.labelTypes.number:
+                            case settings.labelTypes.number: {
```
lubber-de, line 627: Again here: WhatÄs the purpose of that extra brace encapsulation?

src/definitions/modules/slider.js
mvorisek: originally I turned off this rule, but later I decised wrapping it in a block/curly brackets is better to go with the recommended config - it is more readable for larged blocks, also, it should limit the scope of declared variables inside the block

src/definitions/behaviors/form.js
```diff
@@ -730,42 +730,53 @@
...
 
                                         if (date !== null) {
-                                            if (settings.dateHandling == 'date') {
-                                                values[name] = date;
-                                            } else if (settings.dateHandling == 'input') {
-                                                values[name] = $calendar.calendar('get input date');
-                                            } else if (settings.dateHandling == 'formatter') {
-                                                var type = $calendar.calendar('setting', 'type');
+                                            switch (settings.dateHandling) {
+                                                case 'date': {
```
lubber-de, line 734: Extra braces again? What for? There is no extra let/const scope necesary (applies to all other cases)

src/definitions/behaviors/form.js
```diff
@@ -730,42 +730,53 @@
...
-                                                    default:
-                                                        module.debug('Wrong calendar mode', $calendar, type);
-                                                        values[name] = '';
+                                                    break;
+                                                }
+                                                case 'formatter': {
+                                                    var type = $calendar.calendar('setting', 'type');
+
+                                                    switch (type) {
+                                                        case 'date': {
```
lubber-de, line 748: I know i repeat myself, but there is no need for the overall applied scoping as values[names] belong to upper scope already

src/definitions/behaviors/form.js
mvorisek: braces in `case` block everywhere, like in `if`, asserted by default/recommended Unicorn eslint config

src/definitions/behaviors/form.js
mvorisek: see my reply above, a style asserted by CI, it should be fine

src/definitions/behaviors/form.js
mvorisek: > Doesn't hurt

when we need to support jQuery 1.x/2.x, it is needed :)

src/definitions/behaviors/form.js
lubber-de: i dont agree as long as is not really necessary because of scoping. I know you can do it almost everywhere, but without a  need for scoping, whats the point? just to satisly the linter? No. We can add braces once they are really necessary (you havent had them before and without them is valid syntax since ages.

src/definitions/behaviors/form.js
lubber-de: technically fine, i still dont agree without a need of scoping

mvorisek: the need of scoping if not evaluated by eslint for each block, it is only a reason why the rule exists, eslint simply wraps are `case` blocks in brackets

src/definitions/modules/slider.js
lubber-de: I dont agree. Stating better readability can be used almost everywhere 🙂 . We already have the 4 space indent. This is pretty much improved readability already. And there is no need for scoping here as no extra variabled are declared

src/definitions/behaviors/form.js
mvorisek: good catch 👍, I did not done https://github.com/fomantic/Fomantic-UI/pull/2590/commits/5e1720ea963a0e918da79fbc5ce99360edd1f6bb correctly

src/definitions/modules/slider.js
mvorisek: I have resolved other threads for this topic, it is not an issue, it is a CS implied by recommended eslint config.

in https://github.com/fomantic/Fomantic-UI/pull/2590/commits/86018ee62b21a6ad9bf7c31bacd6ccdae61080f7 the whole impact can be seen

we can of course disable it globally, but I actually prefer to keep code blocks wrapped in brackets, which is more readable as start/close bracket can be easily located

src/definitions/behaviors/form.js
mvorisek: it has a story, I have written a custom eslint rule for it and then did not reviewed the result correctly, thankfully, the problem is/was only in this file and these 6 places you found

and we can fix it nicely, by wrapping the user (maybe string) numbers in `Number(xx)` to cast them for a string comparison

related article: https://thisthat.dev/number-constructor-vs-parse-int/

we will actually support more even /wo `parseInt` - `Number('5.0') === 5`

src/definitions/modules/slider.js
lubber-de: ok, convinced

mvorisek:
Is there anything left?

There are still about 20 places where `eqeqeq` is not fixed, so we need to keep the rule turned off, but I do not feel confident enough to fix them, as I am not sure if only strings/numbers are expected there - regex to locate them: ` (?<![!=])[!=]=(?!=) `.

@lubber-de can you have a look? If we can reduce the places, maybe it will be worth to add specific eslint disable comments to the places where really needed and enforce the `eqeqeq` rule globally.

lubber-de:
> Is there anything left?
> 
> There are still about 20 places where `eqeqeq` is not fixed, so we need to keep the rule turned off, but I do not feel confident enough to fix them, as I am not sure if only strings/numbers are expected there - regex to locate them: `(?<![!=])[!=]=(?!=)`.
> 
> @lubber-de can you have a look? If we can reduce the places, maybe it will be worth to add specific eslint disable comments to the places where really needed and enforce the `eqeqeq` rule globally.

@mvorisek uhm, I am done with the PR as it is now and would merge it, no?  Lets do not change the PRs over and over again...re-reviewing takes time 😉

mvorisek:
let's merge it

lubber-de (approved):
LGTM

PR #2591, opened 2022-12-09
fix(chore): typos found by eslint

mvorisek:
The fixed lines were identified by eslint when developing #2589.

As the changes are fixing bugs (and thus change behaviour) or can potentially have side effects, they are made as a separate PR.


src/definitions/behaviors/visibility.js
```diff
@@ -246,8 +246,8 @@
                     ;
                     while (imagesLength--) {
                         cacheImage = document.createElement('img');
-                        cacheImage.onload = handleLoad;
-                        cacheImage.onerror = handleLoad;
+                        cacheImage.addEventListener('load', handleLoad);
+                        cacheImage.addEventListener('error', handleLoad);
```
mvorisek, line 250: eslint error:
```
Prefer `addEventListener` over `onerror`. Note that there is difference between
`{window,element}.onerror` and `{window,element}.addEventListener('error')`
```

src/definitions/modules/calendar.js
```diff
@@ -885,7 +885,7 @@
                         var endDate = module.get.endDate();
                         if (!!endDate && !!date && date > endDate) {
                             // selected date is greater than end date in range, so clear end date
-                            module.set.endDate(undefined);
+                            module.set.endDate();
```
mvorisek, line 888: as discussed in https://github.com/fomantic/Fomantic-UI/pull/2584#discussion_r1039529752

tasks/docs/metadata.js
```diff
@@ -53,7 +53,7 @@ function parser(file, callback) {
             /** @type {string} */
             text     = String(file.contents.toString('utf8')),
             lines    = text.split('\n'),
-            filename = file.path.substring(0, file.path.length - 4),
+            filename = file.path.slice(0, -4),
```
mvorisek, line 56: "should be" safe for typical inputs

src/definitions/behaviors/api.js
```diff
@@ -157,11 +157,6 @@
                 },
                 write: {
                     cachedResponse: function (url, response) {
-                        if (response && response === '') {
```
mvorisek, line 160: as discussed in https://github.com/fomantic/Fomantic-UI/pull/2588#discussion_r1040967294

src/definitions/modules/progress.js
```diff
@@ -606,7 +606,7 @@
                         }
                     },
                     barLabel: function (text) {
-                        $progresses.map(function (index, element) {
+                        $progresses.each(function (index, element) {
```
mvorisek, line 609: this fix of `array-callback-return` rule should be ok (please note this is valid for `$().each(...` and not `$.each(...` in general, as the later has args swapped)

there are 4 more lines with errors:

```
src\definitions\modules\calendar.js
  1106:221  error  Array.prototype.some() expects a value to be returned at the end of function  array-callback-return
  1140:80   error  Array.prototype.some() expects a value to be returned at the end of function  array-callback-return
  1156:70   error  Array.prototype.some() expects a return value from function                   array-callback-return
  1182:101  error  Array.prototype.some() expects a value to be returned at the end of function  array-callback-return
```

currently `.some()` always returns false, which is probably not desired

@lubber-de would you be please so king and help with fixing these 4 lines?

src/definitions/behaviors/visibility.js
lubber-de: As the attached `handleLoad` function has no parameters and does not need to access event properties, this change is fine.
window.onerror would expect a different parameter list.

src/definitions/modules/progress.js
mvorisek: Thank you. Let's merge this PR, I can then finish #2589.

lubber-de (approved):
LGTM

PR #1833, opened 2021-01-08
fix(colors): compile error of php less port

cmoeke:
## Description
Added semicolons after variable declarations

## Closes
fomantic#1832

Issue #1832, opened 2021-01-08
PHP port of the LESS fails to compile colors.less


cmoeke:
# Bug Report
Currently it's not possible to compile the less files with the [wikimedia PHP port of the LESS processor](https://github.com/wikimedia/less.php/).

The problem is that the PHP port requires a semicolon after each variable declaration. So it currently fails at the colors.less.

I know it's a bug of the wikimedia PHP port and i also will create an issue there, but it seems that they do not have the will to fix community errors.

So it would help me a lot if we could make this small adjustment.

## Steps to reproduce
1. Install a fixed version of https://github.com/wikimedia/less.php/ (https://github.com/wikimedia/less.php/pull/33 due to a other bug)
2. Run `/usr/bin/php vendor/wikimedia/less.php/bin/lessc '##FOMANTIC_PATH##/definitions/modules/accordion.less' > '##TARGET_PATH##/definitions/modules/accordion.css'`

## Actual result
![fail](https://user-images.githubusercontent.com/1618434/103990836-1edff180-5192-11eb-92be-e08a2a573fcd.png)

## Way to fix
Add a semicolon to:
https://github.com/fomantic/Fomantic-UI/blob/2f78f4e1cc8c576145667f0c50aff87ba820d1bf/src/themes/default/globals/colors.less#L486
https://github.com/fomantic/Fomantic-UI/blob/2f78f4e1cc8c576145667f0c50aff87ba820d1bf/src/themes/default/globals/colors.less#L604
https://github.com/fomantic/Fomantic-UI/blob/2f78f4e1cc8c576145667f0c50aff87ba820d1bf/src/themes/default/globals/colors.less#L619

Old:
```LESS
    isDark         : true;
    isVeryDark     : true;
  };
}
```

New:
```LESS
    isDark         : true;
    isVeryDark     : true;
  };
};
```

## Version
2.8.7

PR #2592, opened 2022-12-09
feat(lint): add stylelint for LESS and normalize whitespaces

mvorisek:
Like #2501, but for LESS. No functionality change.


.eslintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
         sourceType: 'module',
     },
     ignorePatterns: [
+        '!.*',
```
lubber-de, line 13: I dont get which files this pattern matches in the fui core. Or: What happens if that line is not added?
 Could you please explain?

src/themes/default/globals/colors.less
```diff
@@ -3,617 +3,617 @@
...
-  @info: {
-    color: @infoColor;
-  };
-  @success: {
-    color: @positiveColor;
-  };
-  @warning: {
-    color: @warningColor;
-  };
-};
```
lubber-de, line 619: Don't remove the semicolon here, it was intentially added by #1833 to support a buggy PHP LESS mentioned in #1832

src/themes/default/globals/colors.less
```diff
@@ -3,617 +3,617 @@
...
-    tertiary       : @blackTertiaryColor;
-    tertiaryHover  : @blackTertiaryColorHover;
-    tertiaryFocus  : @blackTertiaryColorFocus;
-    tertiaryActive : @blackTertiaryColorActive;
-    bright         : @blackBright;
-    brightHover    : @blackBrightHover;
-    isDark         : true;
-    isVeryDark     : true;
-  };
-};
```
lubber-de, line 486: Don't remove the semicolon here, it was intentially added by #1833 to support a buggy PHP LESS mentioned in #1832

package.json
```diff
@@ -77,7 +77,10 @@
     "auto-changelog": "^2.4.0",
     "eslint": "^8.28.0",
     "node-fetch": "^2.6.0",
-    "semver": "^7.3.5"
+    "postcss-less": "^6.0.0",
```
lubber-de, line 80: Please also add postcss@^8.3.5 because action ci log says so
```
warning " > postcss-less@6.0.0" has unmet peer dependency "postcss@^8.3.5".
```

.eslintrc.js
mvorisek: files starting with `.` are not fixed by eslint by default, this way they are, so the configs itself are

lubber-de: Ah, ok, thanks 👍🏼

package.json
mvorisek: 👍, fixed in https://github.com/fomantic/Fomantic-UI/pull/2592/commits/7614896e14bd6c69cf9789b56896646fb92c91d5

src/themes/default/globals/colors.less
mvorisek: fixed in https://github.com/fomantic/Fomantic-UI/pull/2592/commits/6c88db28178de841f3fec3eb1e20a511670a3e1c

src/themes/default/globals/colors.less
mvorisek: see https://github.com/fomantic/Fomantic-UI/pull/2592#discussion_r1044866718, fixed

lubber-de (approved):
LGTM

PR #2594, opened 2022-12-10
feat(lint): modernize/unify CSS but maintain IE11 compatibility

mvorisek:
Fix several stylelint `stylelint-config-standard` config advisories.

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -45,7 +45,6 @@ module.exports = {
         'block-no-empty': null, // 25 errors
         'selector-no-vendor-prefix': null, // 19 errors
         'selector-class-pattern': null, // 19 errors
-        'declaration-block-no-redundant-longhand-properties': null, // 12 errors
         'value-no-vendor-prefix': null, // 9 errors
         'font-family-no-missing-generic-family-keyword': null, // 8 errors
     },
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -48,7 +48,6 @@ module.exports = {
         'declaration-block-no-redundant-longhand-properties': null, // 12 errors
         'value-no-vendor-prefix': null, // 9 errors
         'font-family-no-missing-generic-family-keyword': null, // 8 errors
-        'shorthand-property-no-redundant-values': null, // 5 errors
         'declaration-block-no-duplicate-properties': null, // 2 errors
         'font-family-no-duplicate-names': null, // 2 errors
         'function-linear-gradient-no-nonstandard-direction': null, // 1 error
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -19,6 +19,7 @@ module.exports = {
         'function-no-unknown': null,
         'import-notation': null,
         indentation: null, // handled by Prettier
+        'keyframes-name-pattern': null,
         linebreaks: 'unix',
         'max-line-length': null,
         'no-descending-specificity': null,
@@ -44,7 +45,6 @@ module.exports = {
         'selector-no-vendor-prefix': null,
 
         // TODO
-        'keyframes-name-pattern': null, // 50 errors
         'selector-class-pattern': null, // 19 errors
     },
     reportNeedlessDisables: true,
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -48,7 +48,6 @@ module.exports = {
         'declaration-block-no-redundant-longhand-properties': null, // 12 errors
         'value-no-vendor-prefix': null, // 9 errors
         'font-family-no-missing-generic-family-keyword': null, // 8 errors
-        'declaration-block-no-duplicate-properties': null, // 2 errors
         'font-family-no-duplicate-names': null, // 2 errors
         'function-linear-gradient-no-nonstandard-direction': null, // 1 error
         'at-rule-no-vendor-prefix': null, // 1 error
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -14,6 +14,7 @@ module.exports = {
         'color-function-notation': 'legacy',
         'color-hex-case': 'lower',
         'declaration-colon-newline-after': null, // handled by Prettier
+        'font-family-no-missing-generic-family-keyword': null,
         'function-no-unknown': null,
         'import-notation': null,
         indentation: null, // handled by Prettier
@@ -46,7 +47,6 @@ module.exports = {
         'selector-no-vendor-prefix': null, // 19 errors
         'selector-class-pattern': null, // 19 errors
         'value-no-vendor-prefix': null, // 9 errors
-        'font-family-no-missing-generic-family-keyword': null, // 8 errors
     },
     reportNeedlessDisables: true,
 };
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -11,6 +11,7 @@ module.exports = {
         'at-rule-empty-line-before': null,
         'at-rule-name-case': null,
         'at-rule-no-unknown': null,
+        'block-no-empty': null,
         'color-function-notation': 'legacy',
         'color-hex-case': 'lower',
         'declaration-colon-newline-after': null, // handled by Prettier
@@ -43,7 +44,6 @@ module.exports = {
 
         // TODO
         'keyframes-name-pattern': null, // 50 errors
-        'block-no-empty': null, // 25 errors
         'selector-no-vendor-prefix': null, // 19 errors
         'selector-class-pattern': null, // 19 errors
     },
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -43,9 +43,6 @@ module.exports = {
         'alpha-value-notation': 'number', // https://caniuse.com/mdn-css_properties_opacity_percentages
         'selector-not-notation': null, // https://caniuse.com/css-not-sel-list
         'selector-no-vendor-prefix': null,
-
-        // TODO
-        'selector-class-pattern': null, // 19 errors
     },
     reportNeedlessDisables: true,
 };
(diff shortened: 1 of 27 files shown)


src/themes/resetcss/globals/reset.overrides
```diff
@@ -91,8 +91,8 @@ video {
     margin: 0;
     padding: 0;
     border: 0;
-    font-size: 100%;
     font: inherit;
+    font-size: 100%;
```
mvorisek, line 95: this fixes `declaration-block-no-shorthand-property-overrides` stylelint rule

is this change ok or does it have any side effects?

src/definitions/elements/label.less
```diff
@@ -899,7 +899,6 @@ a.ui.active.label:hover::before {
     .ui[class*="bottom pointing"].label::before,
     .ui[class*="pointing below"].label::before {
         border-width: 0 @borderWidth @borderWidth 0;
-        top: auto;
```
mvorisek, line 902: see https://github.com/fomantic/Fomantic-UI/pull/2594/commits/52df34588293a7ceb3d372c9700ac201f0ba979a, maybe there is some typo in the code

src/definitions/modules/dropdown.less
```diff
@@ -436,7 +436,6 @@ select.ui.dropdown {
         backface-visibility: hidden;
         -webkit-overflow-scrolling: touch;
         border-top-width: 0 !important;
-        width: auto;
```
mvorisek, line 439: see https://github.com/fomantic/Fomantic-UI/pull/2594/commits/52df34588293a7ceb3d372c9700ac201f0ba979a, maybe there is some typo in the code

src/themes/default/globals/reset.overrides
```diff
@@ -62,7 +62,7 @@ hr {
  */
 
 pre {
-    font-family: monospace, monospace; /* 1 */
+    font-family: monospace; /* 1 */
```
mvorisek, line 65: just a typo, or is specifying font family twice fixes something?

src/themes/resetcss/globals/reset.overrides
lubber-de: i think, this change is ok.

.stylelintrc.js
```diff
@@ -18,41 +19,31 @@ module.exports = {
         'no-extra-semicolons': null, // fix GH-1832 - workaround for wikimedia/less.php parser
         'number-max-precision': 5,
         'rule-empty-line-before': null,
+        'string-quotes': 'single',
```
lubber-de, line 22: Let's stick to double quotes for HTML/CSS/JSON and single quotes for JS. This will also heavily reduce the changes of this PR. Doublequotes are more prominent and readable in selectors and html attributes.

src/themes/github/modules/dropdown.overrides
```diff
@@ -40,10 +40,10 @@
 
 .ui.dropdown .item .left.dropdown.icon::before,
 .ui.dropdown .left.menu .item .dropdown.icon::before {
-    content: "\f0d9"/*rtl:"\f0da"*/;
```
lubber-de, line 43: This comment hint is needed for rtl build

src/themes/github/modules/dropdown.overrides
```diff
@@ -40,10 +40,10 @@
...
+    content: '\f0d9';
 }
 
 /* Vertical Menu Dropdown */
 .ui.vertical.menu .dropdown.item > .dropdown.icon::before {
-    content: "\f0da"/*rtl:"\f0d9"*/;
```
lubber-de, line 48: This comment hint is needed for rtl build

src/themes/default/modules/dropdown.overrides
```diff
@@ -35,12 +35,12 @@
 
 .ui.dropdown .item .left.dropdown.icon::before,
 .ui.dropdown .left.menu .item .dropdown.icon::before {
-    content: "\f0d9"/*rtl:"\f0da"*/;
+    content: '\f0d9';
 }
 
 /* Vertical Menu Dropdown */
 .ui.vertical.menu .dropdown.item > .dropdown.icon::before {
-    content: "\f0da"/*rtl:"\f0d9"*/;
```
lubber-de, line 43: Comment hint is needed for rtl build

src/themes/default/globals/reset.overrides
lubber-de: Explained here https://github.com/necolas/normalize.css/issues/519#issuecomment-197131966 . So it's intention

src/themes/default/elements/icon.variables
```diff
@@ -455,7 +455,7 @@ Icons are order A-Z in their group, Solid, Outline, Thin (Pro only) and Brand
     folder_minus: \f65d;
     folder_open: \f07c;
     folder_plus: \f65e;
-    font: \f031;
+    font: \f031; // stylelint-disable-line font-family-name-quotes
```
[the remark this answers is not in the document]
lubber-de: ok, we just need to keep in mind, in case we update the icons via https://github.com/fomantic/create-fomantic-icons 
where this hint would otherwise get lost again

src/definitions/elements/label.less
lubber-de: This fix here is ok, top:100% wins all the time

.stylelintrc.js
mvorisek: I was testing the waters here :), I completely agree double quotes seems more natural in HTML attributes and I was supprised there is no separate config for it.

Using double quotes for LESS values is however not much natural too, so I will keep the rule turned off and commit only double quotes -> single quotes changes in LESS values manually.

src/definitions/modules/dropdown.less
lubber-de: Change is ok. below `width: @menuMinWidth;` wins all the time

src/themes/default/elements/icon.variables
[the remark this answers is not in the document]
mvorisek: I figured out the solution is to quote the values - fixed in separated PR, see https://github.com/fomantic/Fomantic-UI/pull/2610/commits/76b17f7eb9f306d133e9cbd1209876884f6471ba#r1049050447

src/themes/default/elements/segment.variables
```diff
@@ -29,7 +29,7 @@
 @groupedBoxShadow: @boxShadow;
 @groupedBorderRadius: @borderRadius;
 
-@nestedGroupMargin: @verticalMargin @verticalMargin;
+@nestedGroupMargin: @verticalMargin;
```
mvorisek, line 32: vertical twice or a typo?

lubber-de: From the git history back in 2015, it seems it was intention to have all margins the same value, but of course only one `@verticalMargin` is necessary to achieve the same result (i tested `@verticalMargin @horizontalMargin` in case it was a type, but this has a complete different visual (not looking right) result)

src/themes/amazon/elements/button.overrides
```diff
@@ -1,5 +1,5 @@
 .ui.button {
-    background-image: linear-gradient(center top, #f7f8fa, #e7e9ec) repeat scroll 0 0 rgba(0, 0, 0, 0);
+    background-image: linear-gradient(to top, #f7f8fa, #e7e9ec) repeat scroll 0 0 rgba(0, 0, 0, 0);
```
mvorisek, line 2: `center` seems to be really not valid, how can it be fixed?

lubber-de: The whole line is invalid (and even the gradient direction is wrong (i tested the theme)
a valid value is
```suggestion
    background: linear-gradient(to bottom, #f7f8fa, #e7e9ec) repeat scroll 0 0 rgba(0, 0, 0, 0);
```

src/themes/material/elements/icon.overrides
```diff
@@ -1524,16 +1524,16 @@ i.icon.lock.outline::before {
 i.icon.looks::before {
     content: "\e3fc";
 }
-i.icon.looks.3::before {
+i.icon.looks.three::before {
```
mvorisek, line 1527: class name starting with digit is not good, is this change ok? what are the source data? it seems some numbers are missing, especially the last change looks fishy

lubber-de: It's ok, we did that for the default font awesome icons as well

lubber-de: The official way to make it work anyway would be to escape it 
`.\32 `
Or use the attribute selector
`[class*="2"]`
But (!) The ever since existing notation `.2` in the material theme, never worked anyway, so nobody was ever complaining about, so most probably nobody ever used those icons 😉

So, again, changing them to words is totally ok and consistent to the default

mvorisek: A lot of changes have been done manually, please double check them, if all good, PR is finished.

src/themes/material/elements/button.variables
```diff
@@ -25,7 +25,7 @@
 
 @borderBoxShadowColor: @solidBorderColor;
 @borderBoxShadow: 0 0 0 1px @solidBorderColor inset;
-@shadowBoxShadow: 0 0 0 0 transparent;
+@shadowBoxShadow: 0 transparent;
```
lubber-de, line 28: according to https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow this is invalid. (and using chrome dev tools confirms that)
At least two length values have to be given, so you should change this to
```suggestion
@shadowBoxShadow: 0 0 transparent;
```

mvorisek: fixed, I did not intended to change box shadow values

lubber-de (approved):
LGTM

PR #2596, opened 2022-12-11
feat(lint): add Airbnb and Unicorn eslint packages

mvorisek:
Add eslint plugin packages with recommended rules to be fixed in separate PRs.


mvorisek:
This PR (part 1 of #2589) is complete.

tasks/docs/serve.js
```diff
@@ -27,7 +28,8 @@ var
 
 module.exports = function () {
     // use a different config
-    config = configSetup.addDerivedValues(config);
+    const config = extend(true, {}, configDocs);
```
lubber-de, line 31: This is still a confusing change. Please revert. addDerivedValues returns config and config is declared via let, so what's the purpose of creating a temporary const which extends the previous config.... This is confusing.

mvorisek: `addDerivedValues()` was written like when a config was given, the config was a) mutated, b) returned

now it keeps mutating it (as you would expect by the function name) and we extend/clone the config from import (before mutation) correctly

with this explanation, is it fine?

mvorisek: `addDerivedValues()` was written quite complited. When no/empty config was given, it created it. Why? When such fallback is removed, no need for return.

By the function name I would expect a void method that adds some config values. This is how it is coded now, is that ok?

The change also fixes the mutation of a config returned from an imported module, previously, it was not extended/cloned. IMHO it is a good practise to never mutate anything imported.

lubber-de: It is still confusing: Why is configDocs now declared outiside of the exports scope? For nothing else than being extended for the const once.... Why not move the `let config      = require('../config/docs'),` into the module.exports block and keep the functionality as it was before?

lubber-de:
@mvorisek 

Don't get me wrong, i highly appreciate what you are doing here  👍🏼  But you are preparing lots of similiar PRs always containing several rules and if only one of them turns into discussion we never get close into getting most of them included :)

I actually dont want to rewrite ancient (but working) code just because airbnb rules now tell us we should code different.
linting rules are still opinion based. 😉 

#### That said, i am totally fine with many other rules which do not need code structure changes , but please keep the whole var/let/const changes in a separate PR which does nothing else. Thanks. 

I am fine with a lot of PRs having a single lint rule. This makes approvement/review much easiert and we can concentrate on one topic and merging gets faster as we wont loose time (or me investigating what code, i never touched before, is doing and why the linter thinks it has to be changed.

### You are doing a great job these days, thanks a lot! 🙂

tasks/docs/serve.js
mvorisek: because it is a good practise to keep imports at the top (and never mutate them) - otherwise eslint with recommended config will warn - otherwise we will need `eslint-disable-line`

lubber-de: ok, understood, but then it should be kept as var instead if let if we get into scrop issues otherwise ;)

mvorisek: If imported on top (with var/let/const), then it is visible to the whole script scope.

mvorisek:
> But you are preparing lots of similiar PRs always containing several rules

I have a "master" PR #2589 which I split and split :)

The older PRs are to be merged later, they hold changes not compatible with Node 12/ (and E11).

Thankfully, I made the changes quite strictly rule by rule so I can reduce the PR size easily.

This PR now adds the linter packages only.

tasks/docs/serve.js
mvorisek: now part of #2597

see https://github.com/fomantic/Fomantic-UI/pull/2597/commits/117dc67a0369bc3d5f70a747081261eb03856825, we import the config as `configDocs`, then create (local/cloned) `config` config.

lubber-de (approved):
LGTM :)

lubber-de:
nevermind.... found the issue :wink:

lubber-de:
@mvorisek Another "issue".
Since the last merge my jetbrains IDE eslint moans about me using windows.
```
Expected linebreaks to be 'LF' but found 'CRLF'
```

If this is a default rule in airbnb-ruleset, how can we override that? `linebreak-style` is not set inside our own eslintrc.

my git cofig is configured to use LF only when pushing but CRLF when checking out on windows (standard git setting under windows "core.autolcrlf=true"). Of coruse i dont want to see every code line now being an error, so we should override that rule. Not everybody is using a mac or linux for web development :)

lubber-de:
This seems to be a reliabled override ?

```
rules: {
        'linebreak-style': ['error', process.platform === 'win32' ? 'windows' : 'unix'],
 }
```

lubber-de:
If possible and you are in doubt, sombody could accidently push CRLF: Could we restrict the rule to LF inside CI only?

mvorisek:
LF must be used in all source text files

if there is any text file stored as CRLF, please fix

if you have LF locally, I advise to reconfigure your IDE to LF which is used by 99% projects across many programming languages

if you have locally CRLF, it might come from git `autocrlf`, disable it, I would not advise to let git transform your text files locally, otheriwse you will hit many problems like eslint wanting LF :)

PR #2597, opened 2022-12-11
feat(chore): use modern "let" instead of "var" in build scripts

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on fixing `no-var` rule.

For src/* files it cannot be fixed (now) because of limited support in IE11.


mvorisek:
Hi @lubber-de, if there is no feedback left let's merge this PR and then #2598 and #2599 (in this order, there should be no conflict).

lubber-de (approved):
I finally understood your "configDocs" change (i actually learned/refreshed a bit of my javascript knowledge today). Thanks for making a single PR out of it, was worth it (for me 😆 )

LGTM

lubber-de:
> Hi @lubber-de, if there is no feedback left let's merge this PR and then #2598 and #2599 (in this order, there should be no conflict).

Both other PRs have merge conflicts now. Please resolve 🙂

PR #2598, opened 2022-12-11
feat(lint): fix multiple rare eslint errors

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on fixing multiple rare rules, done in one PR, as otherwise there will be only a few LOC changed per separate PR.


scripts/nightly-version.js
```diff
@@ -4,8 +4,8 @@ const path = require('path');
 const childProcess = require('child_process');
 
 // npm
-const fetch = require('node-fetch');
-const semver = require('semver');
+const fetch = require('node-fetch'); // eslint-disable-line import/no-extraneous-dependencies
+const semver = require('semver'); // eslint-disable-line import/no-extraneous-dependencies
```
mvorisek, line 8: the ignore comment is legit, nightly needs only a dev dependency

lubber-de (approved):
LGTM... you are on fire :D

PR #2602, opened 2022-12-12
Unify regex and .length check syntax

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on unified regex and .length check usage, also removes useless escapes.

Needs #2599 merged first.


lubber-de:
Please resolve conflicts

PR #2604, opened 2022-12-12
feat(lint): remove global undefined variable shadow

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on fixing `no-shadow-restricted-names` rule.

Discussed in https://github.com/fomantic/Fomantic-UI/pull/2589#discussion_r1045145586, it fixes the rule cleanly and the impact on the total JS size is minimal (<1%).


lubber-de:
@fomantic/maintainers  @fomantic/admins 

What's your opinion about this? See comments regarding negative minimizer impact here
https://github.com/fomantic/Fomantic-UI/pull/2589#discussion_r1042689523

mvorisek:
The `undefined` can be still redefined in parent/including context in theory, but every other global variable like `Proxy` or `window` can, so this cannot be the reason to keep it. The size increase of less than 1% is really negligible and can be gained easily with other changes.

lubber-de (approved):
LGTM

PR #2605, opened 2022-12-12
feat(lint): fix else with return, unify ternary CS

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR focuses on fixing rules like `no-else-return` and other ternary operator related.


lubber-de:
Please revolve conflicts

lubber-de (approved):
LGTM

PR #2607, opened 2022-12-12
feat(lint): fix various eslint rules

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR fixes several smaller rules not worth a separate PRs.


src/definitions/modules/dropdown.js
```diff
@@ -3206,7 +3209,7 @@
                         }
                     },
                     diacritics: function (text) {
-                        return settings.ignoreDiacritics ? text.normalize('NFD').replace(/[\u0300-\u036f]/g, '') : text;
```
lubber-de, line 3209: for stylelint we added a rule to always have lowercase rgb values. here we are forcing uppercase... shouldn't his be consistent?

mvorisek: I noticed this inconsistency too - I have added this as TODO in the master PR - https://github.com/fomantic/Fomantic-UI/pull/2589#issuecomment-1346981377

for now, I prefer this PR to get merged as is to reduce later merge conflicts, once all changes are landed, we can easily update the eslint/stylelint config and autofix the code

lubber-de (approved):
LGTM

PR #2608, opened 2022-12-12
feat(lint): nested ternary formatting and remove unneeded parentheses

mvorisek:
Modernize code using eslint advisory introduced in GH-2596. This PR updates ternary operator formatting for better readability and and removes unneeded parentheses around expressions.


lubber-de (approved):
LGTM

PR #2610, opened 2022-12-14
feat(lint): unify LESS files whitespaces and string quotes

mvorisek:
This PR is part of GH-2594 but separated to integrate many whitespace changes (and other non functional changes) separately.

PR #2611, opened 2022-12-15
feat(lint): add prettier formatter to assert LESS files whitespaces

mvorisek:
Compared to Stylelint, Prettier is very simple tool and I am quite supprised Stylelint is deprecating the whitespace rules in favor of Prettier. For LESS it seems ok, for JS, we should never enable Prettier as formatting is part of a code and improves readability.

As Prettier is an opinionated formatter instead of an linter, we need to satisfy it everywhere without exceptions.

When I was crafting this PR, I found several Prettier issues and proposed changes to Prettier:

- https://github.com/prettier/prettier/pull/14007
- https://github.com/prettier/prettier/pull/14008
- https://github.com/prettier/prettier/pull/14034

In this PR, these changes are contained and Prettier is patched before it is run. Once the changes are merged in prettier and stable release is made, they can be removed.

This PR fixes minor whitespaces unfixed/undetected in GH-2610.

And also asserts:
- GH-1832
- https://github.com/fomantic/Fomantic-UI/pull/2593#discussion_r1045132780

Prettier has no built in support to dump the diff - https://github.com/prettier/prettier/issues/6885 - so we dump it in the CI using git.


.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+    tabWidth: 4,
+    printWidth: Infinity, // TODO
```
mvorisek, line 3: TODO change to 120 before merge

package.json
```diff
@@ -26,8 +26,8 @@
   "scripts": {
     "build": "gulp build",
     "changelog": "auto-changelog -p",
-    "lint": "eslint --ext .js . && stylelint \"**/*.{css,less,overrides,variables}\"",
-    "lint-fix": "eslint --ext .js . --fix && stylelint \"**/*.{css,less,overrides,variables}\" --fix"
+    "lint": "eslint --ext .js . && stylelint \"**/*.{css,less,overrides,variables}\" && prettier --loglevel warn \"!dist\" \"!test/coverage\" \"**/*.XX{css,less,overrides,variables}\" --check",
```
mvorisek, line 29: TODO no `XX` before merge, currently, Prettier is disabled until all CS problems are fixed

src/semantic.less
```diff
@@ -13,63 +13,167 @@
 // stylelint-disable no-invalid-position-at-import-rule
 
 /* Global */
-& { @import 'definitions/globals/reset'; }
-& { @import 'definitions/globals/site'; }
+& {
```
mvorisek, line 16: not desired, see https://github.com/fomantic/Fomantic-UI/pull/2612#discussion_r1050598585

mvorisek: fixed - file is ignored by Prettier

mvorisek:
PR is done, Prettier fixes with patches looks reasonable. Compiled `semantic.min.css` is binary the same as before this PR.

lubber-de:
Please resolve conflicts

lubber-de:
> In this PR, these changes are contained and Prettier is patched before it is run. Once the changes are merged in prettier and stable release is made, they can be removed.

Seriously? This is really hacky. You want to use a tool, which does not do what we need... Why not fork prettier.io instead and use that?

mvorisek:
> Please resolve conflicts

done

> Seriously? This is really hacky. You want to use a tool, which does not do what we need... Why not fork prettier.io instead and use that?

stylelint should be enough to fix style locally, the remaining issues detected by prettier are detected by CI (or you can patch your local prettier) - I have chosen patches as the patches are managed in this repo, not in external fork

non-compiled prettier have also many deps for every supported language, so we do want to add non-compiled prettier dep

src/definitions/collections/grid.less
```diff
@@ -1454,11 +1454,15 @@
         & when (@variationGridDivided) {
             /* Divided Reversed */
             .ui[class*="computer reversed"].divided.grid:not([class*="vertically divided"]) > .column:first-child,
```
lubber-de, line 1456: Why isnt this line also put into a separate line (because of the >) . This rule is very inconsistent to me.

src/definitions/elements/button.less
```diff
@@ -897,11 +904,8 @@
     .ui.tertiary.button {
         transition: color @defaultDuration @defaultEasing !important;
         border-radius: 0;
-        margin:
-            (@verticalPadding - @tertiaryVerticalPadding)
-            (@horizontalMargin)
-            (@verticalPadding + @shadowOffset + @verticalMargin - @tertiaryVerticalPadding)
-            0 !important;
+        margin: (@verticalPadding - @tertiaryVerticalPadding) (@horizontalMargin)
```
lubber-de, line 907: I actually find the previous formatting more readable than the new one tbh

src/definitions/elements/input.less
```diff
@@ -480,7 +480,8 @@
             padding-right: @labeledMargin;
         }
     }
-    .ui.ui[class*="corner labeled"]:not([class*="left corner labeled"]).icon:not([class*="left icon"]).input > .ui.dropdown {
+    .ui.ui[class*="corner labeled"]:not([class*="left corner labeled"]).icon:not([class*="left icon"]).input
+        > .ui.dropdown {
```
lubber-de, line 484: The indentation here is confusing. The > not starts at the same position as the next & . To me this is more confusing than the previous formatting

src/definitions/modules/dropdown.less
```diff
@@ -652,7 +654,8 @@ select.ui.dropdown {
                 }
 
                 .ui.inverted.dropdown .scrollhint.menu:not(.hidden)::before {
-                    border-image: linear-gradient(to bottom, @invertedScrollhintStartColor, @invertedScrollhintEndColor) 1 100%;
+                    border-image: linear-gradient(to bottom, @invertedScrollhintStartColor, @invertedScrollhintEndColor)
+                        1 100%;
```
lubber-de, line 658: This formatting doesnt make sense to me, it's much less readable.

.github/workflows/ci.yml
```diff
@@ -11,9 +11,19 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
     - name: Install dependencies
-      run: yarn && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript
+      run: >
+        yarn
+        && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript github:prettier/prettier#^2.8.1
```
lubber-de, line 16: This adds prettier@^2.8.1. That means it could also add a later version in the same major branch (like for example 2.12.3). I doubt the fixed git patches will always match and work, so i believe in a future situation the ci will break. Thats's why i suggest using an own fork. I also doubt the issues you opened get fixed soon

src/definitions/collections/grid.less
mvorisek: because of maxLength = 120

Prettier tries to put almost everything on 1 line and if it is too long, then it wraps it

impossible to control/setup additional wrapping rules, the only exception is to wrap after after comma `,`, I have patched it to be reasonably ok

src/definitions/elements/button.less
mvorisek: me too, but Prettier is not a linter, and cannot be configured, as long as Stylelint requires Prettier, we cannot do much except patching Prettier more, which is not easy I would like to avoid more patches than really needed

.github/workflows/ci.yml
mvorisek: That is intended, I expect the patches to be integrated and new Prettier released within a month or so, then CI should fail and we should remove the patches :)

src/themes/default/modules/dropdown.variables
```diff
@@ -344,7 +347,8 @@
 @scrollingWidescreenMaxMenuHeight: (@scrollingItemHeight * @selectionWidescreenMaxItems);
 
 /* Upward */
-@upwardSelectionVisibleBorderRadius: @selectionVisibleConnectingBorder @selectionVisibleConnectingBorder @borderRadius @borderRadius;
+@upwardSelectionVisibleBorderRadius: @selectionVisibleConnectingBorder @selectionVisibleConnectingBorder @borderRadius
```
lubber-de, line 350: This also is more confusing than before.

src/themes/default/modules/dropdown.variables
```diff
@@ -252,12 +253,14 @@
...
 @multipleSelectionRightPadding: @selectionIconDistance;
-@multipleSelectionPadding: @multipleSelectionVerticalPadding @multipleSelectionRightPadding @multipleSelectionVerticalPadding @multipleSelectionLeftPadding;
+@multipleSelectionPadding: @multipleSelectionVerticalPadding @multipleSelectionRightPadding
+    @multipleSelectionVerticalPadding @multipleSelectionLeftPadding;
 
 /* Child Elements */
 @multipleSelectionChildVerticalMargin: (@searchSelectionVerticalPadding * (2/3));
 @multipleSelectionChildLeftMargin: (@inputHorizontalPadding - @multipleSelectionLeftPadding);
-@multipleSelectionChildMargin: @multipleSelectionChildVerticalMargin 0 @multipleSelectionChildVerticalMargin @multipleSelectionChildLeftMargin;
+@multipleSelectionChildMargin: @multipleSelectionChildVerticalMargin 0 @multipleSelectionChildVerticalMargin
```
lubber-de, line 262: Same here, putting only the last margin into a separate line makes no sense to me

.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+    tabWidth: 4,
+    printWidth: 120,
```
lubber-de, line 3: This needs to be increased or removed. It too often let into moving single values from into a separate line which makes them more unreadable than before (most for margins where each margin has a tool long variable name

mvorisek: it defaults to 80, but can be set to `Infinity`

.github/workflows/ci.yml
lubber-de: i actually don't want the ci to fail if it's forseeable... really

.prettierrc.js
mvorisek: done - is it really better now?

src/definitions/elements/button.less
lubber-de: Well, then we should really think twice, if we really want to use prettier then, if it causes more trouble than help 🤔

src/definitions/collections/grid.less
lubber-de: As stated and suggest by you below, we should set this to infinity or remove it then. As already said, causes more trouble than it helps

src/definitions/elements/button.less
mvorisek: eslint is perfect, no need for Prettier fo JS

but stylelint for LESS is quite buggy, it does not detect a lot of formatting problems, so I would like to find a solution how to integrate with Prettier, not mentioning next release of Stylelint will rely on Prettier solely

.github/workflows/ci.yml
mvorisek: I have removed the `^` caret symbol, so the Prettier version is fixed

src/definitions/collections/grid.less
mvorisek: done in https://github.com/fomantic/Fomantic-UI/pull/2611/commits/470cd4e7b1dbce900d5e006aa8f3c5d7bf850009

src/themes/default/elements/placeholder.variables
```diff
@@ -34,16 +34,8 @@
...
-rgba(0, 0, 0, 0.08) 0,
-rgba(0, 0, 0, 0.15) 15%,
-rgba(0, 0, 0, 0.08) 30%
-);
-@placeholderInvertedLoadingGradient: linear-gradient(to right,
-rgba(255, 255, 255, 0.08) 0,
-rgba(255, 255, 255, 0.14) 15%,
-rgba(255, 255, 255, 0.08) 30%
-);
+@placeholderLoadingGradient: linear-gradient(to right, rgba(0, 0, 0, 0.08) 0, rgba(0, 0, 0, 0.15) 15%, rgba(0, 0, 0, 0.08) 30%);
```
lubber-de, line 37: Why is this now concatenated into one line while in other cases (box-shadow, every list in variation.variables, font-family, etc. )using comma separated lists, they are kept? Sometimes it helps for better readability, sometime not :(

src/themes/default/elements/segment.variables
```diff
@@ -132,21 +131,9 @@
...
 
 @secondaryInvertedLightness: 0.2;
-@secondaryInvertedBackground:
-    lighten(@black, (@secondaryInvertedLightness * 100))
-    linear-gradient(
-        rgba(255, 255, 255, @secondaryInvertedLightness) 0,
-        rgba(255, 255, 255, @secondaryInvertedLightness) 100%
-    )
-;
+@secondaryInvertedBackground: lighten(@black, (@secondaryInvertedLightness * 100)) linear-gradient(rgba(255, 255, 255, @secondaryInvertedLightness) 0, rgba(255, 255, 255, @secondaryInvertedLightness) 100%);
```
lubber-de, line 134: same here: we should ignore/keep function parameters lists in separates lines just like for usual lists (box-shadow, etc)

src/themes/default/globals/site.variables
```diff
@@ -14,22 +14,168 @@
...
+    U+2C60-2C6C,
+    U+2C74-2C77;
 
 @importFonts: true;
 @fonts: {
     @regularLatin: {
         font-family: @fontName;
-        src: url("@{fontPath}/@{fontFileNameLatin}-Regular.woff2") format("woff2")
-            if(@supportIE, e(",") url("@{fontPath}/@{fontFileNameLatin}-Regular.woff") format("woff"));
+        src: url("@{fontPath}/@{fontFileNameLatin}-Regular.woff2") format("woff2") if(@supportIE, e(",") url("@{fontPath}/@{fontFileNameLatin}-Regular.woff") format("woff"));
```
lubber-de, line 178: can we keep this extra line? its more readable and the "if" otherwise gets visually lost.

src/themes/default/elements/placeholder.variables
mvorisek: https://github.com/fomantic/Fomantic-UI/pull/2611/commits/470cd4e7b1dbce900d5e006aa8f3c5d7bf850009

src/themes/default/elements/segment.variables
mvorisek: impossible, no ignore in file supported, the only way is to patch Prettier but I am not sure if such patch will ever be merged into Prettier directly...

.prettierrc.js
lubber-de: Sometimes yes, sometimes not (lists in functions like linear-gradient)

src/themes/default/elements/segment.variables
lubber-de: > impossible, no ignore in file supported, the only way is to patch Prettier but I am not sure if such patch will ever be merged into Prettier directly...

Another reason to create a fork which does properly what we want :wink:

mvorisek: It is all about 1 file - https://github.com/prettier/prettier/blob/main/src/language-css/printer-postcss.js

I would still prefer no fork and replace the CSS/LESS printer only.

All the required files should be exported - https://github.com/prettier/prettier/blob/main/src/language-css/index.js - so it should be possible to create custom printer that will fallback to the native one. I guess it can be done in ~100 lines of JS code. But not easy ones to find out :)

I can look into it quickly, but maybe we can be fine with one extra patch of function args to be on separate like when separated with comma.

lubber-de: I tried a print-width of 0 and for this case the result is more readable in a whole, yet more lines, but more consistent.
It's still not perfect ( the * 100 is also put into a new line)
However, i did not try the whole code, just a quick test on

https://prettier.io/playground/#N4Igxg9gdgLgprEAuEABAznSUAmBDAJwE8BJKANzgPhwCE8wBrAcwIgFdckAdKAAgF8ANgEtmAC3hQAFKgBGQhowA0fWZmz5iZStTg4AMmMlQ46dHwBUfAIwAGOwEpHvQcJGnCAWlZ4cIhBhpVzdBAmY5PGkAJgBWWNU4hL4k1QwsaC1SCioaIwkYU3NHPjtlENC+cMiY+MS6lIb0zUJs3TzjQrN0Evs7AFIKvhcoAG4QZRAIAAcYEWh0ZFBCNgB3AAVCBEWUPCFVvCJFybkCJTgYAGU8AFs4I1NkADM9zBOzpgvL6YYPZmQYAR2HBJnAbnJ9Dh9AY8FBmOw8Mw4AAxCAEG54GBzOHIEB4dgwCATECSG5CADq4hE8HQPzAcEu22pInI1KIuLA5mJHkw1HWZ2YGOerxBIAAVugAB6XP5COAARXYEHgwqEbxAPwIvNxcq5k2mBA8MHJIhwMHEyDKGrYmHJZ2muINZiolGJAEclfB+TMdnj0F5TPp9MSCHAPSJQ-zEUKkC81aLMDcRACgQnZXAAIJYw1yAlwdZUB5wVXq9DpxXK4uxkWTGB4OQms0WpDRWtnESiOEAYQgNxjIDMsWJ7EwABV6zs4+ryMCyFDYJcwIbZhncJcYEQ5SW4ABfHdAA
image

If there only would be a possiblity to have comment hints to ignore some lines which are already perfect (for us, not what prettiert considers perfect)

mvorisek: 👍 for the finding, but as you saw, this wraps the selectors too...

can you propose a prettier patch against my branch?

mvorisek: I spent several hours in this. The best place to patch is https://github.com/prettier/prettier/blob/2.8.1/src/language-css/printer-postcss.js#L840

where something like:
```
        if (parentParentNode.type === "value-root") {
          parts.push(hardline);
          didBreak = true;

          continue;
        }
```

can be added, then all values separated by spaces are placed on new lines like:

image

but math operations (not inside parentheses) are too:

image

and imports:

image

the imports can be fixed with ` && declAncestorProp` condition but then variable assignments are unfixed (as for ex. `margin` can be assigned to any property name, so we are short on data)

I also tried ` && iNextNode.raws.before.includes("\n")` condition to check if there was a new line originally, but there are only like 15 places where applicable (or not possible to patch on one place) as other places/AST types look like:

image

but once we start to fix function argumnets, should be wrap simple/short `rgb` arguments to? no, we should not

so I would say we cannot do much and I would prefer to get this PR merged with maxWidth=Infinity which fixes some unlogical wrappings

actually, if we look on https://github.com/fomantic/Fomantic-UI/commit/470cd4e7b1dbce900d5e006aa8f3c5d7bf850009, there are not so many places "that would be better formatted differently", so I guess this is acceptable vs. the gain with Prettier integrated

lubber-de: I am honest with you here and i _really_ appreciate all the time you spent into this:
In the current stage, i don't want to merge the PR. It is not satisfying as a whole and that manual patching is too hacky. That's not the way we should implement/use an external package.

Let's see if/when your proposed PRs in the prettier repo get merged and what could be patched as well to get that "linear-gradient()/functions-parameter-each-line (but not rgb as you said)"-formatting could be improved.

A general possibility for comment-hints like in stylelint or eslint would solve it as we simply could define where prettier should not touch the existing code (instead of figuring out some complex "always to use" formatting syntax)

The current formatted code is IMHO much better or "prettier enough" than what prettier makes out of it.

And, in the end, all it does is doing/expecting whitespace (which you already did in the previous PRs) which gives much less benefit to the CI than what you did with eslint/stylelint already.

However, many changes in this PR i would actually agree with and some of them actually fixes errors (like missing or misplaced semicolons), but for many others i am not sure if really like the change (like the list items per line in the variation.variables file) while the "put everything in one line" is either not consistent all the time and is not really "prettier" than before. (yes, that's of course opinion based 😉 but i would like to concentrate again on improving/fixing FUI components instead of discussing whitespace rules.

mvorisek: there are 4 patches currently - 3 regular Prettier bugfixes, 1 patch to adjust the CS as much as possible to the current CS, the first 3 bugfixes were already approved and should be in the next Prettier release

I agree we should focus on FUI other than whitespaces, but I would prefer this topic to be finished, otherwise we will need to focus again, and be in the same exact point

the original review comments are for maxWidth=120, can you remove them, and add comments for current maxWidth=Infinity, ideally only once per CS problem, maybe with number of importance/occurrences, and how you propose it to be fixed by AST

I can then try to fix the problems with rules based on the property/variable name, something that will never be accepted into Prettier, but we can patch it (and possibly create a fork for it)

src/definitions/modules/modal.less
```diff
@@ -33,7 +33,12 @@
     flex: 0 0 auto;
     border-radius: @borderRadius;
     user-select: text;
-    will-change: top, left, margin, transform, opacity;
+    will-change:
```
mvorisek, line 36: this this patching based on comma present ok?

lubber-de: > this this patching based on comma present ok?

It's ok, although the simple liste for will-change would be sufficient in one line. But if your prefer it here, i would agree.

> is semicolon on the same like ok? (in other cases, it cannot be patched easily, also, when Prettier decised to wrap (usually based on maxLength), it also does not add a line break before the semicolon)

Moving the semicolon to the end is ok, although having it in a separate line as it is now more visualises the related block. But we had the same discussion for the JS files (var lists) as well 😉 and the argument of "i still have to change the previous line by adding a comma when i want to add another parameter" would be consistent and i agreed there to have the semicolon in the same line, so i would agree here too, if you think its better 🙂

src/themes/default/elements/segment.variables
lubber-de: > the original review comments are for maxWidth=120, can you remove them, and add comments for current maxWidth=Infinity, ideally only once per CS problem

Yes, but i was not aware of the consequences as i did not use prettier before. (And when i look at that many issues/open PRs at prettiert,i feeld many people are not completely satisfied with it and forked it to their needs (mostly JS related, but anyway)

> I can then try to fix the problems with rules based on the property/variable name, something that will never be accepted into Prettier, but we can patch it (and possibly create a Prettier fork)

If you like, i can fork it to the fomantic organization and give you rights to maintain it. We eventually could then publish it to npm and fomantic-scope to have a "real" package to rely on in CI.

src/definitions/modules/modal.less
mvorisek: > semicolon placement

I want to stay as much as possible with the Prettier CS which is broadly accepted.

In JS we place semicolon at the same line too (for ex. for ternary), in declaration, we should actually switch to multiple `let`/`const`, so then there will be no semicolon on a separate line

mvorisek:
@lubber-de please see the reorganized commits which discuss all changes done 1 by 1.

basically only the 2 last commits are the commits we are worried about:
- https://github.com/fomantic/Fomantic-UI/pull/2611/commits/24e5e29a708cc8593336373131f6eed6eea8dceb `if` on a separate line would be nice, but I would say we can easily say the commit is ok (and also the conditional IE font will be removed sooner or later)
- https://github.com/fomantic/Fomantic-UI/pull/2611/commits/39803afa77e7b14bf89adfdecc9a1cb9b78a5b90 6 not really nice fixes in sense of readability, but should we really be worried about?

Based on this analysis, I would safe we should merge this PR as is.

lubber-de:
> * [24e5e29](https://github.com/fomantic/Fomantic-UI/commit/24e5e29a708cc8593336373131f6eed6eea8dceb) `if` on a separate line would be nice, but I would say we can easily say the commit is ok (and also the conditional IE font will be removed sooner or later)

Ok, agree

> * [39803af](https://github.com/fomantic/Fomantic-UI/commit/39803afa77e7b14bf89adfdecc9a1cb9b78a5b90) 6 not really nice fixes in sense of readability, but should we really be worried about?

Yes, this is still ugly and not "pretty". If we just could ignore lines (or blocks), this would solve the issue (and others too like the above if statement or lists in variation.variables (even if i do not totally disagree with that))
That said, what about using `/* prettier-ignore */` ? Isnt that working?
https://prettier.io/docs/en/ignore.html#css

> Based on this analysis, I would safe we should merge this PR as is.

I still dont like the git patching hack in the CI. If your PRs (and possibly more needed fixes) are not implemented in the original package and basically does not do what we want, we should fork it if we really feel the need to still use prettier. Many other people went the same way and thats the idea of using a package manager.

mvorisek:
> > * [39803af](https://github.com/fomantic/Fomantic-UI/commit/39803afa77e7b14bf89adfdecc9a1cb9b78a5b90) 6 not really nice fixes in sense of readability, but should we really be worried about?
> 
> Yes, this is still ugly and not "pretty". If we just could ignore lines (or blocks), this would solve the issue (and others too like the above if statement or lists in variation.variables (even if i do not totally disagree with that)) That said, what about using `/* prettier-ignore */` ? Isnt that working? https://prettier.io/docs/en/ignore.html#css

I have to admit I misread https://stackoverflow.com/questions/66259151/how-to-make-prettier-to-ignore-a-block-of-code, `prettier-ignore` for the next AST node is supported event outside Markdown, so yes, local problems can be ignored, fixed in https://github.com/fomantic/Fomantic-UI/pull/2611/commits/2a4c08deddcefe9cf7b4630fa24beba529883802

> > Based on this analysis, I would safe we should merge this PR as is.
> 
> I still dont like the git patching hack in the CI. If your PRs (and possibly more needed fixes) are not implemented in the original package and basically does not do what we want, we should fork it if we really feel the need to still use prettier. Many other people went the same way and thats the idea of using a package manager.

I am not against, fork has one advantage - user can run prettier locally easier. So please fork the Prettier under fomantic GH org and apply there all 4 patches.

src/themes/default/globals/variation.variables
```diff
@@ -3,10 +3,42 @@
 ***********************************************************/
 
 /* General */
-@variationAllSizes: mini, tiny, small, large, big, huge, massive;
```
lubber-de, line 6: i would prefer having one line per each setting in this file, please add  `// prettier-ignore` fpr all list variables in the general block. It gets a quicker overview what switches are available and i also fear people could have some CI which replaces single lines

src/themes/default/globals/variation.variables
```diff
@@ -81,7 +115,13 @@
 @variationHeaderDividing: true;
 @variationHeaderBlock: true;
 @variationHeaderAttached: true;
-@variationHeaderTags: h1, h2, h3, h4, h5, h6;
+@variationHeaderTags:
```
lubber-de, line 118: same here, pleas add `// prettier-ignore` instead and leave it in one line

src/themes/default/globals/variation.variables
```diff
@@ -38,7 +70,9 @@
 @variationButtonWrapping: true;
 @variationButtonWrapped: true;
 @variationButtonSpaced: true;
-@variationButtonConsequences: positive, negative;
+@variationButtonConsequences:
```
lubber-de, line 73: same here, pleas add `// prettier-ignore` instead and leave it in one line

src/themes/default/globals/variation.variables
```diff
@@ -682,7 +722,10 @@
 @variationSliderTicked: true;
 @variationSliderVertical: true;
 @variationSliderBasic: true;
-@variationSliderSizes: small, large, big;
+@variationSliderSizes:
```
lubber-de, line 725: same here, pleas add `// prettier-ignore` instead and leave it in one line

src/themes/default/globals/site.variables
```diff
@@ -14,22 +14,168 @@
...
-@unicodeRangeSupplement: U+0180-0191, U+0193-0217, U+021C-0236, U+0238-02C5, U+02C8, U+02CA-02D7, U+02DE-036F, U+0374-0375, U+037A-037E, U+0384-038A, U+038C, U+038E-0393, U+0395-03A1, U+03A3-03A8, U+03AA-03BB, U+03BD-03BF, U+03C1-03CE, U+03D0-0486, U+0488-0513, U+1D00-1DCA, U+1DFE-1E7F, U+1E86-1E9B, U+1E9E, U+1EA0-1EF9, U+1F00-1F15, U+1F18-1F1D, U+1F20-1F45, U+1F48-1F4D, U+1F50-1F57, U+1F59, U+1F5B, U+1F5D, U+1F5F-1F7D, U+1F80-1FB4, U+1FB6-1FC4, U+1FC6-1FD3, U+1FD6-1FDB, U+1FDD-1FEF, U+1FF2-1FF4, U+1FF6-1FFE, U+2000-200F, U+2012, U+2015-2017, U+201B, U+201F, U+202F, U+2034, U+203C-203E, U+205E-205F, U+2070-2071, U+2074-2094, U+20A0-20A2, U+20A5-20A6, U+20A8-20AB, U+20AD-20B5, U+20B8-20BA, U+20DD, U+2105, U+2116-2117, U+2120, U+2132, U+214D-214E, U+2153-215F, U+2183-2184, U+2190-2199, U+21A8, U+221F, U+2229, U+2261, U+2302, U+2310, U+2320-2321, U+2460-2473, U+24EA-24F4, U+24FF-2500, U+2502, U+250C, U+2510, U+2514, U+2518, U+2C60-2C6C, U+2C74-2C77;
+@unicodeRangeLatin:
```
lubber-de, line 44: This is unnecessarily bloating the file and i doubt anybody would like to remove single unicode ranges
so , please add `// prettier-ignore` instead and leave it in one line

lubber-de:
> So please fork the Prettier under fomantic GH org and apply there all 4 patches.

I sent you an invite for Admin access to https://github.com/fomantic/prettier, so you can freely maintain it 🙂

.github/workflows/ci.yml
```diff
@@ -11,13 +11,7 @@ jobs:
...
     - uses: actions/setup-node@v3
     - name: Install dependencies
-      run: >
-        yarn
-        && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript github:prettier/prettier#2.8.1
-        && (cd node_modules/prettier && git apply -v ../../.github/fix-prettier-14007-no-unstable-semicolon.patch)
-        && (cd node_modules/prettier && git apply -v ../../.github/fix-prettier-14008-no-space-after-unary-minus.patch)
-        && (cd node_modules/prettier && git apply -v ../../.github/fix-prettier-14034-no-property-name-lower-if-inside-variable-declaration.patch)
-        && (cd node_modules/prettier && git apply -v ../../.github/fix-prettier-value-with-comma.patch)
+      run: yarn && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript github:fomantic/prettier#ca4fc1130f
```
lubber-de, line 14: I suggest to create a version tag instead of using the commit hash

.github/workflows/ci.yml
```diff
@@ -11,9 +11,13 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-node@v3
     - name: Install dependencies
-      run: yarn && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript
+      run: yarn && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript github:fomantic/prettier#ca4fc1130f
     - name: Lint JS and LESS files
-      run: npm run lint
+      run: >
+        npm run lint
```
lubber-de, line 17: I suggest to move the prettier check into a separate task instead of being part of the linter

.github/workflows/ci.yml
mvorisek: why?

src/themes/default/globals/site.variables
mvorisek: see https://github.com/fomantic/prettier/commit/ca814b79044221606bcc8114bf9d781409201202

I was able to patch Prettier to honor line breaks in the original data, so you can now choose if you want to have comma separated values on one line or not

.github/workflows/ci.yml
mvorisek: done

all feedback should be addressed now

mvorisek: done

all feedback should be addressed now

.github/workflows/ci.yml
lubber-de: Because i want something being "finished" and a version tag declares that (yes, technically it's nothing else than a pointer to a specific commit).
I created a github release out of your tag

mvorisek: why GH release vs. not pure git tag?

lubber-de: For the same reason 😉

lubber-de: And its made out of your tag anyway

.github/workflows/ci.yml
```diff
@@ -14,6 +14,12 @@ jobs:
       run: yarn && yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript
     - name: Lint JS and LESS files
       run: npm run lint
+    - name: Assert LESS files formatting using Prettier
+      run: >
+        yarn add -D eslint-config-airbnb-base eslint-plugin-unicorn eslint-plugin-jest@^26.9.0 @typescript-eslint/eslint-plugin @typescript-eslint/parser typescript github:fomantic/prettier#2.8.1-patched
```
lubber-de, line 19: The eslint packages have already been added by the linter before (and are only need for the linter anyway arent they?) So isnt 
```suggestion
        yarn add -D github:fomantic/prettier#2.8.1-patched
```
sufficient here?

lubber-de: And, i basically thought about having prettier a complete separate task (like lint or test or codeql).

.github/workflows/ci.yml
mvorisek: I have intentionally (originally) disabled releases as in my PoV it is only an additional release step not needed for anything. And the more many manual steps we need to do per release, the more mistakes we can introduce.

As long as it is an internal repo I do not see any advantage than having `main` branch HEAD as the latest internal stable version... But I do not mind.

.github/workflows/ci.yml
mvorisek: Both eslint and stylelint are functional and formatting linters. Why should additional formatter be separated?

Later, once there is less patching and we do not require to support Node 12, I even prefer all these to be fixable with one command. So separate step does not make any sense to me.

.github/workflows/ci.yml
lubber-de: Lets do it as clean as possible and not getting sloppy about it (i understand you want to get this finished now). In case of a new version you would have to change the ci to the new commit anyway.
And i think the minimum of maintenance is doing a proper version history. (and the main head might gets corrupted by untested commits if not protected/checked itself (thats why i asked about the linter error in the last commit) Referring to a specific tag/release is just as using a package manager

And who know who might gets interested in this fork because you fixed sutff, others also stumpled about 😉

.github/workflows/ci.yml
lubber-de: ok, i just remember yourself saying "prettier is not a linter" 😉

lubber-de: > The eslint packages have already been added by the linter before (and are only needed for the linter anyway arent they?) So isnt
> 
> sufficient here?

What about this then?

mvorisek: of course, this was not intended, 👍

lubber-de (approved):
Finally LGTM 👍🏼 

I would like to thank you very much for this and especially for your patience and understanding for all the (critical) feedback i gave 😉

lubber-de:
@mvorisek It seems prettier has merged all of your PRs 🙂 
They havent published a new version yet, but once they do, is everything covered what we need so we can switch back to the original prettier repo and remove the fork again?

mvorisek:
@lubber-de all except the one which adds new lines when comma is present - https://github.com/fomantic/prettier/commit/d9c5a044aeba1ef2559cde61ed0652abdb220287 and https://github.com/fomantic/prettier/commit/ca814b79044221606bcc8114bf9d781409201202

I can try to propose it to Prettier as well, but I am not sure if it will be accepted

lubber-de:
> I can try to propose it to Prettier as well, but I am not sure if it will be accepted

Yes, please 😊 they seem to trust you

mvorisek:
I have about 1500 contributions / year in GH. But it is not only about personal credit, the previous PRs were bugfixes, this one is feature request about styling. AFAK such FR should not be accepted, I will do my best to stress the importance of increased readability with a an example, but I cannot promise anything.

mvorisek:
@lubber-de I have crafter 14208 prettier PR. Currently we enforce NL based on the original presence - https://github.com/fomantic/prettier/commit/edb343f87995dfd5f735b7bf29dc310b60e5db5c - but such behaviour is not possible in prettier offcially (it rebuilds AST into one stable format).

So the 14208 prettier PR is designed to enforce line break in all CSS values when comma is present. See https://github.com/mvorisek/prettier/commit/da28053ff997aa259ee41046a7377bcfc35a7ee9 . Is such format ok with you or do you have any idea of better prettier formatting?

PR #2612, opened 2022-12-15
feat(lint): adjust LESS whitespaces to match prettier format

mvorisek:
This PR fixes whitespaces unfixed/undetected in GH-2610 and major whitespace changes required for future Prettier integration.

Double quotes and lowecased hex colors are required by Prettier - https://github.com/prettier/prettier/issues/5158 - and there is no config to change it. As long as we want to use Prettier, we must accept opinionated formatting by Prettier.


src/semantic.less
```diff
@@ -13,63 +13,167 @@
 // stylelint-disable no-invalid-position-at-import-rule
 
 /* Global */
-& { @import 'definitions/globals/reset'; }
-& { @import 'definitions/globals/site'; }
+& {
```
lubber-de, line 16: Can we ignore the forced extra lines for semantic.less? The idea behind this file is to integrate into an existing less project and having each component in a single line would make it easiert to quickly remove/disable a component.
"Possibly" people could have some own CI which is adjusting this fileand are expecting single lines and would break then.

src/themes/default/modules/accordion.overrides
```diff
@@ -22,5 +22,5 @@
 .ui.accordion .ui.header > .dropdown.icon::before,
 .ui.accordion .title > .dropdown.icon::before,
 .ui.accordion .accordion .title > .dropdown.icon::before {
-    content: '\f0da'/*rtl:'\f0d9'*/;
```
lubber-de, line 25: Dont remove the rtl hint

src/themes/default/modules/dropdown.overrides
```diff
@@ -25,35 +25,35 @@
     width: auto;
 }
 .ui.dropdown > .dropdown.icon::before {
-    content: '\f0d7';
+    content: "\f0d7";
 }
 
 /* Sub Menu */
 .ui.dropdown .menu .item .dropdown.icon::before {
-    content: '\f0da'/*rtl:'\f0d9'*/;
```
lubber-de, line 33: Don't remove the rtl hint

src/semantic.less
mvorisek:
```diff
 & {
-    @import(...);
+//  @import(...);
 }
```

this should be fine

lubber-de: Yes sure, but we actually dont know what others are doing so it might be still a braking change.
_Please_ disable the rule for this file. Thanks.

src/themes/default/modules/accordion.overrides
mvorisek: Sorry about this again, I was expecting a spae added before the comment, but not complete removal. Checking what tool and why it was removed.

mvorisek: the comments were removed wrongly by Stylelint `quotes` rule - https://github.com/fomantic/Fomantic-UI/commit/267ddf4af1b63279e11ecac7e6a2ec60788a5682

fixed in https://github.com/fomantic/Fomantic-UI/pull/2612/commits/d68de62a02103d2dc3f752e2f8d83048b232da9b - are the added spaces ok?

src/themes/default/modules/dropdown.overrides
mvorisek: fixed, related discussion https://github.com/fomantic/Fomantic-UI/pull/2612#discussion_r1050731482

src/themes/github/modules/dropdown.overrides
```diff
@@ -30,20 +30,20 @@
     width: auto;
 }
 .ui.dropdown > .dropdown.icon::before {
-    content: '\f0d7';
+    content: "\f0d7";
 }
 
 /* Sub Menu */
 .ui.dropdown .menu .item .dropdown.icon::before {
-    content: '\f0da'/*rtl:'\f0d9'*/;
```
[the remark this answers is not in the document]
mvorisek: fixed, related discussion https://github.com/fomantic/Fomantic-UI/pull/2612#discussion_r1050731482

src/semantic.less
mvorisek: here fixed and added https://github.com/fomantic/Fomantic-UI/pull/2611/files#r1050733837 comment to the Prettier PR to make sure it stays as wanted

.github/workflows/ci.yml
```diff
@@ -33,3 +33,20 @@ jobs:
       run: yarn
     - name: pre fomantic install & gulp build
       run: yarn gulp install
+  codeql:
```
mvorisek, line 36: moved into `ci.yml` to emit only 1 GH Action workflow instead of 2 separate:

image

.github/workflows/ci.yml
```diff
@@ -33,3 +33,20 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: github/codeql-action/init@v2
+      with:
+        languages: 'javascript'
+        config-file: ./.github/codeql/codeql-config.yml
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v2
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2
```
lubber-de, line 52: The settings from original ci is missing:
```yml
fail-fast: false
      matrix:
        language: [ 'javascript' ]
```

.github/workflows/codeql.yml
```diff
@@ -1,68 +0,0 @@
-# For most projects, this workflow file will not need changing; you simply need
```
lubber-de, line 1: Please do not again put additional non PR related changes ...
Why was it necessary to put this into the main CI?

mvorisek: see https://github.com/fomantic/Fomantic-UI/pull/2612#discussion_r1050773751

mvorisek:
To verify this PR I run build before all changes and after. Then I run Prettier on both data (Prettier seems quite safe in sense of corrupting AST or comments).

There are only a few changes in `semantic.css.min` like:

image

image

image

image

will selectors with added space before `::` match always the same elements?

.github/workflows/ci.yml
lubber-de: but fail-fast is misssing...

mvorisek: without a matrix, for one job only, it is not needed

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast and it is a per (matrixed) job config, not per worflow config

tested in https://github.com/fomantic/Fomantic-UI/pull/2611/commits/19a07d17df713013f39711b6d9e8a359fd0fa50c - `Lint` job (also /wo `fail-fast: false`) failed, but the `CodeQL` continued to run (ie. was not cancelled, even if it took much longer)

lubber-de:
> To verify this PR I run build before all changes and after. Then I run Prettier on both data (Prettier seems quite safe in sense of corrupting AST or comments).
> 
> will selectors with removed space before `::` match always the same elements?

No, those would have a different meaning. the space indicates a child node, and without the space it indicates the exact node
I wonder, where that space is comiong

Damn, this is something that has been removed by the whitespace changes, which i did not recognize.... This has to be reverted to the state is was in 2.9.0

lubber-de:
The wrong removement was done by fc431e8c06ca1c0514d6fb137a29b5b24063e831

mvorisek:
I marked this PR as a draft and will take care of it.

.github/workflows/ci.yml
lubber-de: yes, because there is fail-fast in the test block, but if it would fail inside codeql, i want the whole codeql to continue checking for all possible errors. I beieven without fail-fast (inside the codeql block) it will stop parsing after finding the first error, isnt it?

mvorisek: I would say no, as in `lint` job there is no `fail-fast: false`. In `test` `fail-fast: false` is present, but that should be not relevant, or is it?

lubber-de: 🤔 Seems you are right according to the gh docs you posted above (fail-fast only affects multiple matrix entries). Alright, lets keep it then as you did 😄

mvorisek:
The change of removed space before pseudo classes is fixed/reverted.

lubber-de (approved):
Thank you, LGTM

lubber-de:
@mvorisek 
When i run the lint locally, stylelint moans about the "each" statement... but the CI report does not do this.
Any idea what i am doing differently (yes, i ran the additional yarn packages from the ci as well to get the additional packages and i am running node 18)
image

mvorisek:
Did you installed all the packages that CI installs? It seems like the CSS is parsed without the less preprocessor. Maybe some config issue, you call the command from different pwd...

PR #2615, opened 2022-12-18
feat(lint): remove unneeded vendor prefixes

mvorisek:
Some CSS properties like `transform` - https://caniuse.com/transforms3d - has wide enough support without vendor prefix.

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -641,6 +641,9 @@
         font-weight: @definitionHeaderFontWeight;
         color: @definitionHeaderColor;
         box-shadow: -@coloredBorderSizeCover -@coloredBorderSize 0 @coloredBorderSizeCover @definitionPageBackground;
+        // https://github.com/fomantic/Fomantic-UI/pull/884
+        // stylelint-disable-next-line property-no-vendor-prefix
+        -moz-transform: scale(1);
     }
 
     .ui.definition.table > thead:not(.full-width) > tr > th:first-child:not(:empty) {
@@ -653,6 +656,9 @@
         font-weight: @definitionFooterFontWeight;
         color: @definitionFooterColor;
         box-shadow: -@coloredBorderSizeCover @coloredBorderSize 0 @coloredBorderSizeCover @definitionPageBackground;
+        // https://github.com/fomantic/Fomantic-UI/pull/884
+        // stylelint-disable-next-line property-no-vendor-prefix
+        -moz-transform: scale(1);
     }
 
     /* Highlight Defining Column */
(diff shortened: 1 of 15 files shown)


src/definitions/modules/dimmer.less
```diff
@@ -302,7 +302,6 @@ body.dimmable > .dimmer {
     .ui[class*="center dimmer"] {
         top: 50%;
         transform: translateY(-50%);
-        -webkit-transform: translateY(calc(-50% - 0.5px));
```
mvorisek, line 305: https://github.com/fomantic/Fomantic-UI/pull/186/commits/678695ed447e630a70934ee1b3df5c4fe91a2109

is this hack still needed?

src/definitions/collections/table.less
```diff
@@ -641,7 +641,6 @@
         font-weight: @definitionHeaderFontWeight;
         color: @definitionHeaderColor;
         box-shadow: -@coloredBorderSizeCover -@coloredBorderSize 0 @coloredBorderSizeCover @definitionPageBackground;
-        -moz-transform: scale(1);
```
mvorisek, line 644: introduced in https://github.com/fomantic/Fomantic-UI/pull/884

I tried the fiddle in FF and the fix seems to be not needed anymore, but maybe I do not know all the conditions to reproduce...

src/definitions/modules/dropdown.less
```diff
@@ -434,7 +434,6 @@ select.ui.dropdown {
         overflow-y: auto;
         overscroll-behavior: @overscrollBehavior;
         backface-visibility: hidden;
-        -webkit-overflow-scrolling: touch;
```
mvorisek, line 437: https://caniuse.com/mdn-css_properties_-webkit-overflow-scrolling

supported in Safari until 2019, then no support in any browser

src/definitions/modules/dimmer.less
lubber-de: Doesnt seem so after a quick test

src/definitions/modules/dropdown.less
lubber-de: Since 2.9.0 we stated on the readme to only support the last 4 versions of ios Safari..That is 13-16, so ok for me

src/definitions/collections/table.less
lubber-de: That fix is unfortunately still needed for latest Firefox, so please revert the remove.
See https://jsfiddle.net/7e0yf5pu/ using latest 2.9.0 and try to remove the attribute via web insepector

ffdeftablefixstillneeded

mvorisek: reverted, in this demo, I was able to reproduce in my FF too 👍

lubber-de (approved):
LGTM

PR #2619, opened 2022-12-20
fix(chore): remove unnecessary methods and prefixes

lubber-de:
## Description
Since 2.9.0 we limited the browser support, so many browser methods are available everywhere. There is no need to check for them beforehand so i removed those.


src/definitions/modules/modal.js
```diff
@@ -35,13 +35,7 @@
...
             queryArguments = [].slice.call(arguments, 1),
 
-            requestAnimationFrame = window.requestAnimationFrame
-                || window.mozRequestAnimationFrame
-                || window.webkitRequestAnimationFrame
-                || window.msRequestAnimationFrame
-                || function (callback) {
-                    setTimeout(callback, 0);
-                },
+            requestAnimationFrame = window.requestAnimationFrame,
```
mvorisek, line 38: can be used directly

lubber-de: Yes, sure. I removed the declaration entirely now by e4e0f8226aa78e3ae185cebe7c4a09b6f4e223f6

mvorisek: but access still needs to be done using `window.`, otherwise more eslint undef errors

lubber-de: but the linter CI does not trigger an error?

lubber-de: eslintrc.js already got
```
    env: {
        browser: true,
``` so eslint knows about the global functions

mvorisek: I do not think so, the under rule is disabled.

PR #2621, opened 2022-12-21
feat(build): adjust autoprefixer config to match 2.9.0 browser support

lubber-de:
## Description
The default browserlist settings were still supporting old legacy browsers which we reduced in 2.9.0.
This PR adjust the browserlist setting so much less prefixes (if any) are added to the css files.
This PR also allows to configure a whole autoprefixer prefix-config instead of only the overrideBrowserslist attribute.


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/417

PR #2633, opened 2022-12-28
fix(form): dynamic validation should't check deleted fields

### LINKED ISSUE #2577: [Form] Dynamically Removed Fields Still Throws Error in Console
# Bug Report
Removing form field dynamically still getting error ``"Field identifier `field name` not found"``!!

## Steps to reproduce
https://jsfiddle.net/KiddoV/0wvgbe4q/35/
1. Open Fiddle!
2. Hit Remove Field button!
3. Hit Submit button!

## Expected result
Expecting no ERROR in console!!!

## Version
2.9.0

prudho:
## Description
When a field was dynamically removed after form initialization, validation still tried to check the removed field, resulting in an error.

This PR adds a check for still existing fields, and remove validation rules for those which doesn't exist anymore.

## Testcase
1. Open Fiddle!
2. Hit Remove Field button!
3. Hit Submit button!

Before: [JSFiddle](https://jsfiddle.net/KiddoV/0wvgbe4q/35/)

After: [JSFiddle](https://jsfiddle.net/KiddoV/0amr1c5b)

## Closes
#2577


src/definitions/behaviors/form.js
```diff
@@ -951,6 +951,7 @@
                         if (rule === undefined) {
                             module.debug('Removed all rules');
                             validation[field].rules = [];
+                            delete validation[field];
```
lubber-de, line 954: This breaks into a console error in case one wants to remove rules and validate the field afterwards, because the `validate.field` method expects a field to be existent inside the validation array (even without any rule)
See https://jsfiddle.net/lubber/m51yfaek/7/ and click the purple button

image

prudho: This error is already triggered without my modifications. In this case I only remove the rules if the field is removed. But you're right I should only delete the validation array key if the field doesn't exist anymore.

lubber-de: You could just fix this inside the validate.field function to fetch any non existing validation array access.
And if you leave the "delete", you could also remove the previous `validation[field].rules = [];` because that wont be necessary anymore if you delete the whole field key from validatation array anyway

prudho: Actually your fiddle seems broken since you try to validate a fied that doesn't exist: `field2` instead of `field-2`.

`validate.field` should indeed check if a field exists before trying to validate it.

lubber-de: > Actually your fiddle seems broken since you try to validate a fied that doesn't exist: `field2` instead of `field-2`.
> 
> `validate.field` should indeed check if a field exists before trying to validate it.

Oh, that's indeed right!  However, i still dont understand from the code, why "field.identifier" does not break, although validation[field] was previously deleted and should be undefined .. I investigate

prudho:
I've pushed some additional fixes and tweaked the code a little. This PR can now be merged since the raised issue is not related to this.

lubber-de:
> I've pushed some additional fixes and tweaked the code a little. This PR can now be merged since the raised issue is not related to this.

I am still not sure if the deletion of `validation[field]` is correct. Because validation[field] could also host `identifier` which will be completely lost as well then. If one would like to remove the rules, then add rules again, the previous identifier cannot be found anymore inside the validation object. `add.field` does only add
```
validation[name] = {
    rules: [],
};
```
in case a field is not present inside the validation-object, but omits a possibly presiously provided  identifier key. This could potentionally be a breaking change then. I'll test a bit more

lubber-de:
@prudho  Ok, for get about my "identifier" concern....this is fetched as the identifier is the key ... sorry for that.

However, the code still breaks on `remove rules` and `validate field` afterwards
See https://jsfiddle.net/lubber/m51yfaek/14/
because of the deletion of the field key from validation and validate.field expects it, so we should fetch it as suggested here https://github.com/fomantic/Fomantic-UI/pull/2633#discussion_r1058362656

prudho:
I've update the code to don't remove the field key if the element is still there.

Regarding the `validate.field()` trigerred error, it should silently fail but display an error when `debug` is set to `true`. What to you think about it ?

lubber-de:
> I've update the code to don't remove the field key if the element is still there.

Great! Works now 🙂 👍🏼 

> 
> Regarding the `validate.field()` trigerred error, it should silently fail but display an error when `debug` is set to `true`. What to you think about it ?

Yes, good idea

prudho:
Added. Should be all good now 😃

lubber-de (approved):
LGTM 👍🏼

PR #2637, opened 2023-01-05
feat: typescript transition

prudho:
## Preamble
Following my attempt as providing types for fomantic, I realized that it's possible to transition `js` files to `ts readable` files, opening way for a soft transition to a future fomantic without breaking the whole stuff. So I made some tries, and here is a proof of concept of what can be achieved with minimal work.

## Description
This PR implements the typescript build system for Fomantic, some type definitions and some modules already transitionned.

### Dependencies
2 dependencies has been added to handle typescript: `typescript` and `gulp-typescript`. The first is the typescript compiler(transpiler?) itself, the second allow our build system to use the first one with our gulp pipeline.

2 dev dependencies: `@types/node` and `@types/jquery` are used by the typescript compiler, and also very usefull for code completion.

### The build system
Our build system has been modified to the minimum. I made a task called `buildTS` which is a nearly perfect copy of the `buildJS`. It only watches for ts files and add them in the gulp pipeline, which compile the file.

Since most of the js modules only had been renamed for now, building the whole project triggers a lot of TS errors for now, but thoses errors are not breaking the output, files are still generated and added to the `dist` folder.

### Type definitions
Type definitions have two purposes here:
- provide definitions of modules for users, allowing them to better use the library and speed-up their developments with.
- providing the typescript compiler a "signature" of modules.

They're stored in the new `types` folder at project root. For now only 3 modules are covered in this project, but ~95% of the others modules is already done (see https://github.com/fomantic/DefinitelyTyped/tree/fomantic-ui/types/fomantic-ui) and probably needs only a refresh.

### The modules
Most of files are for now only been renamed for now. `rating` and `slider` does'nt trigger any TS error and `flyout` a few ones. The main final goal is to make all thoses errors disapear.

## What now ?
To start this transition ? Well, finish the minimal conversion of each js module. Make some tests.

On the long term, there's a lot more to do:
- Find the good configuration for typescript and include it in fomantic configuration.
- Increase typed languages surface, step by step.


lubber-de:
I am honest here (before you probably spend much more time into this):

#### Why dont we just use jsdoc blocks at the related code in the source files instead? (used in the gulp scripts already)

jsdocs adds type checking in every IDE since ages and we could generate Docs out of them as well.
This way it stays plain JS which we can still use/test/prototype in the browser as before without any (extra needed) buildstep

Using typescript just for the sake of it (by only renaming the js files and the need to code (and maintain) extra d.ts files) does not gain any real benefit for my taste.
Even worse, it slows down development, as i cannot simply include the src modules js file in the browser anymore, i _always_ have to build it first and i also have to maintain the extra types files.

Otherwise, just renaming the js files to satisfy tsc, you are probably going to add a lot of `//@ts-ignore` lines to the existing code, just to compile without errors without really getting a benefit. With the current code, i really dont see the point as this adds no value. The current code (and JS itself) allows _by design_ to have separate types for some properties, that's probably why ts wont like most of the current code.

Once we really completely rewrite the whole modules code from scratch using ES6, no query, whatever,...  (who knows when/if that is happening)... we can recall the question if we want to use TS.
But for the current existing code, i am against converting it into typescript. Not even for 2.10.x

The "types" folder should be a seperate package instead (if you still want to maintain such), so i suggest to continue the original approach and reuse the definatelytyped fomantic package

mvorisek:
I am mainly a professional PHP developer, so I cannot tell if moving to TS now is best, but I can share my observations for this topic and Fomantic-U:
- static analysis is impossible /wo method types - yes, we need types (JsDoc/TS)
- very little contributors care about types, so when they do not have to define them in JsDoc, they will not define them
- types in code takes less LoC
- TS AST can be exported reliably (and much more easily vs JsDoc) and such info used in dependent projects (would help atk4/ui I maintain)
- local build must be easy and time to rebuild after a code change is critical to not loose focus - when whole Fomantic-UI is migrated, typed, how fast the TS can be recompiled?
- having to maintain interface it on two places (`interface` + `class`) is very time consuming, for example, method descriptions can drifts apart easily - is TS `interface` helpful for the transition?
- code copied into multiple places is hard to maintain as well - currently a lot of same/similar code is copied into different module scripts
- the same with browser compatibility, code can be simplified in favor of transpilation with `core-js`

> Using typescript just for the sake of it (by only renaming the js files and the need to code (and maintain) extra d.ts files) does not gain any real benefit for my taste.
Even worse, it slows down development, as i cannot simply include the src modules js file in the browser anymore, i always have to build it first and i also have to maintain the extra types files.

Several points I mentioned above needs compilation/transpilation, so in the future it is unavoidable. JS without recompilation should be debuggable with non-minified JS, which can be even split into vendor JS and per module.

...and when we compile code, TS should have no disadvantage :)

prudho:
Closing this. Sorry i tried too many hours to respond to this, and your arguments are valid too @lubber-de.

The final purpose of this was to make the project move towards future. We must face that the web is moving quick now, with modular code, that sometimes needs to be included trough frameworks (parcel, webpack, rollup...). And Fomantic is not ready. ES6 migration won't happen here, as well as getting rid of jQuery. I guess i'm frustated about this.

Types will be polished and released as planned to DefinitelyTyped. And i'll try to raise issues that were found during this little `typescript` parenthesis.

@mvorisek thanks for your feedback. I'm not really present into Fomantic right now but I saw what you've done here and I really appreciate.

PR #2647, opened 2023-01-12
feat(dropdown): dont animate labels when no transition is given

### LINKED ISSUE #2646: [DropDown] "set selected" not working as expected
# Bug Report
1. If I'm using "set selected" on dropdown builded (DOM) in JS with "multiple" option before attaching to the page, the values that was selected disappear ([reproduce](https://jsfiddle.net/mb1rghk3/12/))
2. If  I'm attaching to the page and then calling "set selected", it's working as expected  ([reproduce](https://jsfiddle.net/mb1rghk3/13/))
3. If  I'm removing "multiple" option, it's working as expected (even before attaching to the page) ([reproduce](https://jsfiddle.net/mb1rghk3/16/))

lubber-de:
## Description

This PR allows to disable animation transitions for labels of a multiple dropdown via
```javascript
$('.ui.dropdown').dropdown({
  label: {
  	transition: false
  }
});
```
This might be useful in situations where the dropdown element is reattached in the DOM while the animation still takes place (and was triggered via javascript for example by "set selected"

## Testcase
https://jsfiddle.net/lubber/pjw105ht/5/

## Closes
#2646

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2829,7 +2829,7 @@
                         if (settings.label.variation) {
                             $label.addClass(settings.label.variation);
                         }
-                        if (shouldAnimate === true) {
+                        if (shouldAnimate === true && settings.label.transition) {
                             module.debug('Animating in label', $label);
                             $label
                                 .addClass(className.hidden)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/420

PR #2651, opened 2023-01-13
fix(chore): unify invoke method, removed unused error strings

lubber-de:
## Description
Some modules missed the error message when a given behavior does not exist, although the error string was already present in the code.
On the other hand there were some unused error strings, which i removed.


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/423

PR #2656, opened 2023-01-15
feat(popup):  loading variant

lubber-de:
## Description
Added `set loading` behavior as `remove loading` already exists and used them (only) internally as well.
This is now also properly displayed with the usual loading spinner supporting slow/fast/double/colors

This might come in handy when someone want to lazy load popups manually

## Testcase
https://jsfiddle.net/lubber/8e6dbx3h/18/

## Screenshots
image


lubber-de:
I'll improve this for 2.9.2 as popup does not really provide a `loading` state. It's currently just hiding the popup while its still in transition visible state... This is not clean, so i'll postpone this to make it a proper loading state 🙂

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/426

PR #2674, opened 2023-01-23
fix(modal,flyout): avoid multiple autofocus on show animation

lubber-de:
## Description
jquery's focus trigger does not seem to work / detect a non-input element as visible all the time. Although the autofocus is called after the animation has ended a same modal opened a second time (but without input content) does not focus the modal div again, thus focus trap did not work.

This is caused by the new mutationobserver logic fetching class attributes, so the focussing takes place several times, while not necessary when the moda/flyout is animated.

This PR now only observes attributes when the modal/flyout has completed showing as the showing process itself changes class attributes.

## Closes
https://github.com/fomantic/Fomantic-UI/pull/2575#issuecomment-1397852452


src/definitions/modules/modal.js
```diff
@@ -1049,7 +1049,7 @@
                             $input = $inputs.first();
                         }
                         if ($input.length > 0) {
-                            $input.trigger('focus');
+                            $input[0].focus();
```
mvorisek, line 1052: I have tested this fix and it fixes https://github.com/fomantic/Fomantic-UI/pull/2575#issuecomment-1399945576

I tried to debug why jQuery focus is not working and noticed this line is called many times. Is that really needed? Maybe that is the source of the issue, as the line is called once even before close (when you click on the dimmer).

lubber-de: @mvorisek Thanks a lot for that hint! I reverted the change back to `.trigger('focus')` and disabked the attribute observing while the modal gets shown which now makes sure the autofocus does not trigger multiple times anymore which in turn fixes your issue without the need for the native focus() event

mvorisek: Any idea why the jQuery focus was not working? If the problem was frequency, frequent changes can be observed even after the modal is shown.

If I am not mistaken the modal is focused each time a change is observed (when there is no input already focus), but the https://github.com/fomantic/Fomantic-UI/pull/2575#issuecomment-1397852452 repro has no input, and main modal div is focused instead. Can the div focus be detected? If yes, just do nothing if it is in focus already.

lubber-de: > Any idea why the jQuery focus was not working? If the problem was frequency, frequent changes can be observed even after the modal is shown.

no idea (i tried to understand jquery logic but did not spent that much time into it tbh) 
So, i can of course still/again use the native .focus() in addition to the observerattributes logic to be 100% sure.. Should i change the code again? opinion?
The issue was only happening when the modal shown and only when the modal did not contain any input.

> 
> If I am not mistaken the modal is focused each time a change is observed (when there is no input already focus), but the [#2575 (comment)](https://github.com/fomantic/Fomantic-UI/pull/2575#issuecomment-1397852452) repro has no input, and main modal div is focused instead. Can the div focus be detected? If yes, just do nothing if it is in focus already.

I know only remove the tabindex from the div by [022381d](https://github.com/fomantic/Fomantic-UI/pull/2674/commits/022381d75163cbbc5b2db65d358816a0d7357741) when there is not input found inside, so in case it was focused before, it doesnt get refocused as `.filter(':focus')` fetches it already

Also fixed a stupid check for class changes (which _always_ triggered the refreshinput before)

mvorisek: > So, i can of course still/again use the native .focus() in addition to the observerattributes logic to be 100% sure.. Should i change the code again? opinion?

no, I never like changes without understanding the problem, better to keep jQuery and if the problem will become more problematic later, let's dig into it again

> I now only remove the tabindex from the div by [022381d](https://github.com/fomantic/Fomantic-UI/pull/2674/commits/022381d75163cbbc5b2db65d358816a0d7357741) when there is no input found inside, so in case it was focused before, it doesnt get refocused as .filter(':focus') fetches it already

is `observeAttributes = false` still needed? If not, I prefer to remove it to keep code simple as possible.

lubber-de: > is `observeAttributes = false` still needed? If not, I prefer to remove it to keep code simple as possible.

Yes, it is, because the attribute observing for classnames will otherwise trigger the inputrefresh many times while showing/hiding the modal (and if the modal contains inputs).

I thought of completely disconnecting the observer while showing/hiding but that probably breaks projects wich are handling the modal content via ajax and possibly dont trigger the focusing at all.. to risky, so i would leave the code as it is now.

ivangomes:
Somebody, please authorized this pull request.

lubber-de:
Can you please provide a jsfiddle (or adjust the on i posted below) ? I believe there is some additional javascript missing from your website, which i cannot figure out where to copy from.

Here is a jsfiddle where i copied your modal HTML , but unfortunately the issue does not happen  there and works as expected. If i enter anything inside the ISO fields and press tab it focuses the next field.

I believe some form validation is missing ?

https://jsfiddle.net/lubber/yn7qjuer/4/

I already working on fixing search/dropdown components, but in your case its a common text input..
In addition, in your original website, when i change the iso field a second time and press tab it also seems to work as expected.

lubber-de:
@mvorisek I believe your issue is fixed by #2708

mvorisek:
@lubber-de the described issue is the exact duplicate of #2700

PR #2680, opened 2023-01-27
feat(formvalidation): support object list in add errors

### LINKED ISSUE #2677: [Form] The "add errors" behavior does not work with inline errors
# Bug Report

The "inline" form setting is excellent. It works beautifully for form validation. However the `add errors` behavior only works without inline.

The jsfiddle below shows the issue. If you run as-is, it breaks. If you comment out `inline: true`, however, it works as expected.

## Steps to reproduce

1. Run the fiddle.
2. Enter a name, and click CONTINUE
4. Examine the errors in the console.
5. Comment out `inline: true`, re-run the code, and repeat 1-2. This time it works.

## Expected result
The error "That's a lovely name" appearing in the inline popup under the name.

## Actual result
An empty popup/arrow thingie and console errors.

## Testcase
https://jsfiddle.net/h8n4kr7a/1/

## Screenshot (if possible)
Screen Shot 2023-01-26 at 7 33 44 PM

## Version
2.9.0
(Haven't updated to 2.9.1 yet, but the jsfiddle uses 2.9.1, and the error still happens.)

lubber-de:
## Description

This PR fixes formvalidation and  supports
- named objects as `errors` for the `add errors` behavior. This will either create individual prompts when `inline:true` or provide them as before via central error message box. But will support to put the field name infront.
- also fetches falsy values for "errors" as well now
- tries to get the field label for `match` and `different` rules
- centralizes the "get value" which was used inside match/different rules twice and did not check for non standard inputs like dropdown, calender or radio
- reuse / centralize /simplfy field selector 
- given an non name identifier (such as id) to the "get value" behavior (was not working if id!=name)
- init a select-dropdown
- support native checkboxes

## Testcase
Add errors using an object
https://jsfiddle.net/lubber/vcxsd3gt/5/

Match rule returns field label instead of id
@Eng3l This fixes your issue in the SUI repo, where i am not allowed to post anymore
https://jsfiddle.net/lubber/sbd0fmz4/2/

|Before|After|
|-|-|
|image|image|

## Closes
#2677 
https://github.com/Semantic-Org/Semantic-UI/issues/6130

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -596,7 +596,7 @@
                         // refresh selector cache
                         (instance || module).refresh();
                     },
-                    field: function (identifier) {
+                    field: function (identifier, strict) {
                         module.verbose('Finding field with identifier', identifier);
                         identifier = module.escape.string(identifier);
                         var t;
@@ -618,7 +618,7 @@
                         }
                         module.error(error.noField.replace('{identifier}', identifier));
 
-                        return $('<input/>');
+                        return strict ? [] : $('<input/>');
                     },
                     fields: function (fields) {
                         var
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/430

PR #2683, opened 2023-01-30
feat(segment,grid): left and right attached variant

lubber-de:
## Description

This PR finally allows to have a `vertical tabular menu` flawlessly attached to a left or right segment inside a grid column.
To accomplish this the tabular menu has to stay inside a `left/right attached column` and a `left/right attached segment` has to stay in a `left/right attached column`

## Testcase
https://jsfiddle.net/lubber/L9go6tab/13/

### After
image

### Before
image

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/5874


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/432

PR #2688, opened 2023-02-02
feat(formvalidation): size rule to test string length range

### LINKED ISSUE #1279: Add form validation rule for min & max size range
Issue: Form validation does not include a range length rule. This is bad because I have to apply both the minLength __AND__ maxLength rules and if the user causes a validation failure then the message they see only informs them of the specific rule that they broke. For a field with an implied input size range, this is bad UX - we should inform the user of both limits of size in the same message.

Proposal: Add a size validation rule:

size            : '{name} must be between {ruleValue[0]} and {ruleValue[1]} characters in length',

Example - see screenName parameter below.

$('.ui.form')
  .form({
    fields: {
      name     : 'empty',
      gender   : 'empty',
      username : 'empty',
      password : ['minLength[6]', 'empty'],
      skills   : ['minCount[2]', 'empty'],
      terms    : 'checked',
      __screenName__    : __'size[10,20]'__,
    }
  })
;

Note: this requires ruleValue to handle an array of values and therefore may be more than a simple change.

lubber-de:
## Description

Added new `size` rule to check for a character length. It's a shortcut to avoid having 2 separate rules (minLength+maxLength)

## Testcase
https://jsfiddle.net/lubber/o71muv3p/7/

## Screenshot
image

## Closes
#1279

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -536,9 +536,9 @@
                             parts,
                             suffixPrompt
                         ;
-                        if (ancillary && ['integer', 'decimal', 'number'].indexOf(ruleName) >= 0 && ancillary.indexOf('..') >= 0) {
+                        if (ancillary && ['integer', 'decimal', 'number', 'size'].indexOf(ruleName) >= 0 && ancillary.indexOf('..') >= 0) {
                             parts = ancillary.split('..', 2);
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/434

PR #2694, opened 2023-02-08
feat(segment): wrapping in addition to stackable

lubber-de:
## Description

This PR adds an alias `wrapping` to the current `stackable horizontal segments` for consistency as we already use `wrapping` for button groups or menu items which actually wrap single elements instead of stacking _all_ of them at once which is meant by "stackable" in every other element.

The `horizontal stackable` variant gets deprecated and will be removed in 2.10


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/436

PR #2705, opened 2023-02-19
fix(dropdown): possible XSS through select option text

lubber-de:
## Description

This PR fixes a possible XSS through an entity encoded select option text when converted into a FUI dropdown.
Even if `preserveHTML: false` would prevent this, a select tag cannot contain html at all and if it contains entity encoded HTML instead, it  should not be reconverted into html.

The PR also fixes recreating the dropdown menu twice when no values are selected in a multiple dropdown

Thanks to @brian-codes for reporting

## Testcase
- Select the obvious evil dropdown entry

### Broken
The alert message appears upon selection
https://jsfiddle.net/lubber/f5to49ew/3/

### Fixed
All fine
https://jsfiddle.net/lubber/f5to49ew/5/

## Closes
https://github.com/fomantic/Fomantic-UI/discussions/2692#discussioncomment-5009642

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2065,7 +2065,7 @@
                                     values.push({
                                         name: name,
                                         value: value,
-                                        text: text,
+                                        text: module.escape.htmlEntities(text, true),
                                         disabled: disabled,
                                     });
                                 }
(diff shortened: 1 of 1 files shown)


lubber-de:
@all-contributors please add @brian-codes for security

PR #2714, opened 2023-02-23
[build] fix variable inheritance

hugopeek:
## Description

As described in https://github.com/fomantic/Fomantic-UI/pull/2671, you introduced conditional checks when loading theme variables in 2.9. This activated scoping, which I experienced also.

That PR didn't entirely resolve it for me though. In some cases (when importing variables from one theme in another), I still got errors. Turns out, the code was not entirely reverted to how it was in 2.8:

https://github.com/lubber-de/Fomantic-UI/blob/e39a0ae5183c12808ba1eee82082efbb15f4d450/src/theme.less#L21-L24

This PR restores the conditional for the component's default site.variables, which is apparently still needed for proper variable inheritance.

### DIFF
--- a/src/theme.less
+++ b/src/theme.less
@@ -19,7 +19,9 @@
 @import (optional) "@{themesFolder}/@{site}/globals/site.variables";
 
 /* Component's site.variables */
-@import (optional) "@{themesFolder}/@{theme}/globals/site.variables";
+& when not (@theme = 'default') {
+    @import (optional) "@{themesFolder}/@{theme}/globals/site.variables";
+}
 
 /* Site theme site.variables */
 @import (optional) "@{siteFolder}/globals/site.variables";


lubber-de:
Mmh, your change implements the scoping again. Means especially new variables declared inside theme/site.variables are not available inside the compoents.variables. That was fixed by #2671

What kind of situation fixes this change for you?

lubber-de:
@hugopeek I investigated quite some time now and i finally fixed both situations by #2715 
- avoid overriding the default variables
- avoid having them scoped 

So i would like to close this PR in favor of the other (i thought it was easier creating another PR instead of describing what you need to change in this PR to make it work)

lubber-de:
Closing, as #2715 was merged

hugopeek:
Wow, that seems like a thorough investigation @lubber-de.

I've only been able to test it now (sorry), and it builds my scenarios without errors, but there is still a difference in the output..

What I've been doing so far, is use a theme for each project, on top of my library theme. The theme variables inside the project can be changed in the CMS, which alters the projects site.variables and triggers a build.

The file structure looks something like this:

├── default
├── project
│   └── globals
│       ├── site.overrides
│       ├── site.variables
│       ├── theme.overrides
│       └── theme.variables
├── romanesco
│   ├── assets
│   ├── collections
│   ├── elements
│   ├── globals
│   │   ├── colors.less
│   │   ├── reset.overrides
│   │   ├── reset.variables
│   │   ├── site.overrides
│   │   ├── site.variables
│   │   ├── theme.overrides
│   │   ├── theme.variables
│   │   └── variation.variables
│   ├── modules
│   └── views

Content of project site.variables:

```less
@import url('../../romanesco/globals/site.variables');
@import url('./theme.variables');
```

Then, the projects theme.variables only contains values that should be overwritten in the Romanesco site.variables. A sample:

```less
/*******************************
        THEME VARIABLES

// DO NOT EDIT DIRECTLY!
// Generated by MODX when saving the Presentation settings under Configuration.
// If you want to tweak these variables manually, add them to site.variables instead.

@fontFallback: 'Helvetica Neue', Arial, Helvetica, sans-serif;
@headerFont: 'Source Sans Pro', @fontFallback;
@pageFont: 'Source Sans Pro', @fontFallback;

@scaleRatio: 1.333;
@fontSize: 1.25rem;

@pageBackground: #ffffff;

@logoPath: '/assets/img/logo.svg';
@logoInvertedPath: '/assets/img/logo-inverted.svg';
@logoBadgePath: '/assets/img/badge.svg';
@logoBadgeInvertedPath: '/assets/img/badge-inverted.svg';

@logoWidth: 697;
@logoHeight: 142;
@logoTargetWidth: 238;

@relativeBorderRadius: @relative4px;
@absoluteBorderRadius: @4px;
```

And the Romanesco site.variables looks more or less like the default.

So previously, by loading the Romanesco site.variables first, and then the project theme.variables, I was able to modify the Romanesco theme per project, while keeping it as the default for all things not overridden. This trick applies to other elements too, and I even extended it to facilitate multiple themes within the same project (for example for multilingual sites).

Now, after the implementation of #2715, the sequence of the imports doesn't seem to matter anymore. Whatever I try, the variables in the project theme.variables are being ignored.

I have no idea if what I'm doing here was ever supposed to work that way, but it did.. And with the reverted conditional in *this* PR, it still does.

Can you think of any way to respect the sequence in which the variables are imported, so they can be overrules from inside sibling themes? Or do you feel that this is outside of the scope of FUI? (and that I basically turned a bug into my own feature ;))

Thanks for all your efforts and patience!

lubber-de:
if `@site` is anything else than `default` it will be imported twice when the site component itself is imported.

Usually that does not hurt, as they are included right after each other in theme.less.
If you however now have additional includes inside, you should declare them as `multiple` , so i suggest you to try

```less
@import (multiple) url('../../romanesco/globals/site.variables');
@import (multiple) url('./theme.variables');
```

If this does not help, please provide your theme.config for a better understanding.

Basically your approach should work, but when your above mentioned files are included only once (less default) every additional import will override them, which possibly is the cause of your issue.

Issue #2241, opened 2022-02-19
[silder] Customizing the tick and half-tick labels of a ticked slider


mohammadhossein1976:
# Feature Request

  A ticked slider creates tick & half-tick labels automatically based on the ticked slider's width.

# Problems

  Things to consider:
   - Is it possible to add a config for only creating the beginning and ending label (not the inner labels). 
   - Is it possible to add a config to let the developer to decide which inner(tick & half-tick) labels should be created.
   - for sliders which have big min & max (more than 1000), because of creating many half-tick labels, the browser creates the slider with a huge lag time

## Example 

  How to create an example:
   1. Open the following JSFiddle - https://jsfiddle.net/e85wjc1x/10/
      (make the right side column as wide as possible)

   #### for the first issue
   2.  the image below is the price slider of our website, which the min & max price is different every time the page loads.
        and the min max difference is usually big, so i had to set 1000 steps as a constant.
   ![image](https://user-images.githubusercontent.com/55906406/154791681-99b6e8bc-c231-41c0-97bb-bd17bcafde81.png)
   so each step length is calculated by  (max-min)/1000, but I wanted a constant length such as 10, but because of the browser lag for ticked slider, i could only set 1000.
  in the img below, i had to manually hide all the inner tick & half-tick labels because i didn't want them for the small width slider.
   ![image](https://user-images.githubusercontent.com/55906406/154792945-5bb46ac1-60dd-4c4a-9995-f907b5462328.png)

   #### for the second issue
   2.1 if you set the slider width on 300 and the end input on 1000 you will be get these inner labels

![image](https://user-images.githubusercontent.com/55906406/154791783-a52e59db-eead-4b41-b529-33a44aed4d1e.png)

   2.2 but if you set the slider width on 1000 and the end input on 1000 you will be get these inner labels  

![image](https://user-images.githubusercontent.com/55906406/154791796-d8d3cb57-82c8-4ae9-a82e-3312c0a7638a.png)
  
   the labels in case 2.1 and 2.2 is different, I want to know is it possible to  set a config that:
   1. the developer could choose to have inner (tick and half tick) labels or not (just have the beginning and ending labels)
   2. if the developer want the inner labels, he can choose which ones to show (unauto mode)

lubber-de:
Implemented by #2726 
Use `restrictedLabels`

PR #2726, opened 2023-03-04
feat(slider): restrictedLabels, min/maxRange, always ticks, tooltip

### LINKED ISSUE #1731: [range slider] Ability to set a minimum duration
# Feature Request

There are many use cases where the START and END values of a range slider should never be the same.
A `minDuration` setting could be quite useful, and seems quite simple to implement (for someone who knows what they're doing).

## Example

Here's a simple time slider (with 15 minute time blocks) ...

https://jsfiddle.net/designosis/9sn16pta/

I tried building a basic `onChange` function that enforces a minimum duration of 30-minutes, but duration stays at `0` causing it to loop and fail.

Any tips on getting this to work?

## Screenshot

Screen Shot 2020-10-29 at 4 42 38 PM

lubber-de:
## Description
This PR adds new features to the slider module:

- minRange and maxRange settings to be able to restrict a range slider to a specific range by giving edge values 
  - minRange: 5  makes sure that the two thumbs always need to have a difference of 5 
  - maxRange: 10 makes sure that the two thumbs dont exceed a difference of 10
- restrictedLabels: an array of label values which restrict the displayed labels to only those which are defined
- showLabelTicks "always". This differs from "true" and comes in handy when used together with "restrictedLabels"
  - 'always' will always show the ticks for _all_ labels (even if not shown)
  - true will display the ticks only if the related label is also shown	
- showThumbTooltip option will display a tooltip containing the current value to the thumb(s)

## Testcase
https://jsfiddle.net/lubber/fzvynqom/20/

## Screenshot
image

## Closes
#1731
#2241

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -117,7 +117,6 @@
     [data-tooltip]::before,
     [data-tooltip]::after {
         pointer-events: none;
-        visibility: hidden;
         opacity: 0;
         transition:
             transform @tooltipDuration @tooltipEasing,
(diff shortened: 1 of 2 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/440

PR #2734, opened 2023-03-14
feat(menu): simpler equal width

### LINKED ISSUE #2732: [menu] equal width menu
# Feature Request
I would love if we could use `equal width menu` in the same fashion than `equal width grid`, where the menu items would be equally sized no matter how much they are.
This would be very convenient in cases the menu items are displayed programatically.

## Example (if possible)
https://jsfiddle.net/1fjabhs5/

lubber-de:
## Description
This PR unifies the `equal width` logic for menu as we already have in fields, grid and segment

I left the old approach ("x item menu") to avoid being a breaking change.
I will investigate if we can remove the old approach for 2.10.0 as , at the moment, i could not detect if there is a technical reason equal width menu items were done differently than in the other components

## Testcase 
https://jsfiddle.net/lubber/ydwfnur0/2/

## Closes
#2732

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1624,9 +1624,9 @@ Floated Menu / Item
 }
 
 & when (@variationMenuEqualWidth) {
-/* -------------------
-      Evenly Sized
--------------------- */
+    /* -------------------
+          Evenly Sized
+    -------------------- */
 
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/448

PR #2737, opened 2023-03-15
feat(search): onresults callback fromcache

### LINKED ISSUE #909: [Search] Cached results do not trigger callbacks
# Bug Report

If search is set to cache results, when a result is returned from the search cache onSearchQuery and onResults callbacks are not called.

## Steps to reproduce
1. Set up Search component and enable caching
2. Add logging statements in onSearchQuery and onResults callbacks
3. Enter search term, e.g. "Cancer"
4. Delete a character e.g. "Cance" - result is returned by cache
5. Add a character e.g. "Cancer" - result is returned by cache

## Expected result

I would expect the callbacks to be called whether results are retrieved from an API or from the local cache. In my use case, I use onSearchQuery to add the loading class to the search field, and onResults to remove the class. This works as expected with remote searches but due to the above does not work with cached results. I therefore disable caching and lose the benefit of this feature.

## Actual result

Callbacks are not called when results are returned from cache.

## Screenshot (when possible)
N/A

## Version
2.7.6

lubber-de:
## Description

The `onResults` callback was not firedwhen the results came from the cache.
This is now implemented. The callback will retrieve an additional 2nd parameter which will indicate if the call was initiated from remote (false) or cache (true)

## Testcase
- Watch Console
- Enter "test"
- Enter something else
- Enter "test" again

### Broken
onResult is not called when coming from cache
https://jsfiddle.net/lubber/Lphw05ek/5/

### Fixed
onResult is called from Cache aswell indicated by a second parameter
https://jsfiddle.net/lubber/Lphw05ek/7/

## Closes
#909

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -561,6 +561,7 @@
                         if (cache) {
                             module.debug('Reading result from cache', searchTerm);
                             module.save.results(cache.results);
+                            settings.onResults.call(element, cache.results, true);
                             module.addResults(cache.html);
                             module.inject.id(cache.results);
                             callback();
@@ -1345,7 +1346,7 @@
         onResultsAdd: false,
 
         onSearchQuery: function (query) {},
-        onResults: function (response) {},
+        onResults: function (response, fromCache) {},
 
         onResultsOpen: function () {},
         onResultsClose: function () {},


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/447

PR #2744, opened 2023-03-27
feat(flags): add alias for czech

kerolloz:
## Closes

#<issue id>, #<issue id>, #<issue id>

### DIFF
--- a/src/themes/default/elements/flag.variables
+++ b/src/themes/default/elements/flag.variables
@@ -396,7 +396,7 @@
     @1f1e8-1f1ff: {
         countrycode: cz;
         class: czechia;
-        aliasClass: false;
+        aliasClass: czech;
         aliasClass2: false;
     };
     @1f1e9-1f1f0: {


kerolloz:
@lubber-de Do I need to add it to the docs as well?

lubber-de:
> @lubber-de Do I need to add it to the docs as well?

That you be very nice, indeed 😉

src/themes/default/elements/flag.variables
```diff
@@ -396,7 +396,7 @@
     @1f1e8-1f1ff: {
         countrycode: cz;
         class: czechia;
-        aliasClass: false;
+        aliasClass: czech;
```
mvorisek, line 399: I wonder what is the motivation and consistency behind this change. `Czech` itself is a language and for other countries there is nothing like `french` class.

Shouldn't this be `czech_republic` (consistent with `united_states` for example)?

lubber-de: @mvorisek 
I trust in your decision here as you are a czech native speaker 🙂 , so if you think this is wrong i am fine by reverting or changing the alias name as you think it would be correct.
And your reasons are valid. Sorry for merging this too quick
So what should be do? Revert or Rename?

mvorisek: In general I would prefer the data from some external/consistent/popular source, but if the data are managed by Fomantic-UI, it should be renamed to `czech_republic`. Czechia is short form of Czech Republic introduced in 2006 - https://kafkadesk.org/2019/04/19/czech-republic-vs-czechia-a-linguistic-riddle/ but the long form is still used a lot.

lubber-de: Fixed by  #2746 and docs by https://github.com/fomantic/Fomantic-UI-Docs/pull/453

Issue #888, opened 2019-07-22
[Dropdown] Unable to customize the addition item


vincenccee:
# Help Wanted
I need some help to customize the `addition` item on the selection dropdown with the settings: 
```
allowAdditions: true
hideAdditions: false
```
## Problem
The problem here is where and when the option will be displayed. I want to always display the `addition` option (even when the user didn't type anything in the search) and want the `addition` option to always be the last option on the list.

By now I was able to customize the text and the style of the `addition` but not when it is displayed and where it is displayed. 

I tried to create a new option (with the `item` and without the `addition` classes) and bind a select event to it to create a new option, but when I tried to type something the new add option is filtered as well.

If I just create a new element at the end of the options list (without the `item` class) I lose the ability to navigate through the option with the keyboard keys, that's the reason that I need to use the `addition` item.

ivangomes:
@vincenccee are you still working to get this done? Please let me know!

vincenccee:
@ivangomes nope. I am not working with this anymore

PR #2745, opened 2023-03-27
feat(feed): connected, ordered, disabled, divided variant + inverted fix

lubber-de:
## Description
- Added a `connected` variant to the feed view, which now easily offers to use feed as a simple timeline component
- Added an `ordered` variant which automatically creates labels or can create ones via `data-text` attribute including colorization as usual and a basic variant
- Added support for text labels or `ui label` labels
- Added `disabled` support
- Added a simple `divided` variant just like we already have for item, list, grid
- Added optional `right floated` variant for feed sub components such as `date`
- Fixed inverted variants when icons were used

## Testcase
https://jsfiddle.net/lubber/t37qocuk/18/

## Screenshot
image
image
image
image
image


mvorisek:
Masking the line under icons with
```
.ui.connected.feed > .event > .label i.icon::before {
  background: #fff !important;
}
```
requires white/known background color, which imply it will not work correctly with custom background, example with added `background-color: #888;`: https://jsfiddle.net/e218k3fL/ . With background image, the masking concept cannot be even fixed by user.

Is there any way how to "cut" the line under the icons instead of "masking"?

lubber-de:
@mvorisek 
Here is another approach:
https://jsfiddle.net/lubber/arkehjox/

What do you think? 

This one cannot completely connect the icons /images  as the height of the icon is unknown and sometimes differs in height. As a minimum i had to differ between icon and image labels.
I tried the best to make it look nearly identical as before.

As image and icon definately needs different connector sizes (as the line is now implemented at the label (to make sure it starts after the shown icon/image) rather than event, i had to use two selector approaches here: :has() and (as fallback to firefox (and IE *cough* :D ) an additional .image or .icon class.

But at least your mentioned background issue is now gone.

What i also thought / tried (without success and lots of frustration):
- setting a (white) background color to the whole event instead of only the mask (so it's not transparent by default) which the background icon mask would simply inherit then (would still not solve the "background-image" situation, but the inverted variant already uses this approach)
- making use of trickery like mix-blend-mode or background-clip..... did not work as we would need to vanish parts of the icon which have no bits set (the outline circle in that case) and markup was too special/messed up
- using flexbox inside the existing  flexbox label: did not work well, the pseudo selectors (i also tried an additional node) always messed the height and overlayed the next event. Did especially mess up image variants as labels
- completely rewrite the whole component.... not worth it...at least not now ;)

mvorisek:
The FF demo looks good to me in FF. I cannot comment much on the impl. as your CSS skills are much better than mine :).

lubber-de:
I just adopted the approach to the original idea of using the whole .event selector which removes the need for extra classes and eadling with :has
https://jsfiddle.net/lubber/t37qocuk/3/

I adjust the PR accordingly

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/474

PR #2751, opened 2023-03-31
fix(formvalidation): support native validation, after submit only

### LINKED ISSUE #2750: [Form] Input Type Number with Min Max on Error Should not Trigger ``onSuccess()``
# Bug Report
The input as type ``number`` with min max:
```html
<input type="number" min="0" max="99" name="number" placeholder="Number">
```
Should not trigger the ``form.onSuccess()``, because it is clearly a failure!

# Example
https://jsfiddle.net/KiddoV/exaysodp/14/
Expecting the ``onSuccess()`` not calling when input less than 0 or greater than 99!!

lubber-de:
## Description
Form validation will trigger success even if native form validations are still invalid.
This PR respects any native form validation by the browser in addition to the FUI form validation rules

Native invalid fields (like required file inputs) are now only shown as error'ed after a form submit has taken place at least once

I also fixed a possible double submit when the submit button is not a native submit input.

## Testcase
https://jsfiddle.net/lubber/7L4qhzjg/2/

## Closes
#2750

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1344,16 +1344,16 @@
                             if (showErrors) {
                                 $field.closest($group).addClass(className.error);
                             }
-                        } else if (isDisabled) {
+                        } else if (showErrors) {
+                            $field.closest($group).removeClass(className.error);
+                        }
+                        if (isDisabled) {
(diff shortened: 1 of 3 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/456

PR #2758, opened 2023-04-13
feat(form): optional limit the displayed error messages per field

lubber-de:
## Description
Added an optional setting `errorLimit` (default 0 = unlimited).
When set to an integer value, only that given number of possible rule errors per field are displayed 

## Testcase
Submit the form

### Normal
All rule error for the single field are displaye (empty and not long enough)
https://jsfiddle.net/lubber/eqdo1huy/6/

### Improved
Only one rule error of the field is displayed 
https://jsfiddle.net/lubber/eqdo1huy/8/

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/6361

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1355,7 +1355,7 @@
                             module.debug('Field depends on another value that is not present or empty. Skipping', $dependsField);
                         } else if (field.rules !== undefined) {
                             $.each(field.rules, function (index, rule) {
-                                if (module.has.field(identifier)) {
+                                if (module.has.field(identifier) && (!settings.errorLimit || fieldErrors.length < settings.errorLimit)) {
                                     var invalidFields = module.validate.rule(field, rule, true) || [];
                                     if (invalidFields.length > 0) {
                                         module.debug('Field is invalid', identifier, rule.type);
@@ -1607,6 +1607,7 @@
         preventLeaving: false,
         errorFocus: true,
         dateHandling: 'date', // 'date', 'input', 'formatter'
+        errorLimit: 0,
 
         onValid: function () {},
         onInvalid: function () {},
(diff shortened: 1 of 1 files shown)


lubber-de:
> Wouldn’t this limit it to just N errors for every field? What if wanted to show a different number of errors for other fields?

Good hint! I added support for `errorLimit` for each field individually now by f6ccd4be2e9f5e65bba3ecb06a30d2d071cd3c6e so

```javascript
// username will show both error messages
// password will only show 1
$('.ui.form').form({
    fields: {
		username: {
	      identifier : 'username',
	      rules: [
	        {
	          type   : 'empty',
	          prompt : 'Please enter a user'
	        },
	        {
	          type   : 'minLength[6]',
	          prompt : 'Your username must be at least 6 characters'
	        }
	      ]
	    },
        password: {
	      identifier : 'password',
	      errorLimit: 1,
	      rules: [
	        {
	          type   : 'empty',
	          prompt : 'Please enter a password'
	        },
	        {
	          type   : 'minLength[6]',
	          prompt : 'Your password must be at least 6 characters'
	        }
	      ]
	    },
    }
});
```

> I think for the scenario that is in the linked issue (and the JSFiddle), if the empty rule is used, then this rule should be processed first (whether it is the first rule or last rule) - if the field is empty, then that should be the only error to appear and nothing else. It makes no sense to show any other errors, ie. if it is empty then of course it won’t have a minimum number of characters; if it is empty of course it won’t match a reflex pattern; if it is empty of course the value will be smaller than X

The example in the SUI issue indeed makes no sense when 'empty' is checked after the minLenght rule.
I would leave this up to the user as before as the order inside the field rule array is iterated inside the form validation.

We would otherwise need to sort the rules our own and decide ourselves which rule should have what kind of priority.

PR #2761, opened 2023-04-17
feat(item,comment): disabled variants

lubber-de:
## Description

Added a `disabled` variant to comment and item because such elements could contain interactive elements and it's quite a common use case to have all/single comment(s)/item(s) being disabled


ColinFrick (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/461

PR #2774, opened 2023-05-03
fix(menu): dont show right wrapping border on secondary

### LINKED ISSUE #2772: [menu] Option wrapping in secondary menu with border-right in last-child item
# Bug Report
The Wrapping option on new in 2.9.2 is perfect. But when using with the `secondary.menu` option, the last item is gettin a `border-right`. - https://fomantic-ui.com/collections/menu.html#wrapping

## Steps to reproduce
In the collections Menu, adding CSS class `wrapping` with `secondary` option. The last item has a border-right. If you add the CSS `centered` the border-right is removed, but the last item centered is not wanted.

## Expected result
The correct is with border-right none. 

## Testcase
https://jsfiddle.net/crg0qtfo/

## Screenshot (if possible)
The second example is correct.
image

## Version
2.9.2

lubber-de:
## Description
When using the wrapping option for some menu variants (such as secondary, text, tabular or borderless)  a very right border to the last item is not wanted .

## Closes
#2772

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1936,7 +1936,7 @@ Floated Menu / Item
         & .item:first-child::before {
             display: none;
         }
-        &:not(.tabular) .item {
+        &:not(.secondary):not(.text):not(.tabular):not(.borderless) .item {
             &:last-of-type,
             &:last-child {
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

Issue #1868, opened 2021-02-07
JavaScript collection helpers/plugins


jamessampford:
I was putting together something for #1501 to help with overflowing horizontal tabular menus (since they are not great when they overflow on a narrow screen). Though menu does not currently have a JavaScript module to extend, so throwing this in here to see if/how we could implement this kind of functionality since I do not think it can be done purely with CSS

Demo:
https://jsfiddle.net/awcn96p0/

Essentially this could add a new `.menu()` module which could be applied to a menu to allow a menu to allow JavaScript to set the overflow method to none, auto, dropdown, and scroll

- None is essentially the current setup (too many goes offscreen)
- Dropdown does some calculations and if the tab will end up overflowing, will hide them, and add them to a dropdown
- Scroll essentially will allow you to scroll through them with touch (similar way to Github when viewing on mobile)
- Auto is a combination of dropdown and scroll, if the device is touch, then you get scroll, otherwise have dropdown

The slightly troublesome thing is I do not know if this should be a module (one function seems a little silly), and there are probably other examples of things that could potentially be added in where JavaScript makes a bit more sense to provide the functionality

Which made me think should there be some kind of standard way of perhaps making a plugin system which could make these kinds of things available and ensuring that other modules have been included if they need them when they run?

So this could possibly instead look like `.plugin('pluginname', {plugin: settings})` where plugin does all the loading of modules, and providing all the base functionality (like debugging)

PR #2776, opened 2023-05-07
feat(menu): add tab overflow scrolling and shadows

### LINKED ISSUE #1501: Scrollable Tab or icon for scroll when the Tab width exceeds the container width
# Feature Request
The Tabs gets hidden or it makes its parent element scroll when the width of Tab exceeds its container width. Shouldn't this be handled using either of two:

1. The Tab itself gets scrollable
2. An icon gets added to the ends of the Tab

Sample
https://codesandbox.io/s/fomantic-tab-issue-v6er3?file=/src/App.js

I tried it by changing the fomantic CSS as 
```
.ui.secondary.pointing.menu{
    overflow-x: scroll;
    overflow-y: hidden;
    &::-webkit-scrollbar {
      display: none;
      width: 0;
      background: transparent;
    }
 }
```
But this made the active Tab bottom border gone. Also, the scrollbar not hiding.

SeanSun6814:
## Description

Currently, when there are too many tabs on a menu such that they exceed the width of its container width, they overflow off the page. As mentioned in issue #1501, there are two ways this could be fixed:

- The Tab itself gets scrollable
- An icon gets added to the ends of the Tab

In this PR (@seansun6814, @jessxec), we address both of them. It allows the tabs to be scrollable if it overflows. When some tabs are hidden to the left or right, there are shadows on either side indicating that more tabs can be scrolled to.

In order to activate these features, the user needs to add some more HTML `div`s in their code.

Specifically, to have the scroll feature, the user should wrap the menu tabs div with another div:
```
<div class="ui top attached tabular scroll">
```
To have the shadows feature, the user should add the following `div`s right before and after the menu tabs:
```
<div class="ui top attached tabular menu">
  <div class="ui attached tabular shadowHider top shadowHiderStart"></div>
  <div class="ui attached tabular shadow shadowStart"></div>

  <div class="item" data-tab="one">one</div>
  <div class="item active" data-tab="two">two</div>

  <div class="ui attached tabular shadow shadowEnd"></div>
  <div class="ui attached tabular shadowHider shadowHiderEnd"></div>
</div>
```
After these `div`s are added, the CSS added into the fomantic library will automatically apply to them.

In the future, it might be worth looking into having these features without additional user code.

## Testcase
Demo before the fix (from the issue #1501): https://codesandbox.io/s/fomantic-tab-issue-v6er3?file=/src/App.js

Demo after the fix: https://jsfiddle.net/jessxec/64L293uk/31

### Before the fix: 

image

### After the fix:

Fomantic_scrolling_issue-output

## Closes

#1501

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1923,6 +1923,84 @@ Floated Menu / Item
             margin-right: 0;
             width: 100%;
         }
+
+        .ui.top.attached.tabular.scroll {
+            display: flex;
+            position: relative;
+            overflow-x: auto;
+            overflow-y: hidden;
+            overscroll-behavior: contain;
+        }
+        
+        .ui.top.attached.tabular.scroll::-webkit-scrollbar {
+            display: none;
+            width: 0;
+            background: transparent;
+        }
+        
+        .ui.top.attached.tabular.menu>.item.active {
+            border-bottom: 1px solid #ffffff;
+        }
+        
+        .ui.top.attached.tabular.menu {
+            overflow: visible;
+            width: 100%;
+            min-width: fit-content !important;
+        }
+        
+        .ui.bottom.attached.tab.segment {
+            border-top: none !important;
+        }
+        
+        .ui.top.attached.tab.segment {
+            border-bottom: none !important;
+        }
+        
+        .ui.bottom.attached.tabular.menu {
+            overflow: visible;
+            width: 100% !important;
+            min-width: fit-content !important;
+        }
+        
+        .ui.attached.tabular.shadowHider {
+            border: black 1px;
+            margin-right: 1rem;
+            position: absolute;
+            height: 100%;
+            width: 1em;
+            background-color: white;
+            z-index: 2;
+        }
+        
+        .ui.attached.tabular.top.shadowHiderStart {
+            border-bottom: 1px solid #d4d4d5 !important;
+        }
+        
+        .ui.attached.tabular.shadowHiderEnd {
+            height: 100%;
+            position: sticky;
+            margin-left: -1rem;
+            flex: 0 0 1rem;
+            left: calc(100% - 2rem);
+        }
+        
+        .ui.attached.tabular.shadow {
+            position: sticky;
+            z-index: 1;
+            flex: 0 0 1rem;
+        }
+        
+        .ui.attached.tabular.shadowStart {
+            left: 0;
+            background-image: linear-gradient(90deg, #0004, transparent);
+        }
+        
+        .ui.attached.tabular.shadowEnd {
+            right: 0;
+            background-image: linear-gradient(90deg, transparent, #0004);
+            width: 1rem;
+            left: calc(100% - 2.1rem);
+        }
     }
 }
 


lubber-de:
Thanks for investigating and proposing a possible solution. :+1:
The way your approach is made is somehow not completely fitting the way FUI works or is considered to work.

Here are experience while trying your jsfiddle:

- The extra wrapper only seems to be needed to make the extra shadow divs work and to hide the segments top/bottom border. Tricky solution. I wonder if this is somehow working without the extra divs :thinking:
- The wrapper should not have the nearly same classname as the menu itself. confusing and does currently not fit the FUI naming logic. But this is something easily fixable/discussable
- If extra divs are still needed/accepted, we should give them a more neutral name, because covering the left/right side could possible be something else than a shadow
- Your approach does not work on desktop, on Edge you need to use the direction keys to scroll the menu items, on Chrome or firefox it does not work at all (the direction keys controls the cursor inside the heading instead). Only on mobile the menubar is draggable
  - on desktop would either need the scrollbar to be displayed (yes, ugly, i know :wink: ) or the scroll hhandling would need some javascript (my visual approach seen below using extra items)
- the shadow trick, only being shown when not scrolled to the edges, is clever, but, IMHO, does not visually fit the overall tab/FUI appearance, it looks misplaced to me. We could play around using someting being more "visual pleasing" (for example radial gradient, rounded corners...
- Hiding the scrollbar does not work at all in firefox

Here is a quick draft i played around removing the extra divs and placing left/right items instead. Yes, those would need javascript, but  .tab() actually is a module, so we can add such logic here. Such logic could change the scrollposition upon click (or implement a dragging functionality), which works everywhere.
It also allows to use any desired (cover-/click-) item instead of a hardcoded "shadow"  
https://jsfiddle.net/lubber/y98m7u2n/3/

Again, thanks for getting into this, and lets discuss how to proceed.

src/definitions/collections/menu.less
```diff
@@ -1923,6 +1923,84 @@ Floated Menu / Item
...
+        }
+
+        .ui.top.attached.tabular.menu > .item.active {
+            border-bottom: 1px solid #fff;
+        }
+
+        .ui.top.attached.tabular.menu {
+            overflow: visible;
+            width: 100%;
+            min-width: fit-content !important;
```
lubber-de, line 1948: dont use !important, use repeating classnames to increase specifity instead

src/definitions/collections/menu.less
```diff
@@ -1923,6 +1923,84 @@ Floated Menu / Item
...
+        }
+
+        .ui.top.attached.tabular.scroll::-webkit-scrollbar {
+            display: none;
+            width: 0;
+            background: transparent;
+        }
+
+        .ui.top.attached.tabular.menu > .item.active {
+            border-bottom: 1px solid #fff;
```
lubber-de, line 1942: Use a less variable instead of hardcoding the value

lubber-de (changes requested):
Just 2 places inside your PR which need changes, beside my other comment, which you should take care of for future changes, those are needed at several places , which i havent marked because of my other overall comment

jamessampford:
>   * on desktop, it would either need the scrollbar to be displayed (yes, ugly, i know 😉 ) or the scroll handling would need some javascript (my visual approach seen below using extra items)

I did come up with something in #1868 that did scrolling and overflow (using dropdown) menu, depending on what was needed (auto detected if mobile) and would be useful in `.tab()` (though in a way could be useful to have this logic for other menus, like a header menu, which was more what my proposal was around)

jamessampford:
>   * on desktop, it would either need the scrollbar to be displayed (yes, ugly, i know 😉 ) or the scroll handling would need some javascript (my visual approach seen below using extra items)

I did come up with something in #1868 that did scrolling and overflow (using dropdown) menu, depending on what was needed (auto detected if mobile) and would be useful in `.tab()` (though in a way could be useful to have this logic for other menus, like a header menu, which was more what my proposal was around)

Issue #2663, opened 2023-01-17
Dropdown closing behind other form fields


jeffmcfly:
# Bug Report
When dropdown field closes, it passes behind other form fields.

## Steps to reproduce
1. Create a form with a dropdown field
2. Place other fields below
3. Set long transition duration to see effect

NB. Using "slide" as "hideMethod" seems to resolve the issue : 

`transition: {
      showMethod   : 'slide down',
      showDuration : 200,
      hideMethod   : 'slide',
      hideDuration : 5000,
  }` 

## Expected result
Dropdown closing before other elements

## Actual result
Dropdown closing behind other elements

## Testcase
https://jsfiddle.net/1w0vz9tL/1/

## Screenshot (if possible)
![image](https://user-images.githubusercontent.com/94075859/213015825-ddf78dae-b21d-4383-81fd-2cb027f6e730.png)

## Version
2.9.1

krpatel5657:
We are a group of students from CMU and we'd like to work on this issue!

krpatel5657:
Hello! Upon changing a line in src/modules/dropdown.js to have the auto dropdown transition default to "slide" instead of "slide down," I get the following successful result:
![drop-down-working](https://user-images.githubusercontent.com/52473367/236687366-b9af03bd-5209-4253-8aab-6cf327a09cdc.gif)

This change also fixes an issue with the upwards direction of dropdown menus. Here is the original without this change:
![drop-up-notworking](https://user-images.githubusercontent.com/52473367/236687379-0ad6bf35-bfaa-40c6-a704-949847b14f52.gif)
Here is the upwards dropdown with the change:
![drop-down-working](https://user-images.githubusercontent.com/52473367/236687390-60edaff9-327f-4d30-8b85-aa3f401858c4.gif)

The line changed was within the transition property under the get method in dropdown.js, from:
`transition: function ($subMenu) {
                        return settings.transition === 'auto'
                            ? (module.is.upward($subMenu) ? 'slide up' : 'slide down')
                            : settings.transition;
                    },`
to:
`transition: function ($subMenu) {
                        return settings.transition === 'auto'
                            ? (module.is.upward($subMenu) ? 'slide up' : 'slide')
                            : settings.transition;
                    },`

Please let me know if this is considered an acceptable fix!

lubber-de:
Fixed by #2818 
See your adjusted jsfiddle here https://jsfiddle.net/lubber/Ld624p5m/

PR #2777, opened 2023-05-07
fix(dropdown) Changed "slide down" transition to "slide"

krpatel5657:
## Description

Previously, dropdown menus would close behind other form elements if the two were vertically located next to each other. The fix changes the default transition for dropdown menus when opening or closing to be the "slide" method, instead of "slide down." This results in the dropdown menu closing on top of other form elements. Also, there was an error with dropdown menus that open upwards, as shown in the gifs below. These menus would then slide even further up to close, instead of collapsing down on the form field. The change to the "slide" transition also fixed this issue, again as shown by the gifs.

## Testcase

https://jsfiddle.net/krpatel5657/hqe1pnu5/4/

## Screenshot (if possible)

Original issue:
dropdown_wrong
Fix:
drop-down-working

Simultaneous fix with upwards dropdown menus:
Original error:
drop-up-notworking
Fix: 
drop-up-working

## Closes

#2663

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1897,7 +1897,7 @@
                     },
                     transition: function ($subMenu) {
                         return settings.transition === 'auto'
-                            ? (module.is.upward($subMenu) ? 'slide up' : 'slide down')
+                            ? (module.is.upward($subMenu) ? 'slide up' : 'slide')
                             : settings.transition;
                     },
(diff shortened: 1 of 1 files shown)


lubber-de:
Your change does only "heal the wounds", but the real reason for the issue is a double hide transition which messes up both (underlaying menu and upward hiding)
I fixed that by #2818 and will close this PR

Issue #729, opened 2019-05-08
[Site] Color palette with prefixes like dark/light


exoego:
# Feature Request

## Description

I found myself adding different tones of color (e.g. light yellow, dark blue) to CSS theme for styling headers/segments/labels/etc, since SUI/FUI only have single tone for colors.
It would be handy if SUI/FUI provide a consitent and beautiful color palette system OOTB, by  prefixes like `darker red button` or `lightet teal segment`.
This allow users to have colors in several tones for different purposes (e.g., `light` as primary, `darker` as accent)

### [Material Design](https://material.io/tools/color/#!/?view.left=0&view.right=0)
![image](https://user-images.githubusercontent.com/127635/57348103-cde35300-718f-11e9-8e7e-2a853f4729ab.png)

### [Tailwind CSS](https://tailwindcss.com/docs/colors)
![image](https://user-images.githubusercontent.com/127635/57348059-a68c8600-718f-11e9-97dc-6d41180a5685.png)

crewow:
it will be great

crewow:
great idea

PR #2778, opened 2023-05-08
feat(colors): added class names that correspond to different shades of red, green, and blue

wadihPazosJr:
Many people find themselves using different tones of color for CSS styling, and it would be helpful to add class names that correspond to different shades of colors, instead of having to seek the correct Hex value.  Thus, this pr adds classes corresponding to different shades of red, green, and blue. There is the base shade, then there is dark, darker, darkest, light, lighter, lightest shades. 

An example of these changes can be found at the following [jsfiddle](https://jsfiddle.net/wpazos/9fauentq/)

image

This is a potential fix for issue #729.

### DIFF
--- a/src/definitions/globals/site.less
+++ b/src/definitions/globals/site.less
@@ -108,6 +108,84 @@ p:last-child {
     margin-bottom: 0;
 }
 
+/*******************************
+        Color Classes
+*******************************/
+
+.red {
+    color: #e3342f;
+}
+
+.dark-red {
+    color: #cc1f1a;
+}
+
+.darker-red {
+    color: #621b18;
+}
+.darkest-red {
+    color: #3b0d0c;
+}
+.light-red {
+    color: #ef5753;
+}
+.lighter-red {
+    color: #f9acaa;
+}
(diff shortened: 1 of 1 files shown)


lubber-de (changes requested):
Thanks, but this approach does not fit the FUI logic for the following reasons

- colors are centrally defined to be consumed by every component in https://github.com/fomantic/Fomantic-UI/blob/develop/src/themes/default/globals/colors.less
https://fomantic-ui.com/introduction/build-tools.html#colorsless
  - You hardcoded the values without making use of LESS variables
  - You hardcoded color names 
  - You did not cover all existing color names
- You only declared values for `color`. What if somebody needs `ui darker red message` which would also need adjustments for background, border, box-shadow
- what about `inverted`
- FUI logic would follow `dark red` instead of `dark-red`. Yes, this has to be implemented carefully but that was exactly the idea of #729 (only add "darker", "lighter", etc to any color). This in turn would need a rewrite of the coloring concept using CSS variables. That's a complete whole new concept.

PR #2158, opened 2021-11-14
fix(tab): initializing active tab with historytype 'state'

dutrieux:
If I initialize the active tab with the name of second tab in the property `autoTabActivation` and set the property `historyType` to `state`, the active tab is not set to the right tab (the second).

If I set the property `historyType` to `hash`, that's work !

## Broken
https://jsfiddle.net/dutrieux/p3oru58t/ 

## Fixed
https://jsfiddle.net/dutrieux/b4r8n7hg/

## Closes
#2154

PR #2786, opened 2023-05-12
fix(tab): hash historytype did not work anymore

### LINKED ISSUE #2769: [Tab] Tab history using historyType=hash does not restore last active tab in 2.9.x
# Bug Report
Using 2.9.x, tab history using historyType=hash no longer restores the last selected tab upon reload/navigation. This bug even affects the official documentation @ https://fomantic-ui.com/modules/tab.html#/examples (notice that navigating to that link does not activate the "Examples" tab)

## Steps to reproduce
1.  Just follow the documentation on tab history @ https://fomantic-ui.com/modules/tab.html#with-hash-tags to create a simple tabset with minimum two tabs
2. Click on other than the first tab. Ensure the history feature is enabled by observing that the URL hash changes.
3. Reload the URL that includes the tab hash.

## Expected result
The last active tab was restored.

## Actual result
The first tab was activated, ignoring the tab hash.

## Testcase
This is the jsfiddle @ https://jsfiddle.net/ky9was7m/1/ but since jsfiddle uses iframe we can't see nor reload the URL hash.

## Version
2.9.x (since 2.9.0)
This issue does not affect 2.8.x

lubber-de:
## Description
Hash historyType (default) was broken since 2.9.0 , because #2158 was only targeted to historyType "state", but did not fetch it.

## Closes
#2769

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -99,7 +99,7 @@
                     module.bind.events();
 
                     if (settings.history && !initializedHistory) {
-                        module.initializeHistory();
+                        settings.history = module.initializeHistory();
                         initializedHistory = true;
                     }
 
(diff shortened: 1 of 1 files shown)


ko2in:
LGTM.

PR #2793, opened 2023-05-21
feat(form): allow to ignore init of existing fui modules inside form

### LINKED ISSUE #2791: [dropdown] `clearable` is not working when form is initialized with `$('#form_id').form()`
# Bug Report

## Steps to reproduce
1. open https://jsfiddle.net/01y3cf9z/
2. select some country
3. notice there is no clearable icon even if the dropdown was initialized with `$('#dd_id').dropdown({ 'clearable': true })`
4. notice when `$('#form_id').form()` is removed, the clearable icon is present with exactly the same steps as described above!

## Expected result

`$('#dd_id').dropdown({ 'clearable': true })` must imply clearable icon always present

## Actual result

clearable icon is not present when form is initialized with Fomantic-UI JS

## Screenshot (if possible)

image

## Version

2.9.2 / latest

lubber-de:
## Description

In case one wants to handle dropdown or calendar init _after_ form init, this PR adds an optional `ignoreModulesOnInit` setting to support that.

## Testcase
Selecting any value inside the dropdown must be clearable
https://jsfiddle.net/lubber/3pLrgvsa/1/

## Closes
#2791

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1090,14 +1090,16 @@
                                     ? $el.is(':checked')
                                     : $el.val()
                             ;
-                            if (isDropdown) {
-                                if ($parent.is(selector.uiDropdown)) {
-                                    $parent.dropdown('save defaults');
-                                } else {
-                                    $el.dropdown('save defaults');
+                            if (!settings.ignoreModulesOnInit) {
+                                if (isDropdown) {
+                                    if ($parent.is(selector.uiDropdown)) {
+                                        $parent.dropdown('save defaults');
+                                    } else {
+                                        $el.dropdown('save defaults');
+                                    }
+                                } else if (isCalendar) {
+                                    $calendar.calendar('refresh');
                                 }
-                            } else if (isCalendar) {
-                                $calendar.calendar('refresh');
                             }
                             $el.data(metadata.defaultValue, value);
                             $el.data(metadata.isDirty, false);
@@ -1606,6 +1608,7 @@
         duration: 200,
 
         autoCheckRequired: false,
+        ignoreModulesOnInit: false,
         preventLeaving: false,
         errorFocus: true,
         dateHandling: 'date', // 'date', 'input', 'formatter'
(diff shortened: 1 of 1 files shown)


mvorisek:
> https://github.com/fomantic/Fomantic-UI-Docs/pull/469#issuecomment-1556189443

I would prefer `initChildModules` (with `true` as default for BC) setting instead /w a console warning when an uninitialized module is observed when needed.

lubber-de:
This PR is obsolete as the main issue was fixed by #2794

lubber-de:
> I would prefer `initChildModules` (with `true` as default for BC) setting instead /w a console error when an uninitialized module is observed (when needed for some op).

I implemented that in the latest commit, however, as i recognized the basic issue was caused by dropdown rather than form, i dont think we should have the need to implement that setting anymore

> 
> Also, to fix #2791 fully, is it possible to emit a console error when a module is initialize more then once? (as as discovered in the issue, 2nd init setting is ignored and it implies a broken usage).

This would be needed inside every module and in some cases this possibly isnt an error (re-instantiating is possible by design).
Howver, you can simply adopt my attempt insde the last commit and check for a previous instantiation yourself.

```javascript
let $element = $('.ui.dropdown');
     iWasInstantiatedBefore = $element.data('module-dropdown');

if (!iWasInstantiatedBefore) {
    $element.dropdown();
}

```

mvorisek:
> This would be needed inside every module and in some cases this possibly isnt an error (re-instantiating is possible by design).
However, you can simply adopt my attempt inside the last commit and check for a previous instantiation yourself.

Thank you. If re-instantiating is by design and working properly, I agree this PR is not needed anymore.

ko2in:
@lubber-de Do I have to approve this PR?

I've understood that you've said we're not gonna need this implementation [here](https://github.com/fomantic/Fomantic-UI/pull/2793#issuecomment-1556267463) and [here](https://github.com/fomantic/Fomantic-UI/issues/2791#issuecomment-1556265074).

lubber-de:
> @lubber-de Do I have to approve this PR?
> 

No, i'll close this without merging and removed the "awaiting-review" tag

PR #2800, opened 2023-05-24
fix(form): add prompt while still animating

### LINKED ISSUE #2796: [form] Something in form prompt handling changed to asynchronous
# Bug Report

After v2.9.2 some code allowing race condition was introduced.

With https://cdn.jsdelivr.net/npm/fomantic-ui@2.9.2/dist/semantic.js there is no problem. With https://cdn.jsdelivr.net/npm/fomantic-ui@2.9.3-beta.35/dist/semantic.js the problem is present.

This is important as before AJAX/submit, prompts are removed but AJAX response can add them back quickly.

## Steps to reproduce
1. open https://jsfiddle.net/mL3ko1st/1/
2. click "Save"
3. notice prompt displayed with text: "some error action 45"
4. click "Save" again, notice the prompt is gone

## Expected result

prompt must stay as in v2.9.2

## Version

\>2.9.2 \<2.9.3, ie. develop between these releases

lubber-de:
## Description

If a manual `add prompt` is triggered while the form is submitting / ends validation and possibly existing prompts get faded out via transition, then the new prompt is not shown as it is still found in the dom, but not recognized that the prompt will vanish from the dom once the transition has ended.

I also fixed fadein without transition module as the fadeIn did not work at all.

## Testcase
- Click on Submit -> "some error action 45" is displayed
- Click on Submit again -> prompt vanishes although it should be displayed (again)

### Broken
https://jsfiddle.net/mL3ko1st/1/

### Fixed
https://jsfiddle.net/lubber/gvhcaryf/

## Closes
#2796

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -911,7 +911,8 @@
                             $field       = module.get.field(identifier),
                             $fieldGroup  = $field.closest($group),
                             $prompt      = $fieldGroup.children(selector.prompt),
-                            promptExists = $prompt.length > 0
+                            promptExists = $prompt.length > 0,
+                            canTransition = settings.transition && module.can.useElement('transition')
                         ;
                         module.verbose('Adding field error state', identifier);
                         if (!internal) {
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

mvorisek:
Tested 👍

PR #2801, opened 2023-05-24
feat(chore): optional outline on focus-visible

### LINKED ISSUE #2799: [ux] Native tab focus must not be visually supressed
# Feature Request

When `tabindex="0"` is set to a (div) button, the div can be focused by a tab.

Currently when such div is focused, no native glow is present. I have identified this is because:
image
`outline: none` style.

## Example (if possible)
1. open https://dev.atk4.org/demos/form/form.php
2. click the first input field
3. press tab
4. notice the button background changed very slightly, but native tab outline around the button is not present

## Screenshot (if possible)

normal:
image

focused:
image

compare it with tab glow in https://dev.atk4.org/demos/interactive/paginator.php for example:
image

# Expected fix/behaviour

Not sure why there is `outline: none`, but at least for `:focus-visible` pseudo class no `outline: none` style must be applied.

image
(focused button without `outline: none` style, notice the native outline/border around the button, the color is not much different visually, but the visual dimension difference makes it very easy to understand what button is (tab) focused)

lubber-de:
## Description

This PR allows to declare a specific focus-visible outline for some components where the default outline is set to none.
The default will be kept to none because FUI mostly shows focus by providing a different color (e.g. buttons) or would not visually fit.

To reduce CSS selectors only a value != none will be rendered, so by default the CSS output wont change

Now the user can decide to change `@focusVisibleOutline` globally in a custom site.variables or individually inside a custom themes {component.variables}

## Closes
#2799


mvorisek:
See https://stackoverflow.com/questions/31402576/enable-focus-only-on-keyboard-use-or-tab-press for analysis how big players style it.

In #2799 I have described an issue for tab focus for button (and dropdown). As the PR is coded now, it will imply outline for `:focus`, ie. bold outline for buttons clicked by a mouse. This is not wanted.

Is there any way how to simply not add `outline: none` when `:focus-visible` is active? At least for browsers that support this pseudo class - https://caniuse.com/css-focus-visible.

lubber-de:
> Is there any way how to simply not add `outline: none` when `:focus-visible` is active? At least for browsers that support this pseudo class - [caniuse.com/css-focus-visible](https://caniuse.com/css-focus-visible).

Sure
```css
.ui.button:focus-visible {
  outline: revert;
}
```

https://jsfiddle.net/lubber/y0ba824e/10/

mvorisek:
Yes, please add this revert style for buttons and other important components.

mvorisek:
I consider this a bug. `.ui.button:focus-visible` selector will not focus visually on a button click nor if a button has no tabindex set. So there should be no visual change on anything except when a button is navigated via tab which deserves a proper styling. I belive the current Fomantic-UI simply does not account for tab focus styling.

One proof - `a:focus-visible` is styled currently natively as well. Just try it. Another proof - for ex. Google does not style it as well ;-).

So I propose:
a) revert the outline for button with `.ui.button:focus-visible` selector
b) fix tab visual focus of dropdown to match regular (styled by Fomantic-UI) input style (currently dropdown glow is based on open/close, not on (regular) focus)
c) review other outline: none/0 cases case by case

mvorisek:
Here is a default FF style obtained from `about:config` FF:
image

preview:
image

lubber-de:
### Update
I changed the PR to support custom outline and outline-offset styling for :focus-visible now.
Simply setting `@focusVisibleOutline: revert;` will apply the browsers default style on focus-visible everywhere.

#### Next steps/ todo:
- [ ] crossbrowser check the default styling on focus-visible (when reverting as mentioned above) for all the components and their variants
- [ ] adjust/apply a proper FUI like styling and change the defaults for `@focusVisibleOutline` and  `@focusVisibleOutlineOffset` accordingly in `site.variables`
- [ ] check combined variants (like `labeled input`) and add additional variables and selector for such stylings if needed
- [ ] create a custom theme to be able to keep using `none` for people still wanting the old behavior

ko2in (dismissed):
LGTM.

PR #2804, opened 2023-05-27
Fixes for positions of icon, text and upward sub-menu shadow and rounded border

ko2in:
## Description

This PR is a rebasing from my other [PR](https://github.com/fomantic/Fomantic-UI/pull/1838) which messed up with merging that resulted the dist directory included.

### Remove icon alignment
The remove icon wasn't aligned in the best position as it was shifted above comparing the text and dropdown icon.

**Before**
dropdown-remove-icon-not-aligned-properly

**After**
dropdown-remove-icon-aligend-properly

### Sub-menu position in floating menu item
The sub-menu of dropdown item was not positioned correctly when the dropdown has ```floating``` variant. It was overlapping it's menu item.

**Before**
floating-left-sub-menu-bug

**After**
left-floating-sub-menu

### Incorrect rounded corners and shadows of simple dropdown with upward direction
The rounded corners of upward simple dropdown menu should be top-left and top-right and the shadow should be displayed above the menu. That is fixed in this commit.

**Before**
upward-incorrect-shadow-and-rounded-border

**After**
upward-correct-shadow-and-rounded-border

## Testcase
https://jsfiddle.net/ko2in/4f3xh2sd/

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -187,6 +187,7 @@
 
 & when (@variationMenuVertical) {
     .ui.vertical.menu .item > .header:not(.ui) {
+        flex: 0 0 100%;
         margin: @verticalHeaderMargin;
         font-size: @verticalHeaderFontSize;
         font-weight: @verticalHeaderFontWeight;
@@ -200,7 +201,6 @@
 /* Dropdown Icon */
 .ui.menu .item > i.dropdown.icon {
     padding: 0;
-    float: @dropdownIconFloat;
     margin: 0 0 0 @dropdownIconDistance;
 }
 
@@ -211,6 +211,10 @@
     background: @dropdownBackground;
     margin: @dropdownMenuDistance 0 0;
     box-shadow: @dropdownMenuBoxShadow;
+}
+.ui.menu .upward.dropdown.item .menu {
+    border-radius: @dropdownMenuBorderRadius @dropdownMenuBorderRadius 0 0;
+    box-shadow: @dropdownUpwardMenuBoxShadow;
 }
 .ui.menu .dropdown.item:not(.column) .menu {
     flex-direction: column;
@@ -244,14 +248,12 @@
 }
 
 .ui.menu .ui.dropdown.item .menu .item:not(.filtered) {
-    display: block;
+    display: flex;
 }
 .ui.menu .ui.dropdown .menu > .item > .icons,
 .ui.menu .ui.dropdown .menu > .item > i.icon:not(.dropdown) {
-    display: inline-block;
     font-size: @dropdownItemIconFontSize !important;
-    float: @dropdownItemIconFloat;
-    margin: @dropdownItemIconMargin !important;
+    order: 0;
 }
 
 & when (@variationMenuSecondary) or (@variationMenuText) {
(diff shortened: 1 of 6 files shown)


src/definitions/collections/menu.less
```diff
@@ -297,6 +307,7 @@
     }
     .ui.vertical.menu .dropdown.item.upward .menu {
         bottom: 0;
+        top: auto !important;
```
lubber-de, line 310: Could !important get replaced by increasing .ui specificity?

src/definitions/modules/dropdown.less
```diff
@@ -1480,6 +1477,15 @@ select.ui.dropdown {
         margin-top: 0 !important;
     }
     & when (@variationDropdownUpward) {
+        .ui.simple.upward.active.dropdown > .menu,
+        .ui.simple.upward.dropdown:hover > .menu {
+            top: auto !important;
```
lubber-de, line 1482: Can !important be replaced by .ui specificity?

src/definitions/modules/dropdown.less
```diff
@@ -1480,6 +1477,15 @@ select.ui.dropdown {
...
+        }
+        .ui.simple.upward.dropdown > .menu > .item:active > .menu,
+        .ui.simple.upward.dropdown .menu .item:hover > .menu {
+            bottom: 0 !important;
+            top: auto !important;
```
lubber-de, line 1487: Can !important be replaced by .ui specificity?

src/definitions/modules/dropdown.less
```diff
@@ -1574,8 +1580,7 @@ select.ui.dropdown {
         Floating
     --------------- */
 
-    .ui.floating.dropdown > .menu {
-        left: 0;
+    .ui.floating.dropdown .menu {
```
lubber-de, line 1583: The direct child selector > was introduced by #2284
This change re-introduces that behavior again (see your jsfiddle)
image

lubber-de (changes requested):
The example from the previous PR does not work anymore. The submenu icon is misplaced and the text is not left aligned anymore

## Before
Your jsfiddle using FUI 2.9.2 without your changes
https://jsfiddle.net/lubber/8Ly7x2ju/1/
image

### After
Your code
https://jsfiddle.net/lubber/fs3yom7t/2/
image

I'll tag this as breaking change for 2.10.x as you removed/renamed LESS variables, 2.10.x will drop IE support and i think this needs more testing, especially because of the change to flexbox and order.

Please also see my other comments

src/definitions/modules/dropdown.less
ko2in: I don't see that in my fiddle and local version.

ko2in: Sorry. You're right. I was checking the dropdown without scrollbar.

src/definitions/collections/menu.less
ko2in: There are also ```!important``` flags that are not part of my PR which wouldn't allow to override with ```.ui``` specificity.

I would have to refactor and remove them too.

I feel like we've too many bloated CSS blocks in several modules that need to be removed.

Since we dropped the road map for FUI v3, we should consider to refactor this version to keep as much as clean and reduce the file size.

I would find the time for refactoring the CSS first. Then we can consider the refactoring JS modules.

ko2in: I couldn't find any solution unless we wrap the text inside dropdown item with extra container like `<span class="text">Item</span>`.

Please check this fiddle: https://jsfiddle.net/ko2in/twpkudxs/1/

As you can see, it just work without any CSS modifications.

lubber-de: > Should I discard this PR or keep it for another major release? What do you think @lubber-de ?

Let's keep it (as your work is still very good!), as planned for 2.10, and take a look again later. I label that on hold until then and convert to draft.

PR #2809, opened 2023-05-29
feat(embed): add optional alt tag to placeholder

hugopeek:
## Description

This is desirable for web use.

When no alt value is provided, the alt tag will not be shown (same as current behaviour).


src/definitions/modules/embed.js
```diff
@@ -288,6 +292,7 @@
                             .removeData(metadata.id)
                             .removeData(metadata.icon)
                             .removeData(metadata.placeholder)
+                            .removeData(metadata.alt)
```
lubber-de, line 295: This does not have any effect, because there is no call to `.data(metadata.alt, somevalue)` in the whole source and removeData does not remove an existing data-attribute (see https://api.jquery.com/removedata/)
Yes, the same issue exists for metadata.placeholder and metadata.icon (bug in original SUI) as those are also nowhere set via `.data()`

lubber-de (changes requested):
Thanks. Please see my comments. 🙂 
When done, please also adjust the docs 😉

lubber-de (approved):
I added my requested changes myself now and left everything else untouched.
So LGTM now

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/494

PR #2813, opened 2023-06-04
fix(checkbox): prevent readonly toggle via space

### LINKED ISSUE #2812: [checkbox] Read-only checkbox state must not be changeable from UI
# Bug Report

## Steps to reproduce
1. https://jsfiddle.net/4kuejywb/1/
2. click first checbox, press tab and then space key
3. notice, the second/read-only checkbox is changed

## Expected result

The state cannot be changed by mouse nor keyboard.

See https://stackoverflow.com/questions/155291/can-html-checkboxes-be-set-to-readonly for read-only "state"/"value" terminilogy.

Fomantic-UI docs - https://fomantic-ui.com/modules/checkbox.html#read-only - say "state" explicitly, thus I expect every checkbox/radio initialized like `$('.ui.checkbox').checkbox()` to be truly read-only.

## Version

latest

lubber-de:
## Description

Prevent a checkbox change via tab+space when it is declared as readonly

## Testcase
- click first checkbox
- press tab
- press space key

### Broken
- the second/read-only checkbox is changed
https://jsfiddle.net/4kuejywb/1/

### Fixed
- no checkbox change
https://jsfiddle.net/lubber/xud45evh/

## Closes
#2812


mvorisek (approved):

ko2in (approved):
LGTM.

Issue #1205, opened 2019-12-04
Fomantic-UI type definitions for Angular project


smartm0use:
Hello,
I'm looking for Fomanti-UI type definitions for an Angular project. I can't see Fomantic-UI from [here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types).

Is still valid to install typings for Semantic-UI? I mean `npm install --save @types/semantic-ui`

hammy2899:
@smartm0use Yeah just use the semantic types you mentioned for now, I will see what I can do for the modules we have added like calendar etc.

hammy2899:
Related to https://github.com/fomantic/Fomantic-UI/issues/396

smartm0use:
Do you know what is the right way to import Fomantic-UI (and related Semantic-UI typings) into an Angular project?

hammy2899:
I haven't used angular for years so no however you shouldn't need to import them like other packages if your `tsconfig` is setup correctly.

smartm0use:
In my `angular.json' file I have set the following:

```
"scripts": [
    "node_modules/jquery/dist/jquery.min.js",
    "node_modules/fomantic-ui/dist/semantic.min.js"
]
```

and in my `tsconfig.ts` file I have set this:

```
"types":[
    "jquery",
    "semantic-ui"
]
```

But the app fails to compile and I have this error in my IDE:

**TS2339: Property 'dropdown' does not exist on type 'JQuery<HTMLElement>'.**

hammy2899:
Try and remove the `types` property from your `tsconfig` because typescript will automatically pickup any types installed from the `@types` org.

smartm0use:
Nothins has changed. Maybe the problem is that Fomantic-Ui isn't matching with Semantic-UI types?

hammy2899:
Can you post a snippet of code where you use fui?

smartm0use:
```
import {Component, OnInit, ViewEncapsulation} from '@angular/core';
import {ActivatedRoute, Router} from '@angular/router';

@Component({
    selector: 'search-users',
    templateUrl: './search-users.component.html',
    styleUrls: ['./search-users.component.scss'],
    encapsulation: ViewEncapsulation.None
})
export class SearchUsersComponent implements OnInit {
    users: Array<any>;

    constructor(private router: Router, private activatedRoute: ActivatedRoute) {
        // Some code...
    }

    ngOnInit() {
[...]

PR #2822, opened 2023-06-15
feat(chore): added types definitions

### LINKED ISSUE #396: Type definitions
Are there any type definition files associated with it or do we need to use the existing type defs  from @types/semantic-ui

prudho:
Long time overdue, sorry about that !

## Description
This PR adds type definitions for Fomantic's components, enabling them to be more usable on code editors. Typings are now integrated into the actual repository, rather than in the DefinitelyTyped repository, allowing us a better control over type updates and quick fixes. They've been updated to the actual 2.9.2 release, so they should be pretty accurate.

To use them, there's nothing to do since they're included in the `package.json`. Users should also include `jquery types` as well in their package.json: `npm install --save-dev @types/jquery`

There's maybe some comments to add, but it can be released now.

@fomantic/maintainers should I also create a pull request at https://github.com/fomantic/Fomantic-UI-CSS ?

## Screenshot
image

image

## Closes
#396, #1205


lubber-de (approved):
LGTM

prudho:
The DefinitelyTyped repo can be deleted, right.

For the Fomantic-UI-CSS and Fomantic-UI-LESS repo, yes, we just need to add the `types` folder and add the line in `package.json`.

mvorisek:
The current JS files should be tested againts the TS interfaces and any missing/extra method/property should fail CI.

lubber-de:
> All JS files should be tested againts the TS interfaces and any missing/extra method/property should fail CI.

That would be worth another PR as it does not seem to be easy. The whole special "behavior" concept (string as first parameter of module call magically matches existing (nested) javascript method tree) would basically work for each existing method inside the modules, but  only the "intended" methods are listed in the docs (and only those were covered by prudho here)

A CI, as you suggested, would complain about a lot of unfetched methods as those are not intended to be mentioned.
Of course, it would be consistent to document all of them, even if possibly just used internally by the module, but this would also need a ton of missing documentation for each function. As always, volunteers welcome 😉 

I believe this only works out nicely when recoding each module from scratch with a different code structure concept and modern techniques where types and docs can be created/fetched/generated automatically out of the source without the need to manually touch them. 

So, I'll merge this PR as it is now and will look for possibly missing new settings in 2.9.3 as this PR "only" covers 2.9.2 which already must have been a hell of work.

prudho:
@mvorisek yes, this is something i'd like to implement too. A real testing system for each function/property would be great, but like @lubber-de said, this would be a rather large project to test basically everything.

For now, and since we're using an outdated code base, all this work can be done manually, and trust me, there's A LOT of work to do so.

lubber-de:
> The DefinitelyTyped repo can be deleted, right.

@prudho As discussed, I just deleted the DefinitelyType fomantic fork

PR #2838, opened 2023-06-26
feat(types): better readability, intellisense improvements, small fixes

prudho:
## Description
This PR adds small improvements as well as a few quickfixes to the first types iteration:
- Some properties and behaviors where missing for `calendar`, `api`, `modal`, `search` and `toast`.
- The DefititelyTyped link has been removed from `index.d.ts.`

I've also rewritten modules definitions to be more readable, extensible, and to get rid of this awful `_Impl` interface system (new versions of Typescript are way more flexible for that). As the cherry on the top, Intellisense seems to better undertstand definitions and autocompletion works better than before 🎉 

## Screenshot
Before:

Overload issues !
image

Intellisense not working very well...
image

After:

Type checking working for each property, even nested !
image

Intellisense is happy !
image


lubber-de (approved):
LGTM

PR #2841, opened 2023-06-27
fix(types): allow JQueryAjaxSettings objects for modules using apiSettings

### LINKED ISSUE #2840: [Types] APISettings Should Supports JQueryAjaxSettings?
# Bug Report
I see a new implement in [this commit](https://github.com/fomantic/Fomantic-UI/commit/eabc58e7d3526c2073c60054cd6d05546e9c4200) for ``types`` in the lib. Yay! 

However when I do this:
```ts
jq("#test").search({
    ...
    apiSettings: {
        action: ...
        timeout: 8000, //<-- GOT ERROR
        error(error: function(jqXHR, textStatus, errorThrown)) {} //<-- GOT ERROR
    }
});
```
Those ``error keys`` are from JQueryAjaxSettings.
Should the APISettings include the JQueryAjaxSettings as well?, based on this doc:
https://server3.kproxy.com/servlet/redirect.srv/sruj/sre-ovhcqdgy/p2/behaviors/api.html#/settings

> You can pass in any standard [jQuery AJAX setting](https://server3.kproxy.com/servlet/redirect.srv/sruj/shbpgdx/sygs/p2/jquery.ajax/) like timeout or contentType to API's settings and it will be automatically passed to the request's AJAX call.

Correct me if I am wrong, thanks!

## Version
2.9.2-nightly

prudho:
## Description
Following Fomantic's documentation, `apiSettings` can also accept a `JQueryAjaxSettings` object as parameter. So this PR enable correct type checking for modules that includes `apiSettings` as a parameter: `dropdown`, `search` and `tab`.

## Screenshot
Before:
before

After:
after

## Closes
#2840

### DIFF
--- a/types/fomantic-ui-api.d.ts
+++ b/types/fomantic-ui-api.d.ts
@@ -107,10 +107,10 @@ declare namespace FomanticUI {
          */
         (behavior: 'destroy'): JQuery;
 
-        <K extends keyof AccordionSettings>(behavior: 'setting', name: K, value?: undefined, ): Partial<Pick<AccordionSettings, keyof AccordionSettings>>;
-        <K extends keyof AccordionSettings>(behavior: 'setting', name: K, value: AccordionSettings[K]): JQuery;
-        (behavior: 'setting', value: Partial<Pick<AccordionSettings, keyof AccordionSettings>>): JQuery;
-        (settings?: Partial<Pick<AccordionSettings, keyof AccordionSettings>>): JQuery;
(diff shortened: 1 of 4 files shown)


lubber-de:
That was quick! 👍🏼 
What about the api module itself? Doesn't this also need an adjustment to allow the standard jquery ajax parameters?

lubber-de (approved):
LGTM

prudho:
Indeed ! I've added them as well (it's tricky but should works), also fixed a bad naming...

PR #2865, opened 2023-07-31
fix(dropdown): remove clearable icon on disabeled or readonly

### LINKED ISSUE #2862: [dropdown] the cross for disabled dropdown select items is misleading
# Bug Report
`readonly` and `disabled` dropdown still show crosses for the selected items, and to clear the input. 
I would have expected those crosses to not appear, as they suggest users that they can interact with the dropdown.
I would suggest no displaying them.

## Testcase
https://jsfiddle.net/mfowvc2z/

## Screenshot (if possible)
Screenshot 2023-07-28 at 19-01-03 Fomantic-UI Sandbox - JSFiddle - Code Playground

## Version

2.9.2

lubber-de:
## Description

A `disabled` or `read-only` dropdown used in combination with clearable or multiple labels still showed the remove/delete icon which could indicate to still being able to delete the related value

## Testcase
Remove CSS to issue
https://jsfiddle.net/lubber/18okh92g/5/

### Before
image

### After
image

## Closes
#2862

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1186,7 +1186,7 @@ select.ui.dropdown {
     .ui.read-only.dropdown {
         > .remove.icon,
         > .label > .delete.icon {
-            display:none;
+            display: none;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

PR #2866, opened 2023-08-01
feat(calendar): add size variations

prudho:
## Description
This PR adds classic size variations for the calendar module. I've tested it and so far didn't found any issue, but i'm a little bit rusted.

## Testcase
[JSFiddle](https://jsfiddle.net/uokxnbc6/)

## Screenshot
image

### DIFF
--- a/src/definitions/modules/calendar.less
+++ b/src/definitions/modules/calendar.less
@@ -196,6 +196,35 @@
     }
 }
 
+/* --------------------
+        Sizes
+--------------------- */
+
+& when not (@variationCalendarSizes = false) {
+    each(@variationCalendarSizes, {
+        @s: @@value;
+        .ui.@{value}.calendar,
+        .ui.@{value}.calendar .ui.table tr th,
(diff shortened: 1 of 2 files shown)


lubber-de (commented):
Basically LGTM 👍🏼 

I just now recognized that the calendar.less has the default min-width hardcoded.
Would you like to clean this up and use variables for the 15,18,22,20 values also used here
https://github.com/fomantic/Fomantic-UI/blob/09d0671c23fa2710652efaaf0f716aaae85bf42f/src/definitions/modules/calendar.less#L63-L77

?

Otherwise we can just merge it and clean up by a separate PR (i also found some hardcoded padding values as well)

prudho:
Let's merge this one as it for now, I still have to write the doc

prudho:
Doc added by  fomantic/Fomantic-UI-Docs#471.

I think that we can merge as it for now and clean it in a separate PR.

lubber-de (approved):
LGTM!

PR #2867, opened 2023-08-01
feat(calendar): move hardcoded values to less variables

lubber-de:
## Description

Moves some hardcoded values into LESS variables as discussed in https://github.com/fomantic/Fomantic-UI/pull/2866#pullrequestreview-1556622825


prudho (approved):
LGTM 👍

ko2in (commented):
LGTM.

PR #2869, opened 2023-08-02
feat(tab): add removable variation

### LINKED ISSUE #2521: [tab] Closable tabs and dynamically adding new ones
# Feature Request
Have a way to allow tabs to be closed and possibly also have a way to add new tabs in (via JS)

Developer may want to have some kind of management panel, for example, support tickets, where each tab is a ticket being worked on. To view a ticket, a new tab could be added to the top of a list where the content is dynamically loaded. Once the tab is finished with, it can be closed as need be.

## Example (if possible)

IMG_1439

prudho:
## Description
This PR add the ability to remove a tab by clicking on an icon or programmaticaly. It works by adding the `removable` class to the desired tab headers or setting `removable: true` on initialization. Two new callbacks are implemented: `onRemove(tabPath)` when the tab is removed, and `onBeforeRemove(tabPath)`, which allow cancelling the removal of the table by returning `false`. When removing a tab, it will move to the closest which is not disabled.

For now i'll mark it as WIP, since there's some tests to do on nested tabs, and decide what to do if all tabs are closed.

## Screenshot
image

## Closes
#2521

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -240,7 +240,7 @@
                             ;
                             module.removeTab(tabPath);
                             event.stopPropagation();
-                        }
+                        },
                     },
                     history: {
                         change: function (event) {
@@ -450,6 +450,7 @@
                         module.verbose('Call for tab removal', tabPath);
                         if (settings.onBeforeRemove.call(element, tabPath) === false) {
                             module.debug('onBeforeRemove returned false, cancelling tab removal', $tab);
+                            
                             return false;
                         }
                         var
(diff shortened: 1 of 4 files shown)


lubber-de (commented):
Quick thoughts
- What about a lonely "+" icon inside a new (unbound/fake) tab to also add a new tab (optional). Would also need two new events "onAdd" and "onBeforeAdd" just like you implemented for remove

> what to do if all tabs are closed.

 I think it should be an (optional) restriction to keep at least 1 tab (or that setting can define how many tabs have to stay as minimum)

prudho:
> What about a lonely "+" icon inside a new (unbound/fake) tab to also add a new tab (optional). Would also need two new events "onAdd" and "onBeforeAdd" just like you implemented for remove

I'll try that on another PR !

PR #2878, opened 2023-08-16
fix(slider): max value had js floating point issues

### LINKED ISSUE #2877: [slider] Range slider Value does not go to max when step < 1
# Bug Report
When I initialize a Slider with min 0, max 0.94, start 0.94 and step 0.01, I expect the Value to be 0.94 at the start, it is however roughly 0.93

## Expected result
0.94

## Actual result
0.9299999999999999

## Testcase
[Testcase](https://jsfiddle.net/4robkn7u/1/)

## Version
2.9.2

lubber-de:
## Description
Sometimes a calculation of max value or number of labels went into JS floating point issues why dividing

### Wrong Result
Max value is determined as 0.92999999 instead of 0.94
https://jsfiddle.net/4robkn7u/1/

### Fixed
Max value is 0.94 as expected
https://jsfiddle.net/lubber/qbs3cn9e/1/

## Closes
#2877

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -674,7 +674,8 @@
                         var
                             step = module.get.step(),
                             min = module.get.min(),
-                            quotient = step === 0 ? 0 : Math.floor((settings.max - min) / step),
+                            precision = module.get.precision(),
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

PR #2879, opened 2023-08-19
fix(accordion): callbacks for closing exclusive accordions

### LINKED ISSUE #2826: [accordion] `exclusive` accordions should call their callbacks for both the opening and the closing element
# Feature Request

For accordions with the `exclusive` setting set to `true`, when one element is already open and the user opens another element (so that the first one automatically closes), the callbacks in [the settings section of the docs](https://fomantic-ui.com/modules/accordion.html#/settings) are only called for the element that is opened, and not the one that is closed (as of FUI v2.9.2).

It would be useful to have the callbacks called for both affected elements, so that the developer does not have to iterate over the other elements to find which one was affected, in case they want to do something with it. _(In my case - which motivated opening this issue - it's to set the `inert` attribute on closing accordion `.content`, to prevent keyboard focus while the closing animation is playing.)_

## Example (if possible)

If the explanation above is not entirely clear, do post a comment, and I can create an example 🙂

lubber-de:
## Description
- Exclusive accordions only triggered the callback for the opening according but not for all the closing ones.
- Also added the ability to cancel any open/close by returning `false` from the onOpening/onclosing/onChanging callbacks

## Testcase
https://jsfiddle.net/lubber/dmqpnLtu/9/

## Closes
#2826


ko2in (commented):
Shouldn't ```closeOthers``` reflect the ```onClosing``` callback? What if I want to keep opening the accordion item that was already opened? Now, it only listens to the accordion that triggered the click event.

lubber-de:
I put this on hold for now, as, after some investigation, the current logic would need more refactoring to support that properly and be consistent as "closeothers" behaves differently than just closing the active accordion.

PR #2883, opened 2023-08-22
fix(dropdown): support selection floating or pointing

lubber-de:
## Description
When `pointing` or `floating` was used
- background color overlayed the border corners of first and last item if menu was not scrolled/overflowed
- in combination with `selection` borders were missing
- inverted upward pointing had wrong arrow direction

I also added a new `unlimited` variant which does not make the selection menu scrollable but shows all items at once instead. This is especially needed in combination with `pointing`. Otherwise the arrow wont be displayed as the arrow is part of the (overflowing) dom node.

## Testcase
https://jsfiddle.net/lubber/jq3vmLde/7/

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -435,19 +435,23 @@ select.ui.dropdown {
     }
 
     /* Selection Menu */
-    .ui.selection.dropdown .menu {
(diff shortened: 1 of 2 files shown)


lubber-de:
Docs regarding the new unlimited feature added by https://github.com/fomantic/Fomantic-UI-Docs/pull/484

PR #2895, opened 2023-09-01
feat(site): separate variable for whole google Font url

### LINKED ISSUE #2892: [font] Multiple font support
# Feature Request

Currently FUI supports one font, by default Lato. However would be great if it supported multiple without having to resort to some "gymnastics". Our main case - page font and header font are different.

lubber-de:
## Description

This PR simplifies a possible need to import multiple google fonts, so one only needs to change 1 variable `@googleFonts`
I also put the import url into a separate variable to even the whole url could be changed by only one variable

The change is completely backward compatible

## Testcase
This only variable change inside a custom theme will import 3 fonts
```less
@googleFonts: "Tangerine|Inconsolata:italic|Droid+Sans:bold";
```
and, by keeping every other variable as default, would result in
```css
@import url("https://fonts.googleapis.com/css2?family=Tangerine|Inconsolata:italic|Droid+Sans:bold&subset=latin&display=swap");
```

## Closes
#2892

### DIFF
--- a/src/theme.less
+++ b/src/theme.less
@@ -63,7 +63,7 @@
 
 .loadFonts() {
     & when (@importGoogleFonts) {
-        @import (css) url("@{googleProtocol}fonts.googleapis.com/css2?family=@{googleFontRequest}");
+        @import (css) url("@{googleFontUrl}");
     }
     & when (@importFonts) and not (@fontName = "") {
         each(@fonts, {
--- a/src/themes/default/globals/site.variables
+++ b/src/themes/default/globals/site.variables
@@ -124,6 +124,7 @@
 
 @googleProtocol: "https://";
 @googleFontRequest: "@{googleFontName}:@{googleFontSizes}&subset=@{googleSubset}&display=@{googleFontDisplay}";
+@googleFontUrl: "@{googleProtocol}fonts.googleapis.com/css2?family=@{googleFontRequest}";
 
 @bold: bold;
 @normal: normal;
--- a/src/themes/default/globals/site.variables
+++ b/src/themes/default/globals/site.variables
@@ -123,7 +123,8 @@
 @googleFontDisplay: "swap";
 
 @googleProtocol: "https://";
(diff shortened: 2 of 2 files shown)


TivoSoho:
Note that 
```
@import url("https://fonts.googleapis.com/css2?family=Tangerine|Inconsolata:italic|Droid+Sans:bold&subset=latin&display=swap");
```
doesn't work - it needs to be `css`, not `css2`

References; 
https://developers.google.com/fonts/docs/css2
https://developers.google.com/fonts/docs/getting_started#specifying_font_families_and_styles_in_a_stylesheet_url

lubber-de:
Thanks for that info. Well, then one has to either

- adjust the `@googleFonts` variable accordingly (yes, the x-times "&family=" has to be included aswell. A bit ugly, i know, but there is no real easy way to concatenate a string in LESS inside a loop (which would be needed otherwise to automatically add the "family=" GET parameter automatically).
So if multiple googleFonts are needed, there is still the possibility to just change a single variable to accomplish this task more easily

```less
@googleFonts: "Crimson+Pro&family=Literata";
```

AND/OR
- adjust the `@googleFontUrl` variable accordingly
```less
@googleFontUrl: "@{googleProtocol}fonts.googleapis.com/css?family=@{googleFontRequest}";
// or
@googleFontUrl: "@{googleProtocol}fonts.googleapis.com/css?family=@{googleFonts";

// or even completely hardcoded
@googleFontUrl: "https://fonts.googleapis.com/css2?family=Crimson+Pro&family=Literata";
@googleFontUrl: "https://fonts.googleapis.com/css?family=Crimson+Pro:ital|Literata:bold";
```

I don't think that creating even more variables (like `@googleFontVersion: 'css2';` makes sense, it just gets more bloated (and even more complex/confusing)
By the this change you are now completely free to adjust the google font import by the change of a single variable out of some existing ones

ko2in (approved):
LGTM.

PR #2903, opened 2023-09-13
fix(dropdown): create label before changing value as in 2.8.8

### LINKED ISSUE #2902: [dropdown] onchange event on dropdown multiple is trigger differently from 28.8 to 29.3
# Bug Report
In our source code we do this on the onchange of a multiple dropdown.

```
$(document).ready(function() {
  $('.ui.dropdown').dropdown({
    onChange: function(value, text, $selectedItem) {
      $(this).find('.label').each(function() {
        let item = $(this).attr('data-value');
        console.log(item);
      })
    }
  });
});

```
When we add a value in 28.8 every element was correctly printed.
https://jsfiddle.net/gtoniolo/w3850tfn/7/

In 29.3 the just added element is not printed.
https://jsfiddle.net/gtoniolo/w3850tfn/1/

I do not find any clue in documentation. What's changed?

lubber-de:
## Description
This PR reverts the order change of add.value and add.label introduced in #1854
There was no reason doing so.

But this change now had an impact when trying to access the labels inside the onchange callback as the (new) labels are now not available yet when onChange is called (as it was unil 2.8.8) 

## Testcase
- Open console
- Select another language from the dropdown
### Broken
-> The new one is not logged
https://jsfiddle.net/gtoniolo/w3850tfn/1/

### Fixed
- The new one is logged
https://jsfiddle.net/lubber/5r0dv9ha/

## Closes
#2902

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2806,8 +2806,8 @@
                                             module.save.remoteData(selectedText, selectedValue);
                                         }
                                         if (settings.useLabels) {
-                                            module.add.value(selectedValue, selectedText, $selected, preventChangeTrigger);
                                             module.add.label(selectedValue, selectedText, shouldAnimate);
+                                            module.add.value(selectedValue, selectedText, $selected, preventChangeTrigger);
                                             module.set.activeItem($selected);
                                             module.filterActive();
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

PR #2909, opened 2023-09-21
feat(header): seamless attached variant

### LINKED ISSUE #2874: [header] doesn't support seemless attached variant
# Bug Report
Headers don't support `seemless` variants like attached segments do

## Steps to reproduce
```html
<div class="ui seamless top attached segment">
  <h3 class="ui header">
    Works!
  </h3>
</div>
<div class="ui seamless bottom attached segment">
  foo bar
</div>

<h3 class="ui seamless top attached header">
  Doesn't work!
</h3>
<div class="ui seamless bottom attached segment">
  foo bar
</div>
```

## Expected result
It's a nice feature for segments, it should work for headers too

## Actual result
https://jsfiddle.net/jkszwbxe/

## Screenshot (if possible)
image

## Version

2.9.2

ko2in:
## Description
The attached header doesn't support the seamless variant like the segment element which results in showing the border between the header and the attached element below.

This PR allows to have the attached header element with seamless variant that allows to remove the bottom border when it attaches with the another segment below.

## Testcase
**Before:** https://jsfiddle.net/jkszwbxe/

**After:** https://jsfiddle.net/ko2in/yoafg8mx/1/

## Screenshot

**Before:**
seamless-attached-header-not-supported

**After:**
seamless-attached-header-supported

## Closes
#2874

### DIFF
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -467,6 +467,11 @@
     .ui.bottom.attached.header {
         border-radius: 0 0 @attachedBorderRadius @attachedBorderRadius;
     }
+    & when (@variationHeaderSeamless) {
+        .ui.seamless.attached.header {
+            border-bottom: none;
+        }
+    }
 
     /* Attached Sizes */
     .ui.attached.header:not(h1):not(h2):not(h3):not(h4):not(h5):not(h6) {
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -468,7 +468,7 @@
         border-radius: 0 0 @attachedBorderRadius @attachedBorderRadius;
     }
     & when (@variationHeaderSeamless) {
-        .ui.seamless.attached.header {
+        .ui.seamless.attached:not(.bottom).header {
             border-bottom: none;
         }
     }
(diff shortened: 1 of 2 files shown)


src/definitions/elements/header.less
```diff
@@ -467,6 +467,11 @@
     .ui.bottom.attached.header {
         border-radius: 0 0 @attachedBorderRadius @attachedBorderRadius;
     }
+    & when (@variationHeaderSeamless) {
+        .ui.seamless.attached.header {
```
lubber-de, line 471: This would also remove the bottom border on a `seamless bottom attached header`. See https://jsfiddle.net/znLf2exm/

So please add `:not(.bottom)`
```suggestion
        .ui.seamless.attached:not(.bottom).header {
```

lubber-de (changes requested):
Just a little change request 😉 Please see comment and suggestion

src/definitions/elements/header.less
ko2in: Done :+1:

lubber-de (commented):
LGTM

lubber-de (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/478

PR #2911, opened 2023-09-22
feat(form, input): support textarea with action button

### LINKED ISSUE #2814: [textarea] Textarea with action should be styled as close as possible as input
# Feature Request

`textarea` is basically an `input` that support multiple lines. Therefore I expect the same features as for input. The css should be generated for input/textarea as much as possible using deduplicated code/less.

Currently there are at least there differences:

a) the width is not the same for the same markup (input has `width: auto`)

b) textarea border is not "unrounded" for right (probably for left too) action
   - and the action height should be probably limited to standard action height as with regular one-line input

## Repro

https://jsfiddle.net/vhfrb2qt/1/

## Screenshot (if possible)

image

ko2in:
## Description
Supports textarea in the action input to display like the input field and the button stacked together.

## Testcase
**Before:** https://jsfiddle.net/vhfrb2qt/1/

**After:** https://jsfiddle.net/ko2in/5aco9q8m/

### Before
textarea-before

### After
textarea-after

## Closes
#2814

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -441,7 +441,8 @@
         input[type="text"]:focus,
         input[type="file"]:focus,
         input[type="url"]:focus,
-        input[type="week"]:focus {
+        input[type="week"]:focus,
+        textarea:focus {
             border-top-right-radius: 0;
             border-bottom-right-radius: 0;
         }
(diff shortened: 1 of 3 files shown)


ko2in:
Made some adjustments and fixes. Here's the updated fiddle https://jsfiddle.net/ko2in/bgLfuvje/

lubber-de (approved):
LGTM! Great job fixing the width issues!

PR #2914, opened 2023-09-25
fix(types): better declaration files

### LINKED ISSUE #2906: [types] No Overload Error ``ts(2769)``, Conversion of Type Error ``ts(2352)`` and More...
# Bug Report
My VSCode is yelling a lot of types errors when using the new types definition of FomanticUI.

## Steps to reproduce
Example:
```ts
import 'fomantic-ui';

//...

jq("#el").nag({persist: true});
```
Get this error:
```error
No overload matches this call.
  The last overload gave the following error.
    Argument of type '{ persist: true; }' is not assignable to parameter of type 'NagSettings'.
      Type '{ persist: true; }' is missing the following properties from type 'NagSettings': displayTime, context, detachable, expires, and 24 more.ts(2769)
```

When I do this then it won't yell anymore:
```ts
jq("#server-down-msg").nag({persist: true} as FomanticUI.NagSettings);
```

But with type assertion on calendar for example:
```ts
    jQuery(startCalendarEl).calendar({
        today: true,
        initialDate: moment().startOf("day"),
        endCalendar: jQuery(endCalendarEl)
    } as FomanticUI.Calendar);
```

This would yell:
```error
Conversion of type '{ today: boolean; initialDate: moment.Moment; endCalendar: JQuery<HTMLElement>; }' to type 'Calendar' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'settings' is missing in type '{ today: boolean; initialDate: moment.Moment; endCalendar: JQuery<HTMLElement>; }' but required in type 'Calendar'.ts(2352)
```

This is just a few examples of some scopes, there are more...

**NOTED: the application is still running fine, only VSCode that giving me a lot of ``red lines`` which is anoying.

## Expected result
No VSCode linting types errors.

## Version

2.9.3

prudho:
## Description
This is the first round of fixes/improvements for type declarations of Fomantic. Some properties and methods were missing, some were too restrictive.

I've enabled strict type checking on my work project (~500k lines of code and use almost every module of Fomantic). Before the fix I had around 2500 errors during transpilation, 0 after.

I just cannot fix one error, that I circumvent by adding the `//@ts-expect-error` on the line before:
For a `Form`, creating a new rule will raise an error in VSCode (that can be safely bypassed).

image

`$.fn.form.settings.rules` is typed and Intellisense is working, but adding a new rule to it will trigger an error. Anyway, this is not a blocking issue for merging, since this PR will make a lot more good than bad :)

## Closes
#2906

### DIFF
--- a/types/fomantic-ui-api.d.ts
+++ b/types/fomantic-ui-api.d.ts
@@ -143,7 +143,7 @@ declare namespace FomanticUI {
          * UI state will be applied to this element, defaults to triggering element.
          * @default false
          */
-        stateContext: false | string | JQuery;
+        stateContext: false | string | JQuery<any>;
 
         /**
          * Whether to encode parameters with 'encodeURIComponent' before adding into url string.
--- a/types/fomantic-ui-api.d.ts
+++ b/types/fomantic-ui-api.d.ts
@@ -126,6 +126,12 @@ declare namespace FomanticUI {
         on: string;
 
         /**
+         * Object containing all templates endpoints
+         * @default {}
+         */
+        api: {[key: string]: string};
+
+        /**
          * Can be set to 'local' to cache successful returned AJAX responses when using a JSON API.
          * This helps avoid server roundtrips when API endpoints will return the same results when accessed repeatedly.
          * Setting to 'false', will add cache busting parameters to the URL.
@@ -137,7 +143,7 @@ declare namespace FomanticUI {
          * UI state will be applied to this element, defaults to triggering element.
          * @default false
          */
-        stateContext: false | JQuery;
+        stateContext: false | string | JQuery;
 
         /**
          * Whether to encode parameters with 'encodeURIComponent' before adding into url string.
@@ -259,7 +265,7 @@ declare namespace FomanticUI {
          * Method for transmitting request to server.
          * @default 'get'
          */
-        method: 'get' | 'post' | 'put' | 'delete' | 'head' | 'options' | 'patch';
+        method: Uppercase<'get' | 'post' | 'put' | 'delete' | 'head' | 'options' | 'patch'> | Lowercase<'get' | 'post' | 'put' | 'delete' | 'head' | 'options' | 'patch'>;
 
         /**
          * Expected data type of response.
--- a/types/fomantic-ui-calendar.d.ts
+++ b/types/fomantic-ui-calendar.d.ts
@@ -37,7 +37,7 @@ declare namespace FomanticUI {
          * Pass false to updateInput to disable updating the input.
          * Pass false to fireChange to disable the onBeforeChange and onChange callbacks for this change
          */
-        (behavior: 'set date', date: string, updateInput: boolean, fireChange: boolean): JQuery;
+        (behavior: 'set date', date: Date | string | null, updateInput?: boolean, fireChange?: boolean): JQuery;
 
         /**
          * Get the current selection mode (year, month, day, hour, minute)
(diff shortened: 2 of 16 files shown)


lubber-de (dismissed):
LGTM
That must have been a PITA to investigate and fix.
Thanks for your effort! 👍🏼

KiddoV:
Also ``onSuccess`` with ``fields`` should be ``{[key: string]: any}`` instead of ``object[]``?
https://github.com/prudho/Fomantic-UI/blob/17cad4ed74a71c16d1e3b05f19b99e24bd8aff0b/types/fomantic-ui-form.d.ts#L313

KiddoV:
> I just cannot fix one error, that I circumvent by adding the `//@ts-expect-error` on the line before: For a `Form`, creating a new rule will raise an error in VSCode (that can be safely bypassed).
> 
> image
> 
> `$.fn.form.settings.rules` is typed and Intellisense is working, but adding a new rule to it will trigger an error. Anyway, this is not a blocking issue for merging, since this PR will make a lot more good than bad :)

For this I would suggest doing this:
Add this to line [294](https://github.com/prudho/Fomantic-UI/blob/17cad4ed74a71c16d1e3b05f19b99e24bd8aff0b/types/fomantic-ui-form.d.ts#L294):
```ts
rules: Form.RulesSettings & {[key: string]: (value?: any, identifier?: string, module?: any) => boolean};
```
So we can add custom rule but keep the original
Also add this for custom prompt:
```ts
prompt: Form.PromptSettings & {[key: string]: string | undefined};
```
To line number [282](https://github.com/prudho/Fomantic-UI/blob/17cad4ed74a71c16d1e3b05f19b99e24bd8aff0b/types/fomantic-ui-form.d.ts#L282)

lubber-de:
@KiddoV 
This would also be needed for Modal / Flyout templates as those can be enhanced by custom templates as well

@prudho Would you like to investigate into the above mentioned findings and add commits to this PR before merging?

KiddoV:
Please check out my PR #2915 for reference. I fixed almost all of errors from my application based on @prudho PR and some additional fixes from me.
NOTED: There could be more error that I haven't spotted.

prudho:
> > I just cannot fix one error, that I circumvent by adding the `//@ts-expect-error` on the line before: For a `Form`, creating a new rule will raise an error in VSCode (that can be safely bypassed).
> > image
> > `$.fn.form.settings.rules` is typed and Intellisense is working, but adding a new rule to it will trigger an error. Anyway, this is not a blocking issue for merging, since this PR will make a lot more good than bad :)
> 
> For this error I would suggest doing this: Add this to line [294](https://github.com/prudho/Fomantic-UI/blob/17cad4ed74a71c16d1e3b05f19b99e24bd8aff0b/types/fomantic-ui-form.d.ts#L294):
> 
> ```ts
> rules: Form.RulesSettings & {[key: string]: (value?: any, identifier?: string, module?: any) => boolean};
> ```
> 
> So we can add custom rule but keep the original Also add this for custom prompt:
> 
> ```ts
> prompt: Form.PromptSettings & {[key: string]: string | undefined};
> ```
> 
> To line number [282](https://github.com/prudho/Fomantic-UI/blob/17cad4ed74a71c16d1e3b05f19b99e24bd8aff0b/types/fomantic-ui-form.d.ts#L282)

This is clever ! I tried a lot of stuff to make it work but your solution works great and doesn't break Intellisense 👍

I've pushed another bunch of fixes based on your feedback, many thanks @KiddoV !

@lubber-de yeah, let's wait a few more fixes before merging it.

prudho:
Definitely many thanks @KiddoV. All the difficulty here is to set "probable" types to a framework that can accept virtually anything 😄 Your help is very appreciated !

I've implemented your suggested changes, and added a few more by reviewing code and internal uses.

KiddoV:
@prudho Thank you! Your latest commit make my VSCode happy now, so I am cool! 😁
This ``types`` feature still definitely need a lot more works. I will give more suggestions if spotting anything else!

lubber-de (approved):
LGTM! 👍🏼

PR #2917, opened 2023-09-28
feat(step): circular steps variant

### LINKED ISSUE #2170: [steps] circular steps
# Feature Request
Rather than having the “box-like” steps as currently available, how about we add a circular variant that gives just a general overview (no header/description)? Numbers could potentially be used within the circular label or just use solid colours or ticks; but still have disabled options for incomplete steps, etc

A vertical variant could still show headers/descriptions

## Example
Steps - Official Tailwind CSS UI Components

Steps - Official Tailwind CSS UI Components 2

ko2in:
## Description
This PR add the new variant `railing` to the steps which suppose to display the steps with circular railing horizontally or vertically.

I used the name `raliing` since we already have `rail` component which can conflict each other. If you have better name for this variant, please suggest.

## Testcase
https://jsfiddle.net/ko2in/n9dum7fg/

## Screenshot (if possible)

fui-railing-step

## Closes
#2170


ko2in:
I had to fix stylelint errors and rebase to take out this commit, and done forced push. Is that be OK?

lubber-de:
Great! But why dont we just name it `circular` as we already have for image, icon, label and button?

lubber-de (commented):
The inner circle isnt centered, looks misplaced

image

lubber-de (commented):
The `inverted` variant is not properly supported
image

Using any other background than white would need to change the non completed icon background as well. I know you set white as background to cover the connection line (i had a similar issue for `connected` feed)
image

lubber-de:
The connection line cover background could be avoided by

```css
.ui.railing.vertical.steps .step:not(:last-child) .content::before {
margin-top: 2.2em;
}
```
it seems....

I wonder why i had so many issues doing that for connected feed... perhaps more difficult markup, Will take a look for that again

ko2in:
> Great! But why dont we just name it `circular` as we already have for image, icon, label and button?

That sound also OK for me. I will update the PR with that name and other fixes.

lubber-de:
> I didn't implement for inverted yet. I don't know how the inverted should be at the moment.

Not much to be done i believe. I suggest to:
- use dedicated less variables (yes, even if those would have the same value) for the inverted variant
- increase specificity, so the normal inverted variant does not override the railing ones (would already fix the inner circle color from black to green)
- dont use white as background (perhaps no background at all/transparent in case my margin-top idea works for you would alrady fix that)
- use a light color for the connection lines
- use the default inverted green colors

jamessampford:
Nice! Would it be possible to have an option to have a title and/or description on the horizontal variant below the step circles?

lubber-de (commented):
The `ordered` variant is totally messed up unfortunately

image

lubber-de (changes requested):
The size variants need adjustments at least for vertical variant

tiny steps
- check icon too low, but circle icon for uncompleted looks good now :D
image

massive steps
- both icons misplaced
- text for content title and description needs some padding (this is also the case for the normal step without your new variant, i just stumbled upon that) 
image

lubber-de (commented):
The `link step` variant isn't supported

image

lubber-de (commented):
The new variant seems to be always unstackable. This could be a design choice and perhaps supported later if a visually pleasing solution can be found

lubber-de (commented):
my final comment/suggestion: 😉 

`disabled step` does not have any visual effect (you removed that on purpose...why?)

We should support color names like we have for the `connected feed` variant.
Currently the green color is hardcoded (`@completedColor`). The green was supposed to work for the usual step variant where coloring would not make so much sense to me as it would only have effect on the check icon.

But as your new variant concentrates on the icon and connection line there is a possible need for a different color (maybe only for a single completed step) and screams for being colorizable via a color class name.

Support coloring as we have it elsewhere would support

-  `ui purple circular steps`
- `blue completed step`
- `teal step`
- The connection line should inherit the color as we have it in `connected feed`

ko2in:
> `disabled step` does not have any visual effect (you removed that on purpose...why?)

I don't see any difference style for disabled step like box item. Maybe, remove the small circle disc?

lubber-de:
> > `disabled step` does not have any visual effect (you removed that on purpose...why?)
> 
> I don't see any difference style for disabled step like box item. Maybe, remove the small circle disc?

A disabled variant almost always has
- a lower opacity value
- greyed out (if lower opacity doesn't make it visually clear already)
- pointer-events: none (already implemented)

ko2in:
Disabled state would OK with completed or active. The default state already looks similar to the disabled item as it's color and background is grayed out.

ko2in:
If this looks OK for you, then I'll apply also the opacity for the default state.
rail-step-opacity

lubber-de:
the opacity can be applied to either normal state or completed state. But you should still see the default (non-disabled, non completed) state greyed out (if no color is given as suggested above) and opacity:1 (it gets more clear when text is used in the vertical variant

If you meant the same, then: yes 😄 

image

src/definitions/elements/step.less
```diff
@@ -251,49 +381,49 @@
 
 /* Mobile (Default) */
 @media only screen and (max-width: (@largestMobileScreen)) {
-    .ui.steps:not(.unstackable) {
+    .ui.steps:not(.unstackable, .railing) {
```
lubber-de, line 384: If you keep this css level 4 selector syntax, this has to wait until until 2.11 or even 3.0 as only browsers from January 2021 are supporting this syntax.
See https://github.com/fomantic/Fomantic-UI/discussions/2854 for voting results of new CSS level 4 feature inclusion poll

To avoid this use `:not(.unstackable):not(.circular)` instead

ko2in: Thanks for your clarification. Will update PR very soon.

ko2in:
@lubber-de Here we go! I've finished the implementation with the other variantions and all your suggestions.

Due to the nature and design approaches, the circular step doesn't have the following variations: stackable, unstackable, attached and evenly divided.

I'm not sure what is going on with jsfiddle.net. It seems to be stopped working for me since yesterday. I cannot browse the website anymore.

So, I've created the examples and demonstrations in [CodeSandbox](https://codesandbox.io/s/fui-circular-step-hqc7k7?file=/index.html).

jamessampford:
Looking good 😊

The only bit I’m unsure about is the coloured line appearing after the active step, I think it should only go backwards to the previous step (so between completed and active)

Possibly should have an option not to change the header colour as well on the horizontal version (or possibly if by adding `.active` to the header it inherits the same colour?)

Any possibility of having header supported below the horizontal version?

Not sure if possible at all, but could even be pretty cool to have an optional pulsating active version (of the circle, could possibly have the line between active and last completed pulsating as well separately but certainly the active circle by itself)

src/themes/default/elements/step.variables
```diff
@@ -101,6 +117,30 @@
...
+@circularStepsMargin: @stepMargin;
+@circularStepsPadding: @stepMargin;
+@circularStepLineMargin: 2.2em;
+@circularStepDisabledItemOpacity: 0.5;
+@circularStepColor: @borderColor;
+@circularStepBorderWidth: 2px;
+@circularStepBorder: @circularStepBorderWidth solid @circularStepColor;
+@circularStepRingSize: 2.2em;
+@circularStepIconSize: 0.95em;
+@circularStepIconDistance: calc((-@circularStepRingSize + @circularStepIconSize) - 0.35em);
```
lubber-de, line 130: reducing the distance to 0.3 instead of 0.35 made the inner circle look better centered to me (especially when compared to vertical variant)

0.35em:
image

0.3em:
image

vertical (untouched)
image

src/definitions/elements/step.less
```diff
@@ -196,53 +324,107 @@
...
+                    padding-left: @verticalCircularStepContentPadding;
+                }
+
+                &:not(:last-child) {
+                    padding-bottom: @verticalCircularStepPadding;
+
+                    & .content::before {
+                        content: "";
+                        position: absolute;
+                        border-left: @circularStepBorder;
```
lubber-de, line 406: An idea:
What if we make connection line optional as `ui connected circular steps`? This is how we already have it for `ui connected feed`
So only the border-left and each border-color setting would get an additional `connected` selector

What do you think?

lubber-de (changes requested):
👏🏼 This is already very awesome. Your example code is also a good candidate for later docs 😉 
Please see my comments again

src/themes/default/elements/step.variables
```diff
@@ -134,3 +174,10 @@
 @invertedActiveHoverBackground: #444;
 @invertedHoverBackground: #3f3f3f;
 @invertedDisabledBackground: #222;
+
+/* Railing Step */
+@railingStepActiveColor: @completedColor;
+@railingStepCompletedColor: @white;
+@railingStepCompletedBackground: @railingStepActiveColor;
+@railingStepCompletedIconHorizontalPosition: calc(@railingStepDiscPadding / 2);
+@railingStepCompletedIconVerticalPosition: calc((@railingStepDiscPadding * -1.5) + (@railingStepDiscPadding * 1.5));
```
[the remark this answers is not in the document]
ko2in: Look like rebasing brought back these variables.

src/definitions/elements/step.less
```diff
@@ -164,6 +166,113 @@
...
+                &.link ::before,
+                &.link ::after {
+                    cursor: pointer;
+                }
+                &.active,
+                &.completed {
+                    background: @circularStepCompletedBackground;
+                }
+                &.completed::after {
+                    left: calc(-@circularStepLineMargin + (calc((@circularStepLineMargin / 2) / 2)));
```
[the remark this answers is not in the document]
ko2in: You are right. `em` always gives the same result regardless of the size. Let me refactor the `calc` that includes only `em` unit.

`left: calc(-@circularStepLineMargin + (calc((@circularStepLineMargin / 2) / 2)) + 1px);`

We cannot use the fixed value `1px` which can bring different result when the value of `@circularStepLineMargin` has changed. I thought that should be based on the half amount of `@circularStepBorderWidth`.

What do you think?

src/definitions/elements/step.less
ko2in: Do you mean only for vertical step? Then, that's a good idea. 

But for the horizontal step, if the connection line isn't showing, then the UI looks awkward.
horizontal-step-with-no-connection

In this case, we would need title and description like vertical step. And this would be a problem when the screen size is narrower.

ko2in:
> The only bit I’m unsure about is the coloured line appearing after the active step, I think it should only go backwards to the previous step (so between completed and active)

@jamessampford Yeah! That would be better I guess. What do you think @lubber-de ?

lubber-de:
Yes, agree

src/definitions/elements/step.less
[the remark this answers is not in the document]
ko2in: That additional pixel are now evaluated from `@circularStepBorderWidth`. The calc() with em units are kept, since they are evaluating with the variables which could be override by the user, where the default value of `2em / 2` and `4em / 2` in user customized variables would be different outcomes.

ko2in:
Now, the color rule for the rail line is changed as @jamessampford.

The line only after the completed step is now showing the particular color. The rest of the states would use the default color for the rail line.

**Jsfiddle:** https://jsfiddle.net/ko2in/86spora2/
**CodeSandbox:** https://codesandbox.io/s/fui-circular-step-hqc7k7?file=/index.html

jamessampford:
I’ve managed to add optional pulsating to active - https://codesandbox.io/s/fui-circular-step-forked-7tqhs6?file=/index.html - probably needs some tweaking … borrowed from transitions, though needs to target the `::before` - I did try `::after` as well, but think the former looks better on its own instead of the latter or both

ko2in:
> I’ve managed to add optional pulsating to active - https://codesandbox.io/s/fui-circular-step-forked-7tqhs6?file=/index.html - probably needs some tweaking … borrowed from transitions, though needs to target the `::before` - I did try `::after` as well, but think the former looks better on its own instead of the latter or both

@jamessampford You can submit your own PR later for any improvements. I would like to keep it as simple for now.

lubber-de (commented):
That's very awesome by now!
Just a little comment (see below) and a general question (not a request):

There are now a very lot of `:not(.circular)` selector enhancements.
Would it be easier (less code?) to instead override the circular setting (which would be needed for non circular only) instead directly. This is especially noticable for all the evenly divided variants (.x.steps)
I understand you did that because circular steps are styled differently, but perhaps it's worth some minutes to think about that again.

So instead of 
```css
.ui.steps:not(.circular) {
	display: inline-flex;
}
``` 

just override it for the ciscular variant like
```css
.ui.steps {
	display: inline-flex;
}
.ui.circular.steps {
	display: block;
}

```

src/definitions/elements/step.less
```diff
@@ -164,6 +166,111 @@
...
+                    left: -@circularStepRingSize;
+                }
+                &::after {
+                    left: @circularStepIconDistance;
+                }
+
+                &.link ::before,
+                &.link ::after {
+                    cursor: pointer;
+                }
```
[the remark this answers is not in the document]
ko2in: That was a mistake. I'll remove this.

ko2in:
> That's very awesome by now! Just a little comment (see below) and a general question (not a request):
> 
> There are now a very lot of `:not(.circular)` selector enhancements. Would it be easier (less code?) to instead override the circular setting (which would be needed for non circular only) instead directly. This is especially noticable for all the evenly divided variants (.x.steps) I understand you did that because circular steps are styled differently, but perhaps it's worth some minutes to think about that again.
> 
> So instead of
> 
> ```css
> .ui.steps:not(.circular) {
> 	display: inline-flex;
> }
> ```
> 
> just override it for the ciscular variant like
> 
> ```css
> .ui.steps {
> 	display: inline-flex;
> }
> .ui.circular.steps {
> 	display: block;
> }
> ```

I'll try to review again and find out whether it's possible to avoid excluding with `:not(.circular)` and describe the specific style directly.

lubber-de:
I tried to remove the `:not(.circular)` for the evenly divided variants (.two.steps, .three.steps etc.) and it seems it works well with circular variant (because it only affects the width), so at least those can be removed i think, but you may check yourself, too

ivangomes:
Guys this project still good?

lubber-de:
I added support for horizontal content. 
https://jsfiddle.net/lubber/t1j7yxoz/37/
image

lubber-de (approved):

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/538

ko2in:
@lubber-de Thanks your assistance in completing this task for me.

PR #2927, opened 2023-10-14
Jaimin

### LINKED ISSUE #2856: [grid] One wide column not inline when on tiny screen
# Bug Report
With a screen size of 447px and smaller, when a grid contains a `one wide column`, the subsequent `column` is forced onto the next line, even though the number of columns is exactly 16. Displaying on a screen size larger than 448px displays correctly. Changing to `two wide column` (and `fourteen wide column`) also works correctly. 

## Steps to reproduce
1. Create a `ui grid` containing a `one wide column` and a `fifteen wide column`
2. Reduce your window width to 447px

## Expected result
The `fifteen wide column` is displayed **next to** the `one wide column`

## Actual result
The `fifteen wide column` is displayed **under** the `one wide column`. You can see on the jsfiddle that the second `ui grid` with a `two wide column` and a `fourteen wide column` always displays correctly.

## Testcase
https://jsfiddle.net/szxkdre1/1/

## Screenshot (if possible)
One wide column

## Version
2.9.2

JaiminR28:
## Description
Solved the issue with "fifteen wide column" element which is forced to next line on reducing the window size below 447px.
In the semantic.css file by removing the ' flex-wrap' property and setting it to default as no-wrap.

JsFiddle Link of issue: https://jsfiddle.net/xvzns35r/1/

image

## Testcase

Solved JsFiddle Link: https://jsfiddle.net/jaiminrathwa28/5ymr3s1a/9/

## Screenshot (if possible)

chrome_lnBco9GFPq

## Closes

#2856

### DIFF
--- a/dist/components/grid.css
+++ b/dist/components/grid.css
@@ -8,36 +8,35 @@
  *
  */
 
-
 /*******************************
             Standard
 *******************************/
 
 .ui.grid {
-  display: flex;
-  flex-flow: row wrap;
-  align-items: stretch;
-  padding: 0;
+    display: flex;
+    flex-flow: row wrap;
+    align-items: stretch;
+    padding: 0;
 }
 
 /* ----------------------
       Remove Gutters
 ----------------------- */
 .ui.grid {
-  margin: -1rem -1rem;
+    margin: -1rem -1rem;
 }
 .ui.relaxed.grid {
-  margin-left: -1.5rem;
-  margin-right: -1.5rem;
+    margin-left: -1.5rem;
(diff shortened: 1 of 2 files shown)


lubber-de (changes requested):
1. Please don't change the dist/css files directly but do changes inside the /src/*.less files instead 
2. Pleae use a proper PR title 😉 
3. Your change breaks the whole wrapping functionality of the grid and its variants

See
gridflex

lubber-de:
Closing because of the above mentioned reasons and no more response. Please create a new PR if desired

PR #2941, opened 2023-11-21
feat(form): optional ignore native field validations

alvarolm:
adding the attribute "native-validation" with value "false" will cause to skip the native validation message if there is one

## Description
this change allows to disable (skip) the native validation message when adding the attribute "native-validation" with value "false".
non-breaking change.

https://discord.com/channels/453127116427493376/1019661054936096798/threads/1176304818260816007

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1628,6 +1628,7 @@
         errorFocus: true,
         dateHandling: 'date', // 'date', 'input', 'formatter'
         errorLimit: 0,
+        noNativeValidation: false,
 
         onValid: function () {},
         onInvalid: function () {},
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1349,7 +1349,7 @@
                             fieldErrors = [],
                             isDisabled = $field.filter(':not(:disabled)').length === 0,
                             validationMessage = $field[0].validationMessage,
-                            noNativeValidation = $field.filter("[native-validation='false']").length === 1,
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -1349,13 +1349,14 @@
                             fieldErrors = [],
                             isDisabled = $field.filter(':not(:disabled)').length === 0,
                             validationMessage = $field[0].validationMessage,
+                            noNativeValidation = $field.filter("[native-validation='false']").length === 1,
```
lubber-de, line 1352: 1. As the default is to always use native validation, an explicit setting to "false" is IMHO not needed.
Instead, we should just check for an existing "no-native-validation" property
2. I also suggest to support a global setting so a whole form can ignore native validation for each field
3. We should also support `noNativeValidation: true` as a field option when given via field-rules on instantiation

```suggestion
                            noNativeValidation = $field.filter("[no-native-validation]").length > 0 || field.noNativeValidation || settings.noNativeValidation,
```

Please also add `noNativeValidation: false` to the settings object (i suggest after `errorLimit` https://github.com/fomantic/Fomantic-UI/blob/5eb2cfde0c188fb48e504481f0d6be333e8859ff/src/definitions/behaviors/form.js#L1630C23-L1630C23)

lubber-de (changes requested):
Thanks for your contribution. Please see my comments :)

src/definitions/behaviors/form.js
lubber-de: I did some more investigation and i think we need to enhance this even more making use ot the standard novalidate form property and formnovalidate submit property (which we could extend to each input field) i'll update my change request above accordingly

lubber-de (approved):
LGTM now

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/485

PR #2949, opened 2023-11-30
feat(form): deprecated and rename empty to notEmpty rule

### LINKED ISSUE #2937: [form.rule] Rename `empty` rule to be consistent
# Feature Request

originally posted in https://github.com/atk4/ui/issues/2129

Currently, the `empty` form rule is very, very confusing as the logic is negated with the remaining names.

`empty` means "nothing", but the rule validates the form field is "filled / not empty".

here https://github.com/atk4/ui/issues/2129#issuecomment-1816389051 is this fact more evident when compared with other rules like `checked` or `is[dog]`

This is a feature request to rename `empty` form rule to something like `notEmpty` or `mandatory`. For BC, the `empty` can be kept with a console warning.

lubber-de:
## Description

This PR renames the `empty`  rule to `notEmpy` but keeps the `empty` rule working for backward compatibility, so people can just upgrade their code for while. A console warning will tell them about.
Reasons for renaming is that the code actually checks if a field is not empty. More explanation inside the related issue

## Closes
#2937

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1237,20 +1237,20 @@
                                 isRequired = $el.prop('required') || $elGroup.hasClass(className.required) || $elGroup.parent().hasClass(className.required),
                                 isDisabled = $el.is(':disabled') || $elGroup.hasClass(className.disabled) || $elGroup.parent().hasClass(className.disabled),
                                 validation = module.get.validation($el),
-                                hasEmptyRule = validation
+                                hasNotEmptyRule = validation
                                     ? $.grep(validation.rules, function (rule) {
-                                        return rule.type === 'empty';
-                                    }) !== 0
+                                        return ['notEmpty', 'checked', 'empty'].indexOf(rule.type) >= 0;
+                                    }).length > 0
                                     : false,
                                 identifier = module.get.identifier(validation, $el)
                             ;
-                            if (isRequired && !isDisabled && !hasEmptyRule && identifier !== undefined) {
+                            if (isRequired && !isDisabled && !hasNotEmptyRule && identifier !== undefined) {
                                 if (isCheckbox) {
                                     module.verbose("Adding 'checked' rule on field", identifier);
                                     module.add.rule(identifier, 'checked');
                                 } else {
-                                    module.verbose("Adding 'empty' rule on field", identifier);
-                                    module.add.rule(identifier, 'empty');
+                                    module.verbose("Adding 'notEmpty' rule on field", identifier);
+                                    module.add.rule(identifier, 'notEmpty');
                                 }
                             }
                         });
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -1801,10 +1802,16 @@
         rules: {
 
             // is not empty or blank string
-            empty: function (value) {
+            notEmpty: function (value) {
                 return !(value === undefined || value === '' || (Array.isArray(value) && value.length === 0));
             },
 
+            empty: function (value, ancillary, module) {
+                module.debug('Rule "empty" is deprecated and will be removed in a future version. Use the "notEmpty" rule instead.');
```
mvorisek, line 1810: not many developers have debug verbosity enabled, IMHO the lowest visible level by default should be used

prudho (dismissed):
LGTM

bedengler:
Does that mean in future versions, if we are looking for en empty entry, we have to say "not[notEmpty]"?
I'd prefer keeping empty, but doing exactly what it means and if we look for not Empty, we just do a "not[empty]"....

mvorisek:
> Does that mean in future versions, if we are looking for en empty entry, we have to say "not[notEmpty]"? I'd prefer keeping empty, but doing exactly what it means and if we look for not Empty, we just do a "not[empty]"....

I would be fine with `required`, but inverting a meaning is the worst thing possible, as not all devs read all changelogs.

lubber-de:
Infact the `not` rule does not even work this way as it only accepts a string for comparison instead of another rulename. 

`not[empty]` would currently check if the input value String isn't equal to the string "empty"

Of course we could implement supporting this, but In case someone really wants to check for a string "empty" and a rule with the same name exists would then break/ not work again. It might als isn't trivial as we would need to support something like `not[max[6]]` where you could simply do `min[7]` instead.


So existing code simply needs to be updated from `empty` to `notEmpty`.

lubber-de:
> How do you check for an empty entry then? Using "min[1]"?

You can use

`maxLength[0]` or `is[]` to accomplish this

> 
> What about implementing a "notrule" or some characters that signal, we are looking not for a string, but another rule? Something like "not[empty]" looking for the string "empty" and "not[{notEmpty}]" or "notrule[notEmpty]"?

TBH, this gets complicated and all the existing rules have opposite versions already to accomplish such checks.
In doubt you can always create a custom rule

src/definitions/behaviors/form.js
lubber-de: Nevermind: I moved the check to the settings builder so it will be warned into the console on instantiation once.

Please review again

mvorisek: Yes, this looks good!

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/483

PR #2956, opened 2023-12-04
feat(lint): lint ts types using CI

mvorisek:
This PR improves #2822, the Typescript is now tested if compiles and linted againts our coding style.


mvorisek:
@prudho can you make the lint passing, there are especially many "xxx was used before it was defined" and "Do not use a triple slash reference for xxx.ts, use `import` style instead" issues.

mvorisek:
PR is done.

lubber-de (approved):
LGTM, i only changed the "empty" to "notEmpty" rule inside the tests.ts while we are at it ;)

PR #2959, opened 2023-12-07
feat(chore): added asset font license files

### LINKED ISSUE #2954: Appropriately document bundled assets
# Feature Request

Fomantic-UI apparently ships with some third-party assets, which do not seem to be documented appropriately.

Let's say I have https://github.com/fomantic/Fomantic-UI/blob/develop/dist/themes/default/assets/fonts/icons.woff Looking from the outside, it seems like they have been generated by https://github.com/fomantic/create-fomantic-icons using some FontAwesome fonts. In the same assets directory, the Lato fonts (https://fonts.google.com/specimen/Lato/about) seem to be shipped. For both cases, the original font license is OFL-1.1 which imposes some attribution requirements. Unless I missed something, neither the required license terms, attribution or upstream links are provided for the assets.

This should probably be changed to ensure license compliance and to make it clear from the outside where the assets actually originate from.

lubber-de:
## Description

Added missing OFL and Apache 2.0 license files to the respective asset folders according to
https://choosealicense.com/licenses/ofl-1.1/
https://choosealicense.com/licenses/apache-2.0/

## Closes
#2954


src/themes/material/assets/fonts/LICENSE-2.0.txt
```diff
@@ -0,0 +1,202 @@
+
```
mvorisek, line 1: I propose to name it `LICENSE.txt`

lubber-de: Done by 5d137009039de5722d0ae6aa11b06297b00f0eac

PR #2968, opened 2023-12-19
Test TS ExpectType in CI

mvorisek:
Improves #2956 testing.

TBH, I do not know what the issue is and how to debug/fix it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -123,8 +123,13 @@ module.exports = {
     overrides: [{
         files: ['**/*.ts'],
         parser: '@typescript-eslint/parser',
+        parserOptions: {
+            project: ['types/tsconfig.json'],
+        },
+        plugins: ['eslint-plugin-expect-type'],
         extends: [
             'plugin:@typescript-eslint/recommended',
+            'plugin:expect-type/recommended',
         ],
         rules: {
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
No clue, I've never seen this repository before 😄.

mvorisek:
Hi @prudho, in #2822, more precisely in https://github.com/prudho/Fomantic-UI/blob/1acdce3d8a4e3c5684a46192a1d8dd823d0c8aa4/types/fomantic-ui-tests.ts you have added a test with expected types. Might I know if you ever run this test, and if yes, what package did you use and how to test it using CI?

prudho:
Hi @mvorisek ! This file was required as a condition for the type definition to be published to the DefinitelyTyped repo. They seem to need it to run only a simple type check when validating the definitions, using `dtslint` (see https://github.com/Microsoft/DefinitelyTyped-tools/tree/main/packages/dtslint#write-tests). TBH this file was more a stub than a real test file when I "wrote" it (copy from another definition type lib), for a real CI I think that tests should be way more expanded.

mvorisek:
I do not know how to make the TS types CI testing working. Help welcomed - then please tag me on your PR.

Issue #2920, opened 2023-10-08
[Modal] "show" does not recenter modal


mvorisek:
# Bug Report

## Steps to reproduce

1. open https://dev.atk4.org/demos/_unit-test/late-output-error.php
2. click first button (`Test LateOutputError I: Headers already sent`)
3. notice the modal is NOT vertically centered

## Expected result

`show` must always refresh/recenter the modal

## Screenshot (if possible)

![image](https://github.com/fomantic/Fomantic-UI/assets/2228672/eeb1dc1e-4349-4776-9b9d-6fde56869d34)

## Version

2.9.2

mvorisek:
The repro demo does no longer reproduce the issue, as we fixed it on our side temporary: https://github.com/atk4/ui/blob/7cc2e65c144480962433dd4aef5ef351380dace2/js/src/services/modal.service.js#L48

To reproduce it, download the demo (html, js, + the loaded page using AJAX) locally and remove the fix.

It seems the "refresh on show" was removed in https://github.com/fomantic/Fomantic-UI/pull/2477/commits (in last commit) - what about reverting it (reintroducing forced refresh on show) until https://github.com/fomantic/Fomantic-UI/issues/2476 is fixed?

lubber-de:
> UPDATE: I looked into the latest JS code and please have a look at [#2476 (comment)](https://github.com/fomantic/Fomantic-UI/issues/2476#issuecomment-1763039845) repro. The repro is changing classes not only style, and such changes should be observeable and causing always a refresh when `observeChanges: true`.

Always triggering a refresh on any classname change within the whole modal will almost always trigger a refresh especially for modules like dropdown or other elements which make use of the transition module / classes. I fear a big performance lag here then. Especially as the "refresh()" method itself also changes classnames inside the modal (e.g. the "scrolling" or "loading"  class)

lubber-de:
Should be fixed by #2969. The new Resizeobserver automatically triggers a refresh when a modal gets shown so a separate refresh on show isnt needed

PR #2969, opened 2023-12-20
fix(modal): use resizeobserver

### LINKED ISSUE #679: Modal shifts around when scrollbar appears
# Bug Report
A modal with a tabular menu inside (or anything else that causes the modal height to change) will shift when the height exceeds the window height, and a scrollbar appears on the browser window. 

## Expected result
The modal position should not change when a scrollbar is shown/hidden

## Actual result
The modal position shifts slightly.

## Screenshot (when possible)
5476b52882da4a57b3b8dbf6746cae83

lubber-de:
## Description

This PR changes the refresh behavior of modals to recalculate their dimensions. The previous logic was checking for DOM-Nodes only, but the only need was to know when the modal gets resized.

This now works in Firefox as well.

The dimmers scrollbar will now also not trigger content movement (#679) anymore when used in [modern browsers](https://caniuse.com/?search=scrollbar-gutter). However this does not nicely work when `detachable: false` is used (tried a _lot_ dealing with complicated tricks using the html tag as the scrollbar-gutter property does not work for the body element out of the box but it always left a white scrollbar even if the modal fitted into the viewport, so i decided to live with that limitation for now)
As this feature will always occupy a possible scrollbar width (even if a scrollbar won't appear), only modals which are supposed to be scrollable (by interaction or application logic while the modal is shown) should be given a new css class `scrollable` to gain the fix. Otherwise every modal wouldnt be centered anymore.

I also adjusted the dimmer positioning and close icon appearance when `detachable:false` in a custom context is used

## Testcase
Resize the textarea, a scrollbar should appear to the modals dimmer
https://jsfiddle.net/lubber/drb6fpog/4/

## Closes
#679 
#2476
#2920

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved modal resizing/position updates by detecting element size changes and syncing with DOM mutations.
  * Refined modal refresh, fit, and centering logic across body and non-body contexts (including scroll offsets).
  * Corrected dimmer scrollbar handling, height calculations, and cleanup when showing/hiding modals.
* **Styles**
  * Updated dimmer/modal desktop behavior, added conditional `scrollbar-gutter` support, and improved scrolling modal overflow/spacer handling for better browser compatibility.


src/definitions/modules/modal.js
```diff
@@ -263,6 +267,13 @@
 
                 observeChanges: function () {
                     if ('MutationObserver' in window) {
+                        if ('ResizeObserver' in window && settings.observeChanges) {
```
mvorisek, line 270: Does modal use the standard `MutationObserver` for anything else than size? If no, `ResizeObserver` should be enough when supported by the browser.

lubber-de: The mutationobserver is still needed for the autofocus and tabbing functionality

src/definitions/modules/modal.less
```diff
@@ -447,6 +448,9 @@
         overflow: auto;
         overscroll-behavior: @overscrollBehavior;
     }
+    .modals.dimmer {
+        scrollbar-gutter: stable;
```
mvorisek, line 452: Why is this needed, will it imply scrollbars always shown? In the PR description there is mentioned something with `detachable: false`, is this related?

There is mentioned "The dimmers scrollbar will now also not trigger content movement anymore". I tried the demo (in Firefox) and when the textarea is resized, the content is unexpectedly always scrolled to the top when resized.

lubber-de: > Why is this needed, will it imply scrollbars always shown? In the PR description there is mentioned something with `detachable: false`, is this related?

No, the scollbars are not always shown. This will occupy a possible shown scrollbar margin, but will _not_ show an empty scrollbar in case the element is _not_ body
[See example at #679](https://github.com/fomantic/Fomantic-UI/issues/679#issuecomment-1865193573)

The problem with "detachable:false" in case context is body, is that body does not support this feature (firefox+safari dont care at all, becaus those have their own implemented "overlay scrollbar feature")
Because of this the parent element would need to get scrollbar-gutter:stable instead. This basically works, but as the html element does not have any dimmer color set, it would display a white empty scrollbar in case the modal fits the screen. It also left a second scrollbar when closing the modal...very ugly

If you want to see/try my hack:

The jsfiddle does only barely  show the issue as jsfiddle has a black parent background which the html element inside the iframe inherits (so the permanent occupied right scrollbar area is black, but still not the "dimmer black"
https://jsfiddle.net/lubber/cyjnta1d/3/

I also tried messing around with the dimmer animation to also make the html element inherits that, but the transition was not in sync and the code bloated away as we would need to support every possible dimmer variation just for that.

We would also need to support non body context elements... lots of code and cases for, IMHO, little benefit 
```css
@supports (scrollbar-gutter: stable) and (selector(:has(.f))) {
    html:has(body.undetached.dimmable) {
      overflow: auto;
    }
    html:has(body.undetached.dimmable.dimmed) {
      overflow: auto;
      scrollbar-gutter: stable;
    }
    body.undetached.dimmable.dimmed.dimmed {
      overflow: hidden;
    }
}
```

> 
> There is mentioned "The dimmers scrollbar will now also not trigger content movement anymore".
What is meant by this is https://github.com/fomantic/Fomantic-UI/issues/679#issuecomment-1865193573

> I tried the demo (in Firefox) and when the textarea is resized, the content is unexpectedly always scrolled to the top when resized.

mmh, thats the above in between discussion where ko2in and me were not able to reproduce this. 

The modal is only put to the top when the it detects the current modals size would not fit (thus adds the `scrolling` class). Whenever i am able to reproduce this, i take a look again and prepare a separate PR as some browser (firefox) measurements seem to provide questionable values inside `module.can.fit()` in some situations

mvorisek: Thank you for your detailed explanation and your work on this topic.

> > I tried the demo (in Firefox) and when the textarea is resized, the content is unexpectedly always scrolled to the top when resized.
> 
> mmh, thats the [in between discussion](https://github.com/fomantic/Fomantic-UI/issues/2476#issuecomment-1763365472) where i was not able to reproduce this.

I meant the demo of/from this PR - https://jsfiddle.net/lubber/drb6fpog/1/

Steps to reproduce:
1. open the https://jsfiddle.net/lubber/drb6fpog/1/ in Firefox
2. resize the textarea vertically by dragging the right bottom corner to make it larger than the modal
3. scroll down the modal
4. now try to resize the dropdown by dragging the right bottom corner
5. the modal is unexpectedly scrolled to the top immediately - I would expect no scroll/jump
6. also notice when you click the textarea resize corner and have F12 "Inspector" tab open, `div.dimmer div.modal` div in the inspector tab flashes yellow meaning something has changed, but I did not noticed/see any real css/class changes, meaning some same value is set from a mutation observer or even possibly worse, some value is set and immediatelly changed back

lubber-de: i take a look into this

ko2in (dismissed):
LGTM

src/definitions/modules/modal.js
```diff
@@ -865,6 +898,18 @@
                         });
                         module.cache.topOffset = -(module.cache.height / 2);
                     }
+                    if (getScrollingMargins) {
+                        if (module.can.useFlex()) {
+                            $module.addClass(className.scrolling);
+                            scrollingTop = parseInt($module.css('top').replace(/[^\d.]/g, ''), 10) || 0;
```
mvorisek: Is the `top` css set by us - if yes, we should know the format and something like `.replace(/px$/, '')` would be much more cleaner.

mvorisek:
I have tested the JS and the CSS from the jsfiddle and it does indeed fix the "resize textarea issue in Firefox".

@lubber-de might you update jsfiddle (https://jsfiddle.net/lubber/drb6fpog/1/ ATM) to represent the latest compiled CSS?

I am unable to review this PR at the code level, but I would be happy to test all of my reported issues.

ko2in (approved):
LGTM.

PR #2973, opened 2024-01-10
fix(firefox): has selector freezes in certain situations

### LINKED ISSUE #2971: [Dropdown] Search with remote content causes Firefox to hang
# Bug Report
I have noticed strange behaviour in the Firefox browser version 121 and higher. The browser hangs when searching remote from API. Strangely enough, not all sources are causing the problem.

## Steps to reproduce
It can be reproduced on the current docs
1. Open Firefox
2. Go to: https://fomantic-ui.com/modules/dropdown.html#sanitized-version
3. Start typing in "Sanitized Version" dropdown

## Expected result
I expect the list to be filtered.

## Actual result
The page is actually frozen under Firefox and surprisingly this filtering doesn't work in other browsers.
On the other hand the other examples with remote content works fine, e.g.: https://fomantic-ui.com/modules/dropdown.html#match-search-query-on-server

## Testcase
https://fomantic-ui.com/modules/dropdown.html#sanitized-version  

## Screenshot (if possible)
Screenshot 2024-01-06 at 11 24 28
Screenshot 2024-01-06 at 11 28 09

## Version
Fomantic ver 2.9.3  
Firefox 121

lubber-de:
## Description

Firefox's implementation of the `:has` selector (perhaps in combination with the `@supports` selector) makes firefox freeze in certain situations.
After some investigation the reason are 3 of those selectors we implemented in 2.9.3. 
As Firefox had officially enabled the `:has` feature by default starting with FF 121 it was not happening before (and perhaps no FUI user was actively enabling it in older FF versions)

As all other browsers work just fine this PR removed Firefox support for now until it's fixed by mozilla
I detected only the changed three `@supports selector(:has)` selectors (in combination!) to trigger the freeze. In my tests, every other existing `@supports selector(:has)` selectors invented in 2.9.3 were not affected  so i kept them untouched in this PR

## Testcase
- Use Firefox
- starting to type "Computer" in the input field

Keep in mind that the selectors in question do not even belong to the example code! There is no piled segment and no pointing dropdown so the selector will not even trigger. 

### Broken
The tab freezes.
https://jsfiddle.net/e70xjovw/8/

### Fixed
https://jsfiddle.net/e70xjovw/9/

## Closes
#2971

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -250,7 +250,7 @@
             margin-bottom: 0;
         }
     }
-    @supports selector(:has(.f)) and (not(-moz-appearance: none)) {
+    @supports selector(:has(.f)) and (not (-moz-appearance: none)) {
         *:has(> .ui.piled.segment) {
             z-index: 0;
             position: relative;
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -250,7 +250,7 @@
             margin-bottom: 0;
         }
     }
-    @supports selector(:has(.f)) {
+    @supports selector(:has(.f)) and (not(-moz-appearance: none)) {
         *:has(> .ui.piled.segment) {
             z-index: 0;
             position: relative;
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -1635,7 +1635,7 @@ select.ui.dropdown {
         z-index: @pointingArrowZIndex;
     }
 
-    @supports selector(:has(.f)) {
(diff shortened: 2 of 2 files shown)


lubber-de:
We might wait for a (short) while as mozilla is aware of that by https://bugzilla.mozilla.org/show_bug.cgi?id=1874066
Just in case it gets fixed in firefox soon, we probably dont need this PR then

lubber-de:
I can confirm, the issue is fixed in latest firefox 123.a01
https://archive.mozilla.org/pub/firefox/nightly/2024/01/2024-01-20-09-39-31-mozilla-central/

lubber-de:
Firefox 123 fixed the related issue, this PR is not needed anymore

PR #2974, opened 2024-01-11
feat(dropdown): direction update

alvarolm:
## Description
due to changes around the drodown, it might be needed to update the direction
in my case this is needed to keep critical ui components visible

non-breaking change

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2463,7 +2463,7 @@
 
                 update: {
                     direction: function (direction) {
-                        settings.direction = direction
+                        settings.direction = direction;
                     },
                 },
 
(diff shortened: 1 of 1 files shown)


lubber-de:
Each module already has a "setting" behavior to change settings after instantiation to avoid doing methods for each setting separately.

To accomplish your use case you could use

```javascript
  $('.ui.dropdown').dropdown('setting', 'direction', 'upward');
```

I have to admit this is not clearly stated everywhere in the docs. In some examples it is used, so one could stumple upon this to find out.

alvarolm:
I just looked up the dropdown documentation selectively.
thanks for clarifying, maybe a documentation update is needed.

PR #2982, opened 2024-01-19
Do not flash loading class in modal refresh

mvorisek:
introduced in https://github.com/Semantic-Org/Semantic-UI/pull/5578

Is the hack using `loading` class still needed? It causes a problem in out environment because it flashes the class causing our observers to trigger.

Maybe more changes/LoC removed are needed/wanted.


prudho:
Remove the loading hack looks ok to me, but deleting the CSS is not a good idea IMHO. Some users might want to make their modals in a `loading` state, when using ajax for example.

mvorisek:
I still do not understand the https://github.com/Semantic-Org/Semantic-UI/commit/7450aa3c9c32d0d608e73c2f52b7bfa4f32a718f described in https://github.com/Semantic-Org/Semantic-UI/pull/5578#issuecomment-320562100. Maybe flashing the `loading` class is needed.

I am closing this PR as I am unsure about this PR, we needed also the CSS removed (https://github.com/atk4/ui/blob/37a5a0b522/js/src/JqueryPlugin/ReloadViewPlugin.js#L53) and we worked around it on atk4 side.

lubber-de:
Jack was basically right: To be able to calculate/ access dimensions of DOM nodes, they have to be visible (to the browser, off canvas is fine, but display:none is not, as this does not occupy any visible space)
However, i still want #2969 to be implemented and perhaps the whole JS calculation isnt necessary anymore for that purpose, especially after the IE drop. I'll tag this for investigation

PR #2985, opened 2024-01-29
feat(lint): enforce NL before opening curly braces

mvorisek:
the new config is eslint default, only AirBnB CS changed it in https://github.com/airbnb/javascript/blob/eslint-config-airbnb-v19.0.4/packages/eslint-config-airbnb-base/rules/style.js#L21

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
         '/test/helpers',
     ],
     rules: {
+        'brace-style': ['error', '1tbs'],
         'class-methods-use-this': 'off',
         'comma-dangle': ['error', {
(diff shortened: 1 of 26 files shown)


lubber-de (commented):
According to https://eslint.org/docs/latest/rules/brace-style , "native" brace-style is deprecated since eslint 8.9.53 and would need https://eslint.style/packages/js 
So i suggest to use the style package right away . What do you think?
 [](https://eslint.style/rules/js/brace-style)

mvorisek:
I would like to serve and keep everything pristine, but I am currently out of time to do such large migration. This PR is small CS update only.

lubber-de (approved):
LGTM

PR #2990, opened 2024-02-07
Refactor: Apply border-radius styles to last row of table footer

Festiis:
## Description
Changed the CSS selectors to target the last row instead of the first row in the table footer.
Previously, the styles targeted the first row of the table footer. 

## Testcase
https://jsfiddle.net/Festiis/02m8dx13/7/


lubber-de (approved):
LGTM

PR #3003, opened 2024-02-15
fix(checkbox): disallow keynav on readonly radios

### LINKED ISSUE #2812: [checkbox] Read-only checkbox state must not be changeable from UI

lubber-de:
## Description
A readyonly radio checkbox should not change its state via keynav from another radio of the same group.
This PR covers the situation when the readonly radio is already unchecked. 

When it is already checked, it doesn't make sense to set this as the_only_ option to readonly as a radio is part of a radio group which, by nature, only has 1 active selection. And if one does not want this already selected option to be changed, all the other options from the same radio group should be either set to readonly or disabled.

## Testcase
- Click the Z1 option
- Press down arrow, nothing happens because the next option is readonly
https://jsfiddle.net/lubber/xud45evh/12/

## Closes
#2812

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -210,7 +210,7 @@
                         ;
 
                         var
-                            r = module.get.radios(),
+                            r = module.get.radios().not(selector.disabled),
                             rIndex = r.index($module),
                             rLen = r.length,
                             checkIndex = false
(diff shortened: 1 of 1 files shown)


mvorisek:
In my opinion disabled/readonly radios should be skipped instead of keynav blocked.

Imagine:
```
r1 selected
r2 disabled
r3
r4 disabled
```

when r1 is selected, it should be possible to navigate to r3, with this PR, it is not possible. I tested the fiddle in the PR description.

lubber-de:
Disabled radios will be skipped automatically by the browser engine, but skipping read-only gets too complicated as we would have to check each  possible next to next radio

mvorisek:
So the "selection move" it done by browser (not Fomantic-UI JS) and we only reject it if disabled/readonly, right? If so, this PR LGTM and thank you!

PR #3004, opened 2024-02-18
feat(dropdown): support custom data attributes for items

### LINKED ISSUE #1236: [dropdown] How can i set custom data in options when i dynamic generate options?
# Help Wanted
When options are dynamically generated, how to set custom data(like `<option value='...' data-id='...'>...</option>`) in it?

## Problem
I want the dropdown to dynamic generate options by the user input. For example:
when user input a, the dropdown should display a1,a2,a3... , when user input b, the dropdown should display b1, b2, b3 ... . so i set this,
```js
{
    sortSelect: true,
    fullTextSearch: 'exact',
    on: 'hover',
    allowAdditions: true,
    clearable: true,
    forceSelection: false
}
//
$('input').on('input', function() {
    let v = $(this).val();
    $(this)
        .dropdown('change values', [
            { name: v + '1', value: v + '1'},
            { name: v + '2', value: v + '2'},
            { name: v + '3', value: v + '3'}
        ]);
});
```

it's fine. but now i need to add my custom data to the options When they are dynamically generated, because other field's value may depend on this choosen option's id. something like this
```js
$('input').on('input', function() {
    let v = $(this).val();
    $(this)
        .dropdown('change values', [
            { name: v + '1', value: v + '1', data-id: '...'},
            { name: v + '2', value: v + '2', data-id: '...'},
            { name: v + '3', value: v + '3', data-id: '...'}
        ]);
});
```
it's not work. so, how can i set my own data to the option element?

lubber-de:
## Description
This PR allows to render custom data- attributes for each dropdown item by providing a `data` key object.
Every key inside the data object will be rendered as data attribute (only "text" and "value" will be ignored as those are already used by the dropdown module internally)

## Testcase
```javascript
$('.ui.dropdown').dropdown({
    values: [
      {
        name: 'Mother',
        value: 'mom',
        data: {
          husband: 'george',
          age: 73
        }
      }
    ]
  })
;
```

which results in

```html
<div class="item" data-value="mom" data-husband="george" data-age="73">Mother</div>
```

## Screenshot
image

## Closes
#1236

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -793,7 +793,7 @@
                                 }
                                 if (module.is.multiple()) {
                                     $.each(preSelected, function (index, value) {
-                                        $item.filter('[data-value="' + value + '"]')
+                                        $item.filter('[data-' + metadata.value + '="' + value + '"]')
                                             .addClass(className.filtered)
                                         ;
                                     });
@@ -4163,6 +4163,7 @@
             descriptionVertical: 'descriptionVertical', // whether description should be vertical
             value: 'value', // actual dropdown value
             text: 'text', // displayed text when selected
+            data: 'data', // custom data attributes
             type: 'type', // type of dropdown element
             image: 'image', // optional image path
             imageClass: 'imageClass', // optional individual class for image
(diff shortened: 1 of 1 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/500 🥳

PR #3009, opened 2024-03-02
fix(dropdown): clearable icon was missing or misaligned

lubber-de:
## Description
clearable icon ...
- was missing when used inside a search selection dropdown
- was misaligned when used inside an inline dropdown

The basic fixes where already implemented but only if the `clearable` class was set manually, which this PR now automatically adds when the clearable setting is used.

## Testcase
https://jsfiddle.net/lubber/nwjqr7Le/

### Broken
image

### Fixed
image


prudho (approved):
LGTM

PR #3011, opened 2024-03-02
feat(a11y): support alt attribute for images

lubber-de:
## Description
This PR adds a bit of more a11y support to toast, dropdown and search when it comes to generate images and adding an optional alt-attribute

- new option `alt` for toast (will be used when showImage is used as well)
- support for `alt` key in search response object (will be used if an image key is given as well)
- support for `alt` as dropdown-item key (will be used if an image key is given as well)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/494

PR #3023, opened 2024-03-20
fix(feed): inverted metadata actions

Festiis:
## Description
Added support for inverted feed metadata actions by adding styles in feed.less and defining corresponding color variables in feed.variables

## Testcase
https://jsfiddle.net/Festiis/sn82grqL/1/

### DIFF
--- a/src/themes/default/views/feed.variables
+++ b/src/themes/default/views/feed.variables
@@ -190,6 +190,9 @@
 @invertedLikeHoverColor: @invertedSelectedTextColor;
 @invertedLikeActiveColor: @invertedLikeColor;
 
+@invertedMetadataActionColor: @invertedLightTextColor;
+@invertedMetadataActionHoverColor: @selectedTextColor;
+
 @invertedConnectedBorderColor: @whiteBorderColor;
 @invertedDividedBorderColor: @whiteBorderColor;
 @invertedTextLabelColor: @white;
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -117,16 +117,16 @@
 .ui.table > tfoot > tr > td:first-child {
     border-left: none;
 }
-.ui.table > tfoot > tr:first-child > th:first-child,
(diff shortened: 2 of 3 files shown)


src/themes/default/views/feed.variables
```diff
@@ -190,6 +190,9 @@
 @invertedLikeHoverColor: @invertedSelectedTextColor;
 @invertedLikeActiveColor: @invertedLikeColor;
 
+@invertedMetadataActionColor: @invertedLightTextColor;
```
lubber-de, line 193: This color is too dark as it is declared in site.variables as `@invertedLightTextColor: rgba(255, 255, 255, 0.7);`

lubber-de (changes requested):
The hover color is dark
See https://jsfiddle.net/Festiis/sn82grqL/1/

src/themes/default/views/feed.variables
```diff
@@ -190,6 +190,9 @@
...
+@invertedMetadataActionHoverColor: @selectedTextColor;
```
lubber-de, line 194: This color is too dark as it is declared in site.variables as `@selectedTextColor: rgba(0, 0, 0, 0.95);`
See https://jsfiddle.net/Festiis/sn82grqL/1/

lubber-de (changes requested):

src/themes/default/views/feed.variables
lubber-de: It should be `@invertedSelectedTextColor` instead which is `@invertedSelectedTextColor: rgba(255, 255, 255, 1);`
See https://jsfiddle.net/lubber/sdq5zbhv/

```suggestion
@invertedMetadataActionHoverColor: @invertedSelectedTextColor;
```

Festiis:
> The hover color is dark See https://jsfiddle.net/lubber/sdq5zbhv/1/

Of course it should

lubber-de (approved):
LGTM

PR #3042, opened 2024-05-04
fix(deps): downgrade ordered-read-streams

lubber-de:
## Description

Turns out the latest version of ordered-read-streams, 2.0.0, does not always order the files as they were given.
Looks like the issue is inside the new used streamx lib. It seems the given order is modified internally by the end of the file read, so larger files are still merged at the end rather as the desired position.
For example the large emoji.css was always set to the end of the order although the order of that file was given much earlier in the stream array.
Reverting back to to v 1.0.1 fixes that (as it does not use streamx, but readable-stream instead of streamx)

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "merge-stream": "^2.0.0",
     "mkdirp": "^1.0.4",
     "normalize-path": "^3.0.0",
-    "ordered-read-streams": "^2.0.0",
+    "ordered-read-streams": "^1.0.1",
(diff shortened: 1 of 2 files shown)


mvorisek:
Would be great to have an upstream linked.

lubber-de:
- Change the package.json to use 2.0.0 instead of 1.0.1
- yarn install
- yarn build
- compare resulting semantic.css with the previous one. It is completely different

PR #3060, opened 2024-06-19
fix(form): dont add whitespace on empty required field labels

lubber-de:
## Description
Followup of #2711

Whenever an empty label was used for a required single field or non-grouped, non-inline fields group, the label was still receiving an extra space to create a default height and did not display the asterisk. The intention of the space character was for empty non required fields to match same height when used inside a fields row.

If one does not want a label at all and does not want to make this field required, you would not render the label at all and wont get additional space then

## Closes
https://github.com/fomantic/Fomantic-UI/pull/2711#issuecomment-2176221081

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -849,8 +849,8 @@
     }
 }
 
-.ui.ui.ui.ui.form .fields > label:empty::after,
-.ui.ui.ui.ui.form .field > label:empty::after {
+.ui.ui.ui.ui.form .fields:not(.required) > label:empty::after,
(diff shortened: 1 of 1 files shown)


mvorisek:
In my original https://jsfiddle.net/h9ta2wdj/ the space is added for both required/non-required fields with empty label, can you edit the CSS in my repro to show what this change would do? I think the change is not fully enough.

lubber-de:
@mvorisek Thanks for the feedback
I think, i fixed it now
See your adjusted fsfiddle here https://jsfiddle.net/lubber/owayxz3k/22/

Issue #2923, opened 2023-10-09
[form] validation of grouped field shows multiple error messages


hugopeek:
# Bug Report

I noticed that after the 2.9.3 update, grouped fields show a repeating validation error for each field inside the group. This used to be 1 message before.

I tried using the new `errorLimit: 1` setting, but that doesn't seem to work in this case.

## Testcase
https://jsfiddle.net/md9sexog/1/

## Screenshot (if possible)
![image](https://github.com/fomantic/Fomantic-UI/assets/5436121/4f50c6cd-30ba-4a80-9bec-494e2fe67b2c)

## Version

2.9.3

BrianJavier99:
Hello, I hope you don't mind, but I have tried my hand at fixing this issue. Please let me know if it is an acceptable approach.

https://github.com/fomantic/Fomantic-UI/pull/3067

lubber-de:
> Hello, I hope you don't mind, but I have tried my hand at fixing this issue. Please let me know if it is an acceptable approach.
> 
> #3067

Thanks, please see my PR comments

Issue #2797, opened 2023-05-22
[form] `'on': 'blur'` does not clear dropdown field error


mvorisek:
# Bug Report

## Steps to reproduce
1. open https://jsfiddle.net/4jvtch20/
2. click "Save" - an error is added to each field (expected/correct)
3. click "Email" field input, once you focus out of the field, the error for this field is removed
4. click "Dropdown" field input, no click in/out or value change does remove the field error

## Expected result

With `'on': 'blur'` I would expect consistent behaviour, ie. error similarly removed for dropdown as for standard input.

## Version

2.9.2 / latest

mvorisek:
@lubber-de shouldn't the dropdown emit `blur` event as other fields normally does?

lubber-de:
> @lubber-de shouldn't the dropdown emit `blur` event as other fields normally does?

Given your example, the dropdown does not have a validation rule set. And it also does not have a native (non-hidden, non-search) input field to be validated automatically. Thats, why the validator does completely ignore blurring the dropdown.

I'll try to fetch this situation

mvorisek:
In https://github.com/atk4/ui/pull/2052 the issue is described a little better.

We set `on: 'blur'` https://github.com/atk4/ui/blob/4.0.0/src/Form.php#L554 which validates the whole form on blur. For inputs the validation works, but for dropdown it does not.

Currently, custom JS to remove validation errors manually is needed - https://github.com/atk4/ui/pull/2052/files . It works, but I would be happy if this could be addressed in Fomantic UI directly as both components (input, dropdown) are standard Fomantic UI ones.

lubber-de:
Fixed by #3129
See your adjusted jsfiddle here https://jsfiddle.net/lubber/tmhrgz41/1/

mvorisek:
The jsfiddle does not work on my side.

Steps to reproduce:
1. click save
2. errors are set for both fields - expected
3. click input field
4. click somewhere else - error on input field is gone - expected
5. click dropdown field
6. click somewhere else - error on dropdown field is kept - I would expect it gone as well, hence this issue

PR #3067, opened 2024-07-03
fix(form): duplicate error displays on invalid form field value

BrianJavier99:
## Description

Adds a variable `formErrorsTracker` that tracks which field identifiers have already pushed an error prompt in the `formErrors` variable which limits the number of error prompt for every field identifier into 1.

## Test

https://jsfiddle.net/94oer63c/

## Screenshot (if possible)
image

## Closes

#2923

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1398,7 +1398,7 @@
                                                 // Also allow multiple error prompts per field identifier but make sure each prompt is unique
                                                 || !(formErrorsTracker[identifier].includes(fieldError))
                                                 // Limit the number of unique error prompts for every field identifier if specified
-                                                && formErrorsTracker[identifier].length < errorLimit
+                                                && (!errorLimit || formErrorsTracker[identifier].length < errorLimit)
                                             ) {
                                                     fieldErrors.push(fieldError);
                                                     (formErrorsTracker[identifier] = formErrorsTracker[identifier] || []).push(fieldError);
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1394,14 +1394,14 @@
                                         if (!settings.inline) {
                                             if (
                                                 // Always allow the first error prompt for new field identifiers
-                                                !(identifier in formErrorsTracker) 
+                                                (!(identifier in formErrorsTracker)
                                                 // Also allow multiple error prompts per field identifier but make sure each prompt is unique
-                                                || !(formErrorsTracker[identifier].includes(fieldError))
+                                                || !formErrorsTracker[identifier].includes(fieldError))
                                                 // Limit the number of unique error prompts for every field identifier if specified
                                                 && (!errorLimit || formErrorsTracker[identifier].length < errorLimit)
                                             ) {
-                                                    fieldErrors.push(fieldError);
-                                                    (formErrorsTracker[identifier] = formErrorsTracker[identifier] || []).push(fieldError);
+                                                fieldErrors.push(fieldError);
+                                                (formErrorsTracker[identifier] = formErrorsTracker[identifier] || []).push(fieldError);
                                             }
                                         } else {
                                             fieldErrors.push(fieldError);
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -949,7 +949,7 @@
                                     $prompt.css('display', 'none');
                                 }
                                 $prompt
-                                    .appendTo($fieldGroup.filter('.error'))
+                                    .appendTo($fieldGroup.filter('.' + className.error))
                                 ;
                             }
                             $prompt
@@ -1396,7 +1396,7 @@
                                                 // Always allow the first error prompt for new field identifiers
                                                 (!(identifier in formErrorsTracker)
                                                 // Also allow multiple error prompts per field identifier but make sure each prompt is unique
-                                                || !formErrorsTracker[identifier].includes(fieldError))
+                                                || !formErrorsTracker[identifier].indexOf(fieldError))
                                                 // Limit the number of unique error prompts for every field identifier if specified
                                                 && (!errorLimit || formErrorsTracker[identifier].length < errorLimit)
                                             ) {
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1398,7 +1398,7 @@
                                                 // Also allow multiple error prompts per field identifier but make sure each prompt is unique
                                                 || formErrorsTracker[identifier].indexOf(fieldError) === -1)
                                                 // Limit the number of unique error prompts for every field identifier if specified
-                                                && (!errorLimit || formErrorsTracker[identifier].length < errorLimit)
+                                                && (!errorLimit || (formErrorsTracker[identifier] || []).length < errorLimit)
                                             ) {
                                                 fieldErrors.push(fieldError);
                                                 (formErrorsTracker[identifier] = formErrorsTracker[identifier] || []).push(fieldError);
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -949,7 +949,7 @@
                                     $prompt.css('display', 'none');
                                 }
                                 $prompt
-                                    .appendTo($fieldGroup)
+                                    .appendTo($fieldGroup.filter('.error'))
                                 ;
                             }
                             $prompt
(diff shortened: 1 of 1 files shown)


src/definitions/behaviors/form.js
```diff
@@ -1403,7 +1412,7 @@
                                 settings.onValid.call($field);
                             }
                         } else {
-                            if (showErrors) {
+                            if (showErrors && newIdentifier) {
```
lubber-de, line 1415: See above. The extra check isn't necessary as this "else" branch will only be called when `fieldValid` is false which is already set when at least one element was detected as invalid

src/definitions/behaviors/form.js
```diff
@@ -1356,6 +1358,7 @@
                                 : false,
                             fieldValid  = true,
                             fieldErrors = [],
+                            newIdentifier = false,
```
lubber-de, line 1361: not needed, see below

lubber-de (changes requested):
Thanks, it basically works, but the extra `newIdentifier` check isn't needed. Please see my comments

BrianJavier99:
> Thanks, it basically works, but the extra `newIdentifier` check isn't needed. Please see my comments

Hello @lubber-de, the reason why I added a newIdentifier variable is to avoid an error popping up in the console:
image

When I investigated the root cause, I discovered the cause being the call to `module.add.prompt()` have the `fieldErrors` variable as an argument, which was empty since we now check if the error prompt is a duplicate before pushing it to the `fieldErrors` array.

This means that while nothing will be concatenated to the `formErrors` array, the subsequent lines, `module.add.prompt` and `settings.onInvalid.call`, will still also be called with null or empty arguments and will prompt errors.

However while typing this out, I have thought out of a better implementation, which is to check if the `fieldErrors` array has no content instead.

lubber-de (changes requested):
@BrianJavier99 Thanks for the explanation. Infact i forgot this function is called for each of the radio button fields (and each rule applied to it) as all share the same "name" attribute.
However, this now results into new bugs: 
- The `errorLimit` setting is not working anymore as it now always allows 1 error of a unique field identifier regardless of given rules
- `inline: true` is broken now  and renders empty pointing labels to the remaining fields sharing the same identifier	
  - while just removing the empty ones might be ok for radiogroups, it isnt for other fields like textfields send as array to the server using "name[]" syntax 
  - Also only the last element got the label (with errormessage) now
  - When the form got multiple fields the second time the form gets validated, all same identifier prompts will be shown again

Here is my Testcase (without your changes) https://jsfiddle.net/lubber/2v0e3s4m/4/
for you to apply your code and test.

Without your change and the default "errorLimit:0" all rules are shown.
image

Here is a showcase when i applied your change to my local code showing the issues
1identifieronly

BrianJavier99:
@lubber-de , sorry for the delay. With my recent [commits](https://github.com/fomantic/Fomantic-UI/pull/3067#commits-pushed-3a04945) I have tried addressing the cases that you have specified.

Below are the new behaviors.

Not Inline, error limit 0:
image

Not Inline, error limit 1:
image

Inline, error limit 0:
image

Inline, error limit 1:
image

Please let me know if I have missed any scenario or failed to consider a certain feature/functionality. Thank you.

lubber-de:
Please fix the linting issues 😉

BrianJavier99:
While fixing linting issues, I have encountered some bugs:

A formerly valid field does not display the error prompt once it becomes invalid again:
image

Conversely, a formerly invalid field does not remove the error prompt once it becomes valid again:
image

Will try to fix these issues within this week.

BrianJavier99:
Hello again @lubber-de,

I was only able to get back on this today. It seems that the issue I have previously discovered arose due to the fact that the error prompts are added/removed to the field groups as a whole and it lacks the functionality to iterate on every field with a similar identifier. This issue is unrelated from this issue and is replicable in the develop branch (I have used [your fiddle](https://jsfiddle.net/lubber/2v0e3s4m/4/)).

I propose that we put the issue into another ticket and proceed with the merging of this PR should there be no other concerns anymore.

Please let me know what you think, thanks.

lubber-de (changes requested):
> I propose that we put the issue into another ticket and proceed with the merging of this PR should there be no other concerns anymore.

Agreed. 

Please see my small little comments

src/definitions/behaviors/form.js
```diff
@@ -948,7 +949,7 @@
                                     $prompt.css('display', 'none');
                                 }
                                 $prompt
-                                    .appendTo($fieldGroup)
+                                    .appendTo($fieldGroup.filter('.error'))
```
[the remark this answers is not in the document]
BrianJavier99: Fixed at commit 057659145833867b37231b120c45133fa61703d0

src/definitions/behaviors/form.js
```diff
@@ -1388,7 +1390,22 @@
...
                                     if (invalidFields.length > 0) {
                                         module.debug('Field is invalid', identifier, rule.type);
-                                        fieldErrors.push(module.get.prompt(rule, field));
+                                        var fieldError = module.get.prompt(rule, field);
+                                        if (!settings.inline) {
+                                            if (
+                                                // Always allow the first error prompt for new field identifiers
+                                                (!(identifier in formErrorsTracker)
+                                                // Also allow multiple error prompts per field identifier but make sure each prompt is unique
+                                                || !formErrorsTracker[identifier].includes(fieldError))
```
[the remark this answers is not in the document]
BrianJavier99: Fixed at commit 057659145833867b37231b120c45133fa61703d0

src/definitions/behaviors/form.js
```diff
@@ -1388,7 +1390,22 @@
...
                                     if (invalidFields.length > 0) {
                                         module.debug('Field is invalid', identifier, rule.type);
-                                        fieldErrors.push(module.get.prompt(rule, field));
+                                        var fieldError = module.get.prompt(rule, field);
+                                        if (!settings.inline) {
+                                            if (
+                                                // Always allow the first error prompt for new field identifiers
+                                                (!(identifier in formErrorsTracker)
+                                                // Also allow multiple error prompts per field identifier but make sure each prompt is unique
+                                                || !formErrorsTracker[identifier].indexOf(fieldError))
```
lubber-de, line 1399: This won't work as `indexOf` does not return a boolean value. infact if 0 is returned the fieldError is actually found at position 0
As you want to check if fieldError does not exists in the whole array, check for -1 then
```suggestion
                                                || formErrorsTracker[identifier].indexOf(fieldError) === -1)
```

lubber-de (changes requested):
The indexOf needs some adjustments 😉

src/definitions/behaviors/form.js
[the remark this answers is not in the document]
lubber-de: not completely. See https://github.com/fomantic/Fomantic-UI/pull/3067#discussion_r1685792602

src/definitions/behaviors/form.js
BrianJavier99: Thank you for pointing out my careless mistake, I did not see any problems while testing it out. I do agree that a stricter comparison is safer.

Applied the suggestion under commit [ae3e0ec](https://github.com/fomantic/Fomantic-UI/pull/3067/commits/ae3e0ecfed6706e2bc7bd66e6ab3ac2987c9a7f9).

src/definitions/behaviors/form.js
```diff
@@ -1388,7 +1390,22 @@
...
-                                        fieldErrors.push(module.get.prompt(rule, field));
+                                        var fieldError = module.get.prompt(rule, field);
+                                        if (!settings.inline) {
+                                            if (
+                                                // Always allow the first error prompt for new field identifiers
+                                                (!(identifier in formErrorsTracker)
+                                                // Also allow multiple error prompts per field identifier but make sure each prompt is unique
+                                                || formErrorsTracker[identifier].indexOf(fieldError) === -1)
+                                                // Limit the number of unique error prompts for every field identifier if specified
+                                                && (!errorLimit || formErrorsTracker[identifier].length < errorLimit)
```
lubber-de, line 1401: In case error limit is > 0 and also the identifier wasn't set yet, the .length access triggers 'cannot read .length from undefined'

```suggestion
                                                && (!errorLimit || (formErrorsTracker[identifier] || []).length < errorLimit)
```

lubber-de (changes requested):
Please see my hopefully last comment. You already have a similar  fetch for non existing identifier object key, but it needs to be checked earlier as well

src/definitions/behaviors/form.js
BrianJavier99: Done. See commit db538f5

lubber-de (approved):
LGTM

hugopeek:
@BrianJavier99 @lubber-de Thank you both for the effort :pray:

BrianJavier99:
@

> Hello again @lubber-de,
> 
> I was only able to get back on this today. It seems that the issue I have previously discovered arose due to the fact that the error prompts are added/removed to the field groups as a whole and it lacks the functionality to iterate on every field with a similar identifier. This issue is unrelated from this issue and is replicable in the develop branch (I have used [your fiddle](https://jsfiddle.net/lubber/2v0e3s4m/4/)).
> 
> I propose that we put the issue into another ticket and proceed with the merging of this PR should there be no other concerns anymore.
> 
> Please let me know what you think, thanks.

Hello, 

I have noticed that the issue found here is already in a separate ticket and has a PR already: #2797

Just putting it here for better tracking.

PR #3080, opened 2024-08-03
fix(types): correct transition behavior "set looping"

kutoman:
fixing the transition looping enabling call signature. The passed behavior gets interpreted as a nonexistent animation otherwise.


lubber-de (approved):
LGTM

lubber-de:
Docs updated by https://github.com/fomantic/Fomantic-UI-Docs/pull/509

PR #3082, opened 2024-08-04
fix(types): missing/wrong types for modal. dropdown, popup, transition

### LINKED ISSUE #3081: Modal Settings  Reference mismatch with typescript type definition
# Bug Report
the type definition [here ](https://github.com/fomantic/Fomantic-UI/blob/develop/types/fomantic-ui-modal.d.ts#L180) (`transition: string | Partial<Pick<TransitionSettings, keyof TransitionSettings>>`) does not match with
the example for the transition settings from https://fomantic-ui.com/modules/modal.html#/settings:
```js
{
    showMethod   : 'fade',
    showDuration : 200,
    hideMethod   : 'zoom,
    hideDuration : 500,
}
```

## Expected result
The keys from the documentation's example object are correct. The type definition should reflect that. Thus, something like:
`transition: string | ModalTransitionSettings`

lubber-de:
## Description
Some types fixes to reflect settings in modal, dropdown, popup and transition

## Closes
#3081

### DIFF
--- a/types/fomantic-ui-dropdown.d.ts
+++ b/types/fomantic-ui-dropdown.d.ts
@@ -216,6 +216,31 @@ declare namespace FomanticUI {
         (settings?: Partial<Pick<DropdownSettings, keyof DropdownSettings>>): JQuery;
     }
 
+    interface DropdownTransitionSettings {
+
+        /**
+         * Named animation show event to used.
+         * Must be defined in CSS.
+         */
+        showMethod: string;
+
+        /**
+         * Duration of the CSS show transition animation
+         */
+        showDuration: number;
+
+        /**
+         * Named animation hide event to used.
+         * Must be defined in CSS.
+         */
+        hideMethod: string;
+
+        /**
+         * Duration of the CSS hide transition animation
+         */
+        hideDuration: number;
+    }
+
     /**
      * @see {@link http://fomantic-ui.com/modules/dropdown.html#/settings}
      */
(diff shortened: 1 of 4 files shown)


lubber-de:
@prudho @KiddoV 
Could you please review and test my changes? I was wondering why the original approach always uses confusing definitions  like `Partial<Pick<DropdownSettings, keyof DropdownSettings>>` 
which sounds unnecessary as just pointing to "DropdownSettings" should be enough, isn't it? 
So for the transition settings in this PR, i just used the simple way and the parser does not complain, so i believe my change is good enough? Perhaps the `Pick...keyof` syntax is somehow needed?

KiddoV:
The use of `Partial<Pick<DropdownSettings, keyof DropdownSettings>>` is often employed to create a type that represents a subset of `DropdownSettings`, where all properties are optional. This is useful when you only want to update or provide some fields of the original type while leaving the others with default values. 

In your case, if you use `DropdownSettings` directly, it would **require all properties to be provided**, which may not be ideal if you only want to specify a few fields and keep the rest at their default values. 

I found myself doing something like this a lot:
```typescript
{
    transition: {
        showMethod: "fly left",
        hideMethod: "fly left"
        // I just leave the other fields alone and they will take their default values
    },
}
```

Using `Partial<Pick<DropdownSettings, keyof DropdownSettings>>` allows you to specify only the fields you want to change, while the rest remain optional and can default to their initial values.

Take a look at the `Toast` module:
https://github.com/fomantic/Fomantic-UI/blob/fb2d8fcbc3f4de690f6df6a75a76452a8bf7e1a2/types/fomantic-ui-toast.d.ts#L355

I think you should follow the approach used here...

lubber-de:
@KiddoV Thanks for clarification 👍🏼, i just changed the PR accordingly.

KiddoV:
This LGTM!

PR #3085, opened 2024-08-11
fix(types): clear cache parameter in search module is optional

kutoman:
Complies with respective JSDoc and it works like that


lubber-de (approved):
LGTM

PR #3094, opened 2024-09-05
fix(flag): update macedonia flag name

lubber-de:
## Description
According to https://en.wikipedia.org/wiki/North_Macedonia , the related country flags name needs correction

To avoid a breaking change, i still left the previous name as a second alias. This will be removed again in 2.10.0

## Closes
https://github.com/Semantic-Org/Semantic-UI/issues/7143


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/514

PR #3100, opened 2024-09-24
fix(slider): correct computing of gapRatio

Inselhopper:
While the distance between two labels is too short for text at every label, we find out if we use only ticks with no text (gapRatio > 1) and apply text at any other tick (gapRatio = 2), any third tick (gapRatio = 3) and so on


lubber-de:
:thinking: Having your changes applied, the sliders end position value is almost never shown anymore, so people have no idea of the maximum possible value.
The shown value also now totally differ on different sizes.
While this usecase could be of interest for others too, it should need a new setting because of the different behavior.

See the very noticable difference here:

Your changes:
https://jsfiddle.net/lubber/6bzcq2hx/3/
gapratioPR

Current FUI
https://jsfiddle.net/lubber/6bzcq2hx/2/
gapratio

lubber-de (changes requested):
See my comment above

Inselhopper:
The current code is buggy. You can see this if you use a prime number. (Anigif) https://jsfiddle.net/inselhopper/uczsp3w6/14/
You wrote: "The slider's end position value is almost never shown anymore" → that is true.
The last label is not shown because of the logic, not because of a bug.
How to fix this? Always showing the last label could lead to overlapping text.
Users could use tooltipConfig to show the last value.

You wrote: "Also, the shown value totally differs on different sizes, now." → it seems like that, but not really.
[Current FUI - JSFiddle - Code Playground.webm](https://github.com/user-attachments/assets/198a6897-05ef-4322-beec-6b9899ef3967)

prudho:
I wrote this code 5 years ago, and barely touched it since... But here's my two cents:
Yes, the actual code is not optimal. Handling some numbers (primes) will not trigger the autogap adjustment. The idea of using the ratio system is nice, but like @lubber-de said, the slider's end position isn't visible some times, and this disturb me. Your solution should be implemented, but only used when specified by a user (so we keep the actual setting by default).

lubber-de:
@Inselhopper I adopted your change into a new option `fixed` for the `autoAdjustLabels` Option and added a permanent max slider tick as desired by prudho and me ;) It's implemented into my slider PR #3111 , so  would close this PR in favor of the other .

When `autoAdjustLabels: 'fixed'` is set, it uses your approach but makes sure the before last label does not overlap to the always shown max tick label
See https://jsfiddle.net/lubber/6bzcq2hx/5/
gapratio2

lubber-de:
Closing, as your change was basically implemented by #3111 [as mentioned above](https://github.com/fomantic/Fomantic-UI/pull/3100#issuecomment-2438683445)

PR #3111, opened 2024-10-23
feat(slider): improve slider variants and gapratio

lubber-de:
## Description
This PR enhances the slider component, so it
- fixes some slider issues
	- tick color on inverted variant
	- tick position on vertical variant
	- wrong gapratio calculation when calculated number of labels is a prime number 
	- adds internal caching of otherwise often repeated calculations
- improves labelType "letter" option, 
	- which now supports a custom letter string or array (was hardcoded before) 
	- "increases" the letters in case more labels than given letters are needed (like the columns of a spreadsheet)
- adds another variant `vertical right aligned`
- adds an optional value `fixed` for `autoAdjustLabels`
- adds another option `highlightRange` to slightly highlight the selected area range
- supports the usage of `ui label` 

## Screenshot
image
image
image

## Testcase
https://jsfiddle.net/lubber/1tvumckq/48/

## Closes
#3103

PR #3113, opened 2024-10-25
feat(types): missing and new option for slider

lubber-de:
## Description

Missing and new types for slider matching the feature s of #3111


prudho (approved):
LGTM

PR #3133, opened 2024-12-15
feat(search): support remove icon

lubber-de:
## Description

Support remove icon for search module similar to dropdown module
Either via separate `remove icon` or by using a native `input[type="search"]` element

### Sidenote
I party reused code which was already implemented in 2015 but removed again (by accident or intentionally , only the original authors knows...)
That's why i also removed the left dead css code for `selection search` which never made it into any release and wasn't ever documented. 
Was commited by 4ea251490579a262f65d12ac75415e222b6ed8e0
... and immediatly removed again the same day by 6d34ae21e595d104b59481848ebf2f7873028dac

## Testcase
https://jsfiddle.net/lubber/m2k9gn8o/126/

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -203,9 +203,9 @@
                         }
                     },
                     remove: {
-                        click: function() {
+                        click: function () {
                             module.clear.value();
-                        }
+                        },
                     },
                     result: {
                         mousedown: function () {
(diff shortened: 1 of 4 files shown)


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/535

PR #3140, opened 2024-12-22
feat(build): customizable output filename

lubber-de:
## Description

Support a custom filename instead of hardcoded "semantic" (which is still default).
This also prepares for a later change for the default from "semantic" to "fomantic" and a quicker fallback


lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/523

PR #3141, opened 2024-12-23
feat(calendar): "get date" can return formatted date

dreaming-augustin:
```
$('.ui.calendar').calendar('get date');
$('.ui.calendar').calendar('get focusDate');
$('.ui.calendar').calendar('get startDate');
$('.ui.calendar').calendar('get endDate');

```
could so far only return a Javascript Date object.

Now, given a `format` argument, they return a formatted date string:

```
$('.ui.calendar').calendar('get date',      "YYYY-MM-DD");
$('.ui.calendar').calendar('get focusDate', "YYYY-MM-DD");
$('.ui.calendar').calendar('get startDate', "YYYY-MM-DD");
$('.ui.calendar').calendar('get endDate',   "YYYY-MM-DD");

```

The `format` string is passed to the existing formatter.

This is backward compatible as the default behaviour does not change.

It makes it convenient for developers to get exactly what they need when they need it.

### DIFF
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -1145,7 +1145,7 @@
                     },
                     dateObjectOrFormatted: function (format, date) {
                         format = format || '';
-                        date = module.helper.sanitiseDate($module.data(metadata.date)) || null;
+                        date = module.helper.sanitiseDate(date) || null;
 
                         if (!date) {
                             return null;
--- a/src/definitions/modules/calendar.js
+++ b/src/definitions/modules/calendar.js
@@ -802,28 +802,32 @@
                     formattedDate: function (format, date) {
                         return module.helper.dateFormat(format || formatter[settings.type], date || module.get.date());
                     },
-                    date: function (format = "") {
+                    date: function (format = '') {
                         if (!module.helper.sanitiseDate($module.data(metadata.date))) {
                             return null;
                         }
-                        else if (format == "") {
+
+                        if (format === '') {
                             return module.helper.sanitiseDate($module.data(metadata.date));
                         }
+
                         return module.helper.dateFormat(format, $module.data(metadata.date));
                     },
                     inputDate: function () {
                         return $input.val();
                     },
-                    focusDate: function (format = "") {
+                    focusDate: function (format = '') {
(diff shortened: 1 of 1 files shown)


src/definitions/modules/calendar.js
```diff
@@ -802,24 +802,64 @@
...
+
+                        if (!module.helper.sanitiseDate($module.data(metadata.endDate))) {
+                            return null;
+                        }
+
+                        if (format === '') {
+                            return module.helper.sanitiseDate($module.data(metadata.endDate));
+                        }
+
+                        return module.helper.dateFormat(format, $module.data(metadata.endDate));
```
lubber-de, line 862: 1. Default function parameter values are not supported in IE11 (which FUI 2.9.x still supports)
2. I would reduced the call to the sanitisedate function
```suggestion
                        format = format || '';
                        if (endModule) {
                            return endModule.get.date(format);
                        }
                        
                        var endDate = module.helper.sanitiseDate($module.data(metadata.endDate)) || null;
                        if (!endDate) {
                            return null;
                        }

                        if (format === '') {
                            return endDate;
                        }

                        return module.helper.dateFormat(format, endDate);
```

lubber-de (changes requested):
Please see my comments / suggestions.
The new code is now somewhat duplicated four times (using a different metadata value) we should even think of putting that in a unique function

dreaming-augustin:
> 1. Default function parameter values are not supported in IE11 (which FUI 2.9.x still supports)

Ok. I am fluent in JavaScript. I think more in term of C++, so I couldn't imagine parameters that were undefined.

dreaming-augustin:
> The new code is now somewhat duplicated four times (using a different metadata value) we should even think of putting that in a unique function

 Done. You make suggest another function name...

lubber-de (approved):
Thanks, but cleaner! LGTM
Please take a short look at your docs PR where i was already commenting https://github.com/fomantic/Fomantic-UI-Docs/pull/524

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/524

dreaming-augustin:
Thank you for your reviews.
I apologize for dumping so many PR just before Christmas!  
I hope you had/are having great holidays.

PR #3142, opened 2024-12-23
[Search] Inject id and class attr into individual result

dreaming-augustin:
Give much more flexibility on the type of content and theming possibilities for each individual search result.

We could already manipulate the response received by the Search API:

```
onResponse: function(original_response) {
    var response = {
       results : {}
    };
    $.each(original_response.results, function(index, item) {
          // Manipulate each item.
          // Possibly add item.id
         // Possibly add item.classes
    }
    return response;
}
```

The problem is that we couldn't inject the `id` and `class` attributes to each individual item rendered by the Search module.

Now, depending on the status of each response, the status potentially reflecting a given website's logic and needs, we can inject an id and additional classes that can then be used for theming and manipulating.

Also, before each item was wrapped within a `<a>` tag. Because HTML prohibits having nested `<a>` tags (which makes sense), we couldn't inject any link into the body of individual items (e.g. inject relevant links to open in a different browser tab, while keeping the result popup menu open).

Now, if the result item has `result[fields.url]`, then keep the `<a>` tag, because it makes sense: the whole result is assigned an URL.
If `result[fields.url]` is absent, the result is wrapped within a `<div>` tag, making it possible to inject one or even several links into the result's description, as may be required.

This feature is backward-compatible, as the default behavior is unchanged.


dreaming-augustin:
If requested, I can submit a PR for documentation later.

dreaming-augustin:
Here are two examples of theming achieved with this new feature: 

Fomantic-search-classes-PR3142-2024-12-23-1
Fomantic-search-classes-PR3142-2024-12-23-2

src/definitions/modules/search.js
```diff
@@ -1611,7 +1632,10 @@
                         }
                         html += ''
                             + '</div>';
-                        html += '</a>';
+
+                        html += result[fields.url]
+                            ? '</a>'
+                            : '</div>';
```
lubber-de, line 1638: See above, this should be reverted to always create an achor `a` tag

src/definitions/modules/search.js
```diff
@@ -1591,8 +1603,17 @@
...
-                            ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                            : '<a class="result">';
+                            ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                            : '<div ';
+
+                        html += result[fields.id] !== undefined
+                            ? ' id="' + result[fields.id] + '" '
+                            : '';
+
+                        html += result[fields.classes] !== undefined
```
lubber-de, line 1613: we should rename the fields property to `class` ins tead of `classes`

src/definitions/modules/search.js
```diff
@@ -1538,8 +1538,17 @@
...
-                                    ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                                    : '<a class="result">';
+                                    ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                                    : html += '<div ';
+
+                                html += result[fields.id] !== undefined
+                                    ? ' id="' + result[fields.id] + '" '
+                                    : '';
+
+                                html += result[fields.classes] !== undefined
```
lubber-de, line 1548: we should rename the fields property to `class` ins tead of `classes`

src/definitions/modules/search.js
```diff
@@ -1538,8 +1538,17 @@
                             html += '<div class="results">';
                             $.each(category.results, function (index, result) {
                                 html += result[fields.url]
-                                    ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                                    : '<a class="result">';
+                                    ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                                    : html += '<div ';
```
lubber-de, line 1542: Not needed and a breaking change, keep using the anchor tag

src/definitions/modules/search.js
```diff
@@ -1558,7 +1567,10 @@
                                 }
                                 html += ''
                                     + '</div>';
-                                html += '</a>';
+
+                                html += result[fields.url]
+                                    ? '</a>'
+                                    : '</div>';
```
lubber-de, line 1573: See above and revert. Keep using the anchor tag

src/definitions/modules/search.js
```diff
@@ -1459,6 +1457,8 @@
             results: 'results', // array of results (standard)
             title: 'title', // result title
             url: 'url', // result url
+            id: 'id', // HTML 'id' attribute
+            classes: 'classes', // Classes specific to each result to add to the HTML 'class' attribute.
```
lubber-de, line 1461: We should name the property `class`  instead of `classes`

lubber-de (changes requested):

src/definitions/modules/search.js
```diff
@@ -1538,8 +1538,17 @@
...
-                                    : '<a class="result">';
+                                    ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                                    : html += '<div ';
+
+                                html += result[fields.id] !== undefined
+                                    ? ' id="' + result[fields.id] + '" '
+                                    : '';
+
+                                html += result[fields.classes] !== undefined
+                                    ? ' class="result ' + result[fields.classes] + '">'
```
lubber-de, line 1549: needs to be sanitized (double quotes removed)
```suggestion
                                    ? ' class="result ' + result[fields.classes].replace(/"/g, '') + '">'
```

src/definitions/modules/search.js
```diff
@@ -1538,8 +1538,17 @@
                             html += '<div class="results">';
                             $.each(category.results, function (index, result) {
                                 html += result[fields.url]
-                                    ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                                    : '<a class="result">';
+                                    ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                                    : html += '<div ';
+
+                                html += result[fields.id] !== undefined
+                                    ? ' id="' + result[fields.id] + '" '
```
lubber-de, line 1545: Needs to be sanitzed (doubple quotes removed
```suggestion
                                    ? ' id="' + result[fields.id].replace(/"/g, '') + '" '
```

src/definitions/modules/search.js
```diff
@@ -1591,8 +1603,17 @@
                     // each result
                     $.each(response[fields.results], function (index, result) {
                         html += result[fields.url]
-                            ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                            : '<a class="result">';
+                            ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                            : '<div ';
+
+                        html += result[fields.id] !== undefined
+                            ? ' id="' + result[fields.id] + '" '
```
lubber-de, line 1610: Needs to be sanitized (double quotes removed)
```suggestion
                            ? ' id="' + result[fields.id].replace(/"/g, '') + '" '
```

src/definitions/modules/search.js
```diff
@@ -1591,8 +1603,17 @@
...
+                            : '';
+
+                        html += result[fields.classes] !== undefined
+                            ? ' class="result ' + result[fields.classes] + '">'
```
lubber-de, line 1614: Needs to be sanitized (double quotes removed)

lubber-de (changes requested):

lubber-de:
Seeing my comments, i currently disagree (or at least saying it's a breaking change) to the "use a div if no url exists" logic.
You can already provide a complete custom render template function as documented in https://fomantic-ui.com/modules/search.html#templates
Yes, this might need an improved docs page and perhaps an example as well

src/definitions/modules/search.js
```diff
@@ -1591,8 +1603,17 @@
                     // each result
                     $.each(response[fields.results], function (index, result) {
                         html += result[fields.url]
-                            ? '<a class="result" href="' + result[fields.url].replace(/"/g, '') + '">'
-                            : '<a class="result">';
+                            ? '<a href="' + result[fields.url].replace(/"/g, '') + '" '
+                            : '<div ';
```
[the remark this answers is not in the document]
lubber-de: As the next version allows for breaking changes, i now agree :)

ko2in (approved):
LGTM.

lubber-de (approved):

lubber-de:
@dreaming-augustin I finally agreed, fixed my requested changes in your PR and merged it.
As you suggested: If you like, you may create a proper docs example PR accordingly

Issue #3143, opened 2024-12-24
[Search] Bring active result into view


dreaming-augustin:
Search supports `scrolling search` since the following: 
> 
> [Search] Scrolling Results #771  
> [Enhancement]: [Search] Scrollable search result #1350 
> 

However, this does not fix a common scenario, depending on the business logic of a website, whereby active search items are now brought into view.

Put the search box at the bottom of the page. When the search results come in, they open in a drop-down menu _below_ the input field, _out of the viewport, out of view_. 
What the user **must then do** is grab the mouse, scroll the window and bring the fields of results into view. This is true even with a `scrolling search` or even when the result set is short.
What the user **cannot do** is, after having typed their query, use the keyboard down-arrow to select the entry they want, because they are outside the viewport.

Depending on the website needs and the business logic, that can be a very common scenario.

A real life scenario that I am currently trying to fix is for a business workflow that is very data-entry intensive. By necessity, and in accordance with the workflow, the search input field is at the bottom of the screen. Since there is a lot of data to search and enter, we try to make the process as smooth and painless as possible. It means we try to have a data-entry workflow that does not require grabbing the mouse and scrolling down the page for each single search. The most efficient workflow requires a keyboard-only interaction: type query => use keyboard down-arrow to select the wanted result => press enter to validate => repeat by typing next query.

So, basically, it is necessary to implement a hack that will **_automatically bring the active search result item into view_**.

This is a long-standing issue and other users have asked for it in the past, for Semantic-UI:

> 
> [Dropdown] How to receive an event when scrolled down to to bottom? https://github.com/Semantic-Org/Semantic-UI/issues/6166 
> 
> [Search] Option to set max-height for search results https://github.com/Semantic-Org/Semantic-UI/issues/6488

dreaming-augustin:
Within the pull request, I shared a solution (hack) to solve the problem above, that depends on the following pull request: 
 [Search] trigger 'focusin', 'focusout'. #3144

PR #3144, opened 2024-12-24
[Search] trigger 'focusin', 'focusout'.

dreaming-augustin:
Trigger 'focusin' and 'focusout' whenever a search result item becomes active.

It allows websites to implement a solution for this:

>  
> [Search] Bring active result into view #3143
>

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -329,8 +329,10 @@
                             ;
                             $result
                                 .removeClass(className.active)
+                                .trigger('focusout')
                                 .eq(newIndex)
                                 .addClass(className.active)
+                                .trigger('focusin')
                                 .closest($category)
                                 .addClass(className.active)
                             ;
@@ -346,8 +348,10 @@
                             ;
                             $result
                                 .removeClass(className.active)
+                                .trigger('focusout')
                                 .eq(newIndex)
                                 .addClass(className.active)
+                                .trigger('focusin')
                                 .closest($category)
                                 .addClass(className.active)
                             ;


dreaming-augustin:
This patch at least allows for a partial solution to  [Search] Bring active result into view #3143  

It allows websites to implement their own solution, according to their needs.
A website in now able to implement something like the following on their own:

```
        // Scroll  search results into view.
        $('#search').on('focusin', '.result', function() {
                var offset = $(this).offset();
                var viewportTop    = $(window).scrollTop();
                var viewportHeight = $(window).height();
                var viewportBottom = viewportTop + viewportHeight;
                var elementTop     = $(this).offset().top;
                if (elementTop > (viewportBottom - 150)) {
                        offset.top = elementTop - (viewportHeight / 4);
                        $('html, body').animate({
                                scrollTop: offset.top,
                        }, 200);
                }
                if (elementTop < (viewportTop + 150)) {
                        offset.top = elementTop - (3 * viewportHeight / 4);
                        $('html, body').animate({
                                scrollTop: offset.top,
                        }, 200);
                }
        });

```
This feels a bit hackish and too specific. It can be discussed whether something more appropriate could be considered to be included within Fomantic.

At the very least, triggering `focusin()`, developers have more options to deal with their own business needs.

src/definitions/modules/search.js
```diff
@@ -329,8 +329,10 @@
                             ;
                             $result
                                 .removeClass(className.active)
+                                .trigger('focusout')
```
lubber-de, line 332: Do we really need this? The direct focusin trigger 3 lines below should automatically trigger the focusout form the previous focused element, isnt it?

lubber-de (commented):
While this change does not seem to hurt the code (check my comment if focusout is really needed if we add focusin at the same time),
i wonder if this is really needed. The user could already make use of the mutationobserver and check for the added ".active" class

lubber-de:
Btw we should not implement automatic scrolling of the core page into the FUI Code. I believe this will mess up exsiting projects in unexpected scenarios

dreaming-augustin:
> Btw we should not implement automatic scrolling of the core page into the FUI Code. I believe this will mess up existing projects in unexpected scenarios

Of course not. I said so myself. The code I shared within a comment is a hack that works well in a very specific scenario where I needed it. I shared it so that interested Fomantic users can adapt it to their needs. But it is too specific to be included in Fomantic itself. 

... Now, if the community can come up with something more general and acceptable for Fomantic, I'd be happy to test it out...

dreaming-augustin:
> i wonder if this is really needed. The user could already make use of the mutationobserver and check for the added ".active" class

I will investigate and come back to you soon.

dreaming-augustin:
I didn't know the existence of MutationObservers... 
I will learn: 
https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver

PR #3160, opened 2025-01-31
Fix "$tab.activate.tab" called after renavigate

### LINKED ISSUE #3156: [Tab] Async context must not replace different tab
# Bug Report

## Steps to reproduce
1. open https://dev.agiletoolkit.org/demos/interactive/tabs.php
2. click "Dynamic Form" tab
3. click (quickly - before the AJAX completes) other/"Hello" tab
4. notice when the AJAX load of "Dynamic Form" tab finishes, the "Hello" tab is replaced

## Expected result

When tab is switched, any ongoing async load must be canceled.

## Actual result

Tab content must never be replaced with content of different tab.

## Version

2.9.3/latest

mvorisek:
fix #3156

Currently, the only "fetch" is invoked after https://github.com/fomantic/Fomantic-UI/blob/13b76ee714/src/definitions/modules/tab.js#L367 thus the "navigation" should be always active when the tab "is beiing activated and loading".

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -475,11 +475,14 @@
                                     'X-Remote': true,
                                 },
                                 onSuccess: function (response) {
+                                    let $navigation = module.get.navElement(tabPath);
+                                    let isNavigationActive = $navigation.hasClass(className.active);
+
                                     if (settings.cacheType === 'response') {
                                         module.cache.add(fullTabPath, response);
                                     }
                                     module.update.content(tabPath, response);
-                                    if (tabPath == activeTabPath) {
+                                    if (isNavigationActive && tabPath == activeTabPath) {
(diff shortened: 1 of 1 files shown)


mvorisek:
Hi @lubber-de, might I have your review?

lubber-de:
> Hi @lubber-de, might I have your review?

Quickly looking at the logic, the activeTabPath should be set to the latest clicked tab, so whenever it changes in between the ajax request, tabPath should not be equal to activeTabPath anymore, thus isn't active, so the additional check _should_ not be needed.
However, given your examples, it doesn't seem to work as expected as especially the wrong tab gets updated.

Yes, also the later check for the active class seems not quite right as you pointed out.

I have to debug all of this using some simplfied testcase myself. I suspect something else / more we need to fix.

PR #3161, opened 2025-01-31
No "X-Remote" header in Tab AJAX by default

mvorisek:
This is documented in the docs - https://github.com/fomantic/Fomantic-UI-Docs/blob/5ad17488bc/server/documents/modules/tab.html.eco#L644.

But I belive this concept is not unified across the whole Fomantic UI and forcing an extra HTTP header is not wanted in general.

It is always possible to detect the AJAX load like in https://github.com/atk4/ui/blob/5.2.0/src/App.php#L343 if header approach is wanted and it is guaranteed by the XMLHttpRequest spec.

If any special handling for "tab load" is wanted, "apiSettings.data" should be used instead like in https://github.com/atk4/ui/blob/5.2.0/src/TabsTab.php#L39. This has an advantage to work even thru proxies which can strip unknown headers easily.


lubber-de:
While your arguments are valid and these days the backend can check "X-Requested-With" instead, i'll tag this as breaking change for 2.10 as peoples code might still rely on it.
Other components, like dropdown and search, do not use this header as the related backend for such components is supposed to always being used for ajax requests, but tab content could be whole HTML code, so the extra header was implemented only for tabs

We also should change the docs for a compatible workaround by using 

```javascript
$('tabselector').tab({
	apiSettings: {
	      headers: {
	          'X-Remote': true,
	      },
	}
});
```

lubber-de (approved):
LGTM

Issue #3152, opened 2025-01-12
[Dropdown] Unify non-focused UX of dropdown


mvorisek:
# Feature Request

Currently, when an `input` is out-of-focus all icons are grayed (have limited opacity), this is expected:

![image](https://github.com/user-attachments/assets/335b699a-b50e-4637-9a8d-dc3e03dde750)

However dropdown icons/controls are not:

![image](https://github.com/user-attachments/assets/ff3f77b1-0414-4591-8b0b-ffff20d28bde)

I would be happy if dropdown would gray their icons as well to improve the UX. Currently the "dropdown/arrow" and "remove" icons are too distracting when page has a lot of dropdown controls.

demo: https://dev.agiletoolkit.org/demos/form-control/input2.php

And one minor related issue: when dropdown is focused with tab, the dropdown should stay closed (it already does) but should be displayed as focused, ie. with blue neon border.

lubber-de:
I think, this is implemented/fixed by #3165

PR #3165, opened 2025-02-05
fix(dropdown, form): unify focus and hover borders and icons

lubber-de:
## Description

Unify border and icon visuals between input and dropdown

- fix search selection dropdowns not activating focus border when tabbed
- add hover border to inputs just as dropdown has
- same icon opacity logic

## Testcase
https://jsfiddle.net/lubber/jkLqbd1u/55/

## Closes
https://github.com/fomantic/Fomantic-UI/issues/3152

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -640,6 +640,11 @@
             box-shadow: @formStates[@@state][inputFocusBoxShadow];
         }
 
+        .ui.form .field.@{state} > textarea:hover:not(:focus),
+        .ui.form .field.@{state} > input:hover:not(:focus) {
+            border-color: @formStates[@@state][inputHoverBorderColor];
+        }
+
         /* Preserve Native Select Stylings */
         .ui.form .field.@{state} select {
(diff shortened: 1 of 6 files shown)


mvorisek:
image

The regular `select` is still natively formatted.

lubber-de:
> The regular `select` is still natively formatted.

Thanks for the hint. native select support isn't focussed by this PR. 
However, i am already working on another PR supporting native elements styling without the use of JS and will support native dropdown then as well

lubber-de:
@mvorisek Do you think we should release this for 2.9.4 or should we wait for the native select element to be styled as well (this will now be postponed to 2.10, so this PR would be postponed as well then)

mvorisek:
I am not sure if I am the right person to decide this, but I am fine to land this now. "native select styling" can be landed later.

mvorisek:
I just tested v2.9.4. Big thank you for this release!

Just one minor finding:

image

the dropdown caret icon should be dark when hovered the same way as the clearable icon is. repro: https://jsfiddle.net/2ut0ypxh/

PR #3168, opened 2025-02-08
fix(deps): remove re2 and url-regex through rework-import

### LINKED ISSUE #3167: [build] Local Python is currently required to be locally available
# Bug Report

I tried to update Fomantic UI code for atk4/ui and I found some deps were raised and the deps caused local Python is currently needed to pass a build!

```
npm ERR! > re2@1.20.5 rebuild
npm ERR! > node-gyp rebuild
npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp info using node-gyp@9.4.1
npm ERR! gyp info using node@19.8.1 | win32 | x64
npm ERR! gyp ERR! find Python
...
```

I figured out it is caused by:

https://github.com/fomantic/Fomantic-UI/blob/2f37293a27/package.json#L47
->
https://github.com/fomantic/gulp-concat-css/blob/3.2.1/package.json#L27
->
https://github.com/fomantic/rework-import/blob/3.1.0/package.json#L31
->
https://github.com/uhop/node-re2/blob/1.20.5/package.json#L16

This is very problematic for my build machine as I am on Windows without Python available on my setup.

I belive there is a way to not require `re2` dependency and use PCRE/JS regexes instead.

## Steps to reproduce

```
{
    "name": "atk-external",
    "private": true,
    "dependencies": {
        "fomantic-ui": "^2.9.4-beta"
    }
}
```

## Expected result

No Python nor other executable required as before.

## Version

2.9.4 nightly / current develop

lubber-de:
## Description

I published a new version of[ fomantic/rework-import](https://github.com/fomantic/rework-import) (through [gulp-concat-css](https://github.com/fomantic/gulp-concat-css)) which gets rid of re2 and url-regex(-safe) dependency.
The only need was to check a possible prototcol inside the import string. It does not matter if the whole URL is valid, as rework-import only needs to know if the import is a local file or not. This logic was already the case in 2.0 of the original rework-import, so basically there never was a need to switch to a separate package just for that check.

By doing all of this, fomantic does not need re2 dependency anymore which itself needs an installed python instance which especially on windows is not available by default.

## Closes
#3167

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "gulp-clean-css": "^4.3.0",
     "gulp-clone": "^2.0.1",
     "gulp-concat": "^2.6.1",
-    "@fomantic/gulp-concat-css": "^3.2.1",
+    "@fomantic/gulp-concat-css": "^3.3.0",
     "@fomantic/gulp-concat-filenames": "^1.2.1",
     "@fomantic/gulp-dedupe": "^0.0.3",
     "gulp-flatten": "^0.4.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -112,13 +112,13 @@
     chalk "^4.1.2"
     cli-table "^0.3.11"
 
-"@fomantic/gulp-concat-css@^3.2.1":
-  version "3.2.1"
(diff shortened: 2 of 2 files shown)


mvorisek:
I can confirm https://github.com/fomantic/rework-import/commit/28af4c117f23d5b9c99caa632673145a7ccbe6d1 fixed the #3167 and by side effect it also removed a lot of extra deps from the lock file! Thank you!

PR #3086, opened 2024-08-12
feat(emoji): update to emoji v15.1

lubber-de:
## Description
This PR updates the emoji component to support emoji 15.1 characters by upgrading to the twemoji fork by @jdecked and optional joypixels v9 theme

I stumpled upon this article https://www.reddit.com/r/opensource/comments/15xalsm/the_fate_of_twemoji to recognize former twitter employees (laid off by the time twitter became x) created the fork to support  emoji 15.1 characters

PR #3169, opened 2025-02-08
fix(flag): use jdecked twemoji fork

mvorisek:
I belive this was missed in #3086.

### DIFF
--- a/src/themes/default/elements/flag.variables
+++ b/src/themes/default/elements/flag.variables
@@ -5,7 +5,7 @@
 /* --------------
     Path
 --------------- */
-@flagPath: "https://cdn.jsdelivr.net/gh/twitter/twemoji@latest/assets/svg/";
+@flagPath: "https://cdn.jsdelivr.net/gh/jdecked/twemoji@latest/assets/svg/";
 @flagFileType: "svg";
 
 /* -------------------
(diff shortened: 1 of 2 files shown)


lubber-de:
This was basically untouched intentionally

- Flag did not receive a change/update in 15.1 (no flags updated/added)
- twitter theme is still meant for the original twitter (14) emojis. If we just change the url, the v15 emojis still arent available as the new emojis are not liste inside the twitter/emoji.variables file

So:

- The extra twitter theme (for emoji) wouldn't make sense anymore (perhaps never was) it it's always the same as the default themes emojis, we might either drop it (breaking change?) or leave it as it is, so it remains the "real" twitter theme
- For flag, we could/should change the repo url to at least share the same source as the default emojis

lubber-de (approved):
LGTM

mvorisek:
Twitter theme - flags can this be used there, so I would say it should be upgraded as well for unified UX as long as no flag was removed. But I reverted this change.

Flags - this should definitely be upgraded. And there is actually change in one flag 😀:
image

before:
image

after:
image

(height has changed)

The reason for this PR is: https://github.com/atk4/ui/blob/5.2.0/public/external/postinstall.js#L151 - in atk4/ui we bundle all the flags so the atk4/ui can work fully offline.

PR #3191, opened 2025-03-04
feat(chore): new default settings monthFirst and preserveHTML

lubber-de:
## Description
This PR changes default values as discussed in #3148


mvorisek:
As the develop is already open for 2.10.x, I would prefer this to be merged soon, so it is included in the beta releases asap.

If another settings need to be updated, they can be landed later incl. separate entry in changelog.

lubber-de:
@mvorisek Agreed, so people can probably adjust their code while beta testing as early as possible

PR #3200, opened 2025-03-08
feat(header): rename to heading

lubber-de:
## Description

As discussed in #2407 this PR renames `header` to `heading`.

- rename all `.ui.header` selectors to `.ui.heading`
- rename all `@header...` LESS variables to `@heading...`
- rename all related files for building to `heading` as well (so header.less and header.variables dont exists anymore)

A new variation variable is introduced: `@variationHeadingLegacyHeader` (default false)
When set to true, it will generate the old `.header` classes in addition to the `.heading` classes for easier migration (but will double all selectors which increases css filesize)

### This is a __HUGE BREAKING CHANGE__ as

-  custom themes won't work when they override `@header...` variables as those do not exist anymore
- headings won't be built anymore unless "header" is renamed to "heading" in each custom `theme.config` and each custom `semantic.json`
- existing projects wont render headings properly unless all `header` usages are replaced by `heading` class usages

### Still to be decided / taken into consideration

I leave this PR as draft for now as we still have to check/decide how to deal with other "header" usages in other components:

- The `th` tag in a table is still a "header", not a "heading". However the table component  shares the same `@heading` variables as the heading element itself. This is inconsistent, i think there are dedicated new variables needed
- There are accordion, popup, message, toast, modal, flyout, list item, menu item, ... which all have/use a `header` (which either is a "heading",  a "title" or still a "header"

If this gets merged, we may announce all upcoming breaking changes  (espcially this one) long before 2.10.0 gets released, so people are prepared.


ko2in (commented):
The following themes still have old file names `header.overrides` and `header.variables`:

- classic
- github
- material

lubber-de:
@ko2in Thanks! fixed in f66a9fb95774590b0d6e19e0f940c0161e9fb0ee

ko2in:
LGTM.

PR #3201, opened 2025-03-08
feat(menu): rename tabular to tabbed

lubber-de:
## Description

As discussed in #2407 this PR renames `tabluar menu` to `tabbed menu`.

- rename all `.ui.tabular.menu` selectors to `.ui.tabbed.menu`
- rename all `@...tabular...` LESS variables to `@...tabbed...`

A new variation variable is introduced: `@variationMenuTabbedLegacyTabular` (default false)
When set to true, it will generate the old `.tabular` classes in addition to the `.tabbed` classes for easier migration (but will double all selectors which increases css filesize)

### This is a __BREAKING CHANGE__ as

- custom themes won't work when they override `@...tabular..` variables as those do not exists anymore
- existing projects won't render tab menus properly unless all `tabular` usages are replaced by `tabbed` class usages


src/definitions/collections/menu.less
```diff
@@ -629,156 +629,156 @@ Floated Menu / Item
...
+            color: @tabbedActiveColor;
+            border-color: @tabbedBorderColor;
+            margin: 0 0 0 -@tabbedBorderWidth;
+            border-radius: 0 @tabbedBorderRadius @tabbedBorderRadius 0 !important;
         }
     }
 
     /* Dropdown */
-    .ui.tabular.menu .active.dropdown.item {
+    .ui.tabbed.menu .active.dropdown.item {
```
ko2in, line 767: Do we also use dropdown with tab menu? I couldn't find any example.
tabbed-menu-with-dropdown

I tried to manipulate dropdown with a tab menu. When it's in active state, the dropdown is not able to open, and the border is always missing. Is this the correct behavior?

JSFiddle: https://jsfiddle.net/ko2in/5f24pm0s/1/

lubber-de: I think, the idea is/was to support a dropdown menu item in the same menu next to tabs but without being a tab item itself.
See here: https://jsfiddle.net/lubber/7f09ubtw/1/

ko2in (approved):
LGTM.

lubber-de:
Docs added by https://github.com/fomantic/Fomantic-UI-Docs/pull/543

Issue #2817, opened 2023-06-08
'multiple' Dropdown Refreshing on First Selection or emptying selection of option in Dropdown


PMSD:
# Bug Report
While making use of multiselect dropdowns created from a select element I noticed that upon the first selection of an item, the entire dropdown and all of its items were refreshing. I noticed this because I was making use of the parentNode of the item selected in an event listener and when I recently updated fomantic, the event listener broke.

So I did some digging. I set my dropdown config debug and verbose to true and I ran the older version fomantic and compared it to the newer. I found that in the new version, the dropdown was refreshing 3 times, where as in the old it wasn't refreshing at all. I went digging through the code and I believe it has to do with the selectObserver in dropdown.js
What I found was that attributes were added to a select mutation observer. Now whenever an attribute changes on the underlying select element, the whole dropdown is refreshed.

So for instance, whenever the class of the select is changed from "noselection" to empty or vice versa it refreshes the whole dropdown. 
As in my case, where I... 
1.) select one element
2.) the select goes from "noselection" to empty class
3.) the dropdown is refreshed over and over
4.) poor little item is orphaned
5.) my evenListener sees null for parentNode

I have a feeling this isn't desired behavior and may be causing other problems with 'multiple' dropdowns; but, I'm not familiar enough with the code base to say.

## Steps to reproduce
1. create a select element with items
2. call dropdown on the select
3. select an item from the dropdown
4. see in the element hierarchy that the dropdown refreshes and all of the items are refreshed

## Expected result
1. I would expect that the dropdown shouldn't totally refresh just because the class changed on the select element. 

## Actual result
1. The dropdown is refreshed 3 times when the select element's class changes

## Testcase
//THIS TEST IS WITH 2.9.2
https://jsfiddle.net/wga63zv9/

//THIS TEST IS WITH 2.8.8
https://jsfiddle.net/f46roL25/

## Screenshot (if possible)
dropdown debug from 2.9.2
![fomantic 3 refresh](https://github.com/fomantic/Fomantic-UI/assets/5914488/b6dd88a8-c3bc-4c15-9a17-af596fd6b996)

----------------------------------------------------------------------------------------------------------------------------------------------
dropdown debug from 2.8.8
 ![fomantic no refresh](https://github.com/fomantic/Fomantic-UI/assets/5914488/92329f6b-e120-408e-a8a9-26109bc62635)

I believe this addition of attributes to the selectObserver.observe is the culprit
[...]

PR #2790, opened 2023-05-15
fix(dropdown): encoded select values, multiple maxselections

lubber-de:
## Description

Some dropdown fixes:
a) multiple and maxselections allowed more than the limit to be selected when labels are used
b) uselabels:false + saveremotedata:true und preselected values created labels
c) init from select and preselected value including an ampersand wasn't preselected

## Testcase
a)
### Broken
- Open the dropdown via click on dropdown icon. Although the selection should be limited to 2 entries, you can add more entries
- Clicking inside the dropdown does nothing
https://jsfiddle.net/lubber/1oeLn7h6/5/

### Fixed
- Clicking inside the dropdown as well as opening it from the dropdown icon just shows the "max 2 selected values" as expected
https://jsfiddle.net/lubber/1oeLn7h6/6/

b) 
### Broken
The dropdown shows 2 preselected labels, although labels should not be used
https://jsfiddle.net/lubber/c8eub9fo/2/

### Fixed
"2 selected" shown as expected
https://jsfiddle.net/lubber/c8eub9fo/3/

c) 
### Broken
Dropdown made out of a select does not create the preselected labels which contains an ampersand
https://jsfiddle.net/lubber/qykf58tc/3/

### Fixed
Preselected as expected
https://jsfiddle.net/lubber/qykf58tc/2/

## Closes
#2782

Issue #2782, opened 2023-05-09
[Dropdown] Multiple dropdown doesn't show selected label if the selected value has ampersand "&"


gfkateo:
# Bug Report
When creating a multiple dropdown using `<select>` element and the selected option's value contains ampersand, the label for the selected value doesn't show.

## Expected result
Should show the label of the selected option 

![Screenshot from 2023-05-09 16-33-50](https://user-images.githubusercontent.com/100657275/237040985-2e852575-3bf0-4c26-b4d8-31ae413b2923.png)

## Actual result
 
![Screenshot from 2023-05-09 16-21-34](https://user-images.githubusercontent.com/100657275/237037922-f998d53c-e6b0-41d4-9264-f593c53f1175.png)

Ran this code in the console
```javascript
$('.ui.dropdown.sample1').dropdown('get value')
```
and it returns the correct value
```
["b & c"]
```

Also, I've tested creating a dropdown with using `<div class="ui multiple selection dropdown">` element, and it doesn't have this issue.  (See Sample 3 field in the Codepen testcase)

## Testcase
[Codepen URL](https://codepen.io/katedo/pen/ExdENQW)

### HTML
```html
<div class="field">
  <label>Sample 1: Using select element with ampersand</label>
  <select name="values1" class="ui fluid dropdown sample1" multiple>
    <option value="">Select at least one</option>
    <option value="a">A</option>
    <option value="b & c" selected>B & C</option>
    <option value="d">D</option>
  </select>
</div>
```

### Javascript
```javascript
$('.ui.dropdown').dropdown();
```

## Version

2.9.2

ko2in:
I can't reproduce that problem in your codepen.

![fui-ampersand-in-dropdown](https://github.com/fomantic/Fomantic-UI/assets/930315/3f7e84ec-be38-4914-b0bf-c475ab5d6192)


What is your browser and it's version?

lubber-de:
@ko2in The issue is, that the entry containing the ampersand has got the "selected" attribute set. So it should be preselected / and the label should be shown when the codepen is run (without selecting anything)

ko2in:
Oh I see. I tried with encoded characters (```&amp```;, ```&#x26```, ```&#38```) and still have the problem. Native select element doesn't have that issue.

lubber-de:
Fixed by #2790 
See https://jsfiddle.net/lubber/qykf58tc/2/

PR #3205, opened 2025-03-10
Fix quote and ampersand HTML escape in dropdown

### LINKED ISSUE #3199: [Dropdown] multiple issues with remote data and multiple
# Bug Report

repro: https://jsfiddle.net/5m1d6rjk/

In atk4/ui we have two components:
- Dropdown - component /wo remote API, all data are listed in `<div class="menu">`
- Lookup - component /w remote API, all data are loaded from URL

In the repro, I generated all html+js for all combinations, ie. Dropdown + Lookup, single + multiple, listed only + allow addition.

Dropdown seems to work as expected.

With Lookup, ie. Fomantic-UI dropdown without local data, I see multiple issues, but lets discuss them one by one.

## "Lookup multiple" (as labeled/named in the repro)

Here the "set text" call is honored, but the text is not rendered "as entered option":

Image

I would expect:

Image

Am I using Fomantic-UI wrongly as is it a bug?

## "Lookup multiple allow addition"

When "allowAdditions: false", then the text is displayed "as entered option", but not the text, but the value.

Image

I would expect:

Image

mvorisek:
merge after #3224

fix #3199

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -232,7 +232,7 @@
                             : [values];
                         $.each(values, function (index, value) {
                             if (module.get.item(value) === false) {
-                                html = settings.templates.addition(module.add.variables(message.addResult, value));
+                                html = settings.templates.addition(module.add.variables(message.addResult, settings.templates.escape(value, settings)));
                                 $userChoice = $('<div />')
                                     .html(html)
                                     .attr('data-' + metadata.value, value)
@@ -2787,7 +2787,7 @@
                                 .attr('data-' + metadata.text, value)
                                 .removeClass(className.filtered);
                             if (!settings.hideAdditions) {
-                                html = settings.templates.addition(module.add.variables(message.addResult, value));
+                                html = settings.templates.addition(module.add.variables(message.addResult, settings.templates.escape(value, settings)));
                                 $addition
                                     .html(html);
                             }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3787,7 +3787,7 @@
 
         maxSelections: false, // When set to a number, limits the number of selections to this count
         useLabels: true, // whether multiple select should filter currently active selections from choices
-        delimiter: ',', // when multiselect uses normal <input >, the values will be delimited with this character
+        delimiter: ',', // when multiselect uses normal <input>, the values will be delimited with this character
 
         showOnFocus: false, // show the menu on focus
         allowReselection: false, // whether current value should trigger callbacks when reselected
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1971,7 +1971,9 @@
                                     values.push({
                                         name: name,
                                         value: value,
-                                        text: module.escape.htmlEntities(text),
+                                        text: settings.preserveHTML
+                                            ? text
+                                            : module.escape.htmlEntities(text),
                                         disabled: disabled,
                                     });
                                 }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2539,7 +2539,6 @@
                         } else {
                             $input.addClass(className.noselection);
                         }
-                        let escapedValue = module.escape.value(value);
                         let hasInput = $input.length > 0;
                         let currentValue = module.get.values();
                         let stringValue = value !== undefined
@@ -2557,10 +2556,10 @@
                                 module.debug('Adding user option', value);
                                 module.add.optionValue(value);
                             }
-                            module.debug('Updating input value', escapedValue, currentValue);
+                            module.debug('Updating input value', value, currentValue);
                             internalChange = true;
                             $input
-                                .val(escapedValue);
+                                .val(value);
                             if (settings.fireOnInit === false && module.is.initialLoad()) {
                                 module.debug('Input native change event ignored on initial load');
                             } else if (preventChangeTrigger !== true) {
@@ -2568,8 +2567,8 @@
                             }
                             internalChange = false;
                         } else {
-                            module.verbose('Storing value in metadata', escapedValue, $input);
-                            if (escapedValue !== currentValue) {
+                            module.verbose('Storing value in metadata', value, $input);
+                            if (value !== currentValue) {
                                 $module.data(metadata.value, stringValue);
                             }
                         }
@@ -2694,19 +2693,18 @@
                         let $next = module.is.searchSelection()
                             ? $search
                             : $text;
-                        let escapedValue = module.escape.value(value);
                         let $label;
                         if (settings.ignoreCase) {
-                            escapedValue = escapedValue.toLowerCase();
+                            value = value.toLowerCase();
                         }
                         $label = $('<a />')
                             .addClass(className.label)
-                            .attr('data-' + metadata.value, escapedValue)
-                            .html(templates.label(escapedValue, text, settings));
-                        $label = settings.onLabelCreate.call($label, escapedValue, text);
+                            .attr('data-' + metadata.value, value)
+                            .html(templates.label(value, text, settings));
+                        $label = settings.onLabelCreate.call($label, value, text);
 
                         if (module.has.label(value)) {
-                            module.debug('User selection already exists, skipping', escapedValue);
+                            module.debug('User selection already exists, skipping', value);
 
                             return;
                         }
@@ -2745,8 +2743,7 @@
                         }
                     },
                     optionValue: function (value) {
-                        let escapedValue = module.escape.value(value);
-                        let $option = $input.find('option[value="' + CSS.escape(escapedValue) + '"]');
+                        let $option = $input.find('option[value="' + CSS.escape(value) + '"]');
                         let hasOption = $option.length > 0;
                         if (hasOption) {
                             return;
@@ -2758,7 +2755,7 @@
                             $input.find('option.' + className.addition).remove();
                         }
                         $('<option/>')
-                            .prop('value', escapedValue)
+                            .prop('value', value)
                             .addClass(className.addition)
                             .text(value)
                             .appendTo($input);
@@ -2922,8 +2919,7 @@
                         module.remove.empty();
                     },
                     optionValue: function (value) {
-                        let escapedValue = module.escape.value(value);
-                        let $option = $input.find('option[value="' + CSS.escape(escapedValue) + '"]');
+                        let $option = $input.find('option[value="' + CSS.escape(value) + '"]');
                         let hasOption = $option.length > 0;
                         if (!hasOption || !$option.hasClass(className.addition)) {
                             return;
@@ -2931,7 +2927,7 @@
                         // temporarily disconnect observer
                         module.disconnect.selectObserver();
                         $option.remove();
-                        module.verbose('Removing user addition as an <option>', escapedValue);
+                        module.verbose('Removing user addition as an <option>', value);
                         module.observe.select();
                     },
                     message: function () {
@@ -3020,9 +3016,8 @@
                         return values;
                     },
                     label: function (value, shouldAnimate) {
-                        let escapedValue = module.escape.value(value);
                         let $labels = $module.find(selector.label);
-                        let $removedLabel = $labels.filter('[data-' + metadata.value + '="' + CSS.escape(settings.ignoreCase ? escapedValue.toLowerCase() : escapedValue) + '"]');
+                        let $removedLabel = $labels.filter('[data-' + metadata.value + '="' + CSS.escape(settings.ignoreCase ? value.toLowerCase() : value) + '"]');
                         module.verbose('Removing label', $removedLabel);
                         $removedLabel.remove();
                     },
@@ -3133,13 +3128,12 @@
                         return $menu.children(selector.message).length > 0;
                     },
                     label: function (value) {
-                        let escapedValue = module.escape.value(value);
                         let $labels = $module.find(selector.label);
                         if (settings.ignoreCase) {
-                            escapedValue = escapedValue.toLowerCase();
+                            value = value.toLowerCase();
                         }
 
-                        return $labels.filter('[data-' + metadata.value + '="' + CSS.escape(escapedValue) + '"]').length > 0;
+                        return $labels.filter('[data-' + metadata.value + '="' + CSS.escape(value) + '"]').length > 0;
                     },
                     maxSelections: function () {
                         return settings.maxSelections && module.get.selectionCount() >= settings.maxSelections;
@@ -3544,26 +3538,6 @@
                 },
 
                 escape: {
-                    value: function (value) {
-                        let multipleValues = Array.isArray(value);
-                        let stringValue = typeof value === 'string';
-                        let isUnparsable = !stringValue && !multipleValues;
-                        let hasQuotes = stringValue && value.search(regExp.quote) !== -1;
-                        let values = [];
-                        if (isUnparsable || !hasQuotes) {
-                            return value;
-                        }
-                        module.debug('Encoding quote values for use in select', value);
-                        if (multipleValues) {
-                            $.each(value, function (index, value) {
-                                values.push(value.replace(regExp.quote, '&quot;'));
-                            });
-
-                            return values;
-                        }
-
-                        return value.replace(regExp.quote, '&quot;');
-                    },
                     string: function (text) {
                         text = String(text);
 
@@ -3879,7 +3853,6 @@
 
         regExp: {
             escape: /[\s#$()*+,.:=?@[\\\]^{|}-]/g,
-            quote: /"/g,
         },
 
         metadata: {
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1973,9 +1973,7 @@
                                     values.push({
                                         name: name,
                                         value: value,
-                                        text: settings.preserveHTML
-                                            ? text
-                                            : module.escape.htmlEntities(text),
+                                        text: text,
                                         disabled: disabled,
                                     });
                                 }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -795,7 +795,7 @@
                             if (values.length === 0 && !settings.allowAdditions) {
                                 module.add.message(message.noResults);
                             } else {
-                                let value = module.is.multiple() ? module.get.values() : module.get.value();
+                                let value = module.is.multiple() ? module.get.values(true) : module.get.value();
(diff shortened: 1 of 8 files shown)


src/definitions/modules/dropdown.js
```diff
@@ -3838,24 +3838,21 @@
 
                         return text.replace(regExp.escape, '\\$&');
                     },
-                    htmlEntities: function (string, forceAmpersand) {
-                        forceAmpersand = typeof forceAmpersand === 'number' ? false : forceAmpersand;
+                    htmlEntities: function (string) {
```
mvorisek: IDK why `forceAmpersand` was introduced in https://github.com/fomantic/Fomantic-UI/pull/2705. What is the usecase when ampersand "must not be escaped to make things safe"?

lubber-de: It was introduced as described in the related PR description because of " a possible XSS through an entity encoded select option text when converted into a FUI dropdown."
Here is the original jsfiddle (adjusted to run FUI 2.9.2 which had the issue back then) to be able to reproduce the issue.

Select the obvious evil dropdown entry to see the issue

https://jsfiddle.net/lubber/norj3k7p/1/

lubber-de: We probably need to completely rewrite the whole data handling of dropdown in all possible situations (seek through all the dropdown issues the past years to find related jsfiddles...) to clean this up. 
Most difficult part always were
-  comparing/check for existing values, epecially on "mutlitple"
- already entity encoded values, avoid "double encoding" (which was one reason not to encode the ampersand again, otherwise the result of `&amp;` was `&amp;amp;` for data which was meant to be stored, not rendered
- select converting into html dropdowns

src/definitions/modules/dropdown.js
```diff
@@ -3811,51 +3805,26 @@
...
-                                '"': '&quot;',
-                                "'": '&#x27;',
-                                '`': '&#x60;',
                             },
                             escapedChar  = function (chr) {
                                 return escape[chr];
                             }
                         ;
-                        if (shouldEscape.test(string)) {
-                            string = string.replace(forceAmpersand ? /&/g : /&(?![\d#a-z]{1,12};)/gi, '&amp;');
```
lubber-de: Just saying this will encode already encoded values again

```javascript
&quot; becomes &amp;quot;
&amp; becomes &amp;amp;
...
```
The internal logic needs to ensure that this does not lead into other issues (comparing existing data)

src/definitions/modules/dropdown.js
mvorisek: > XSS issue

I tested it using my PR and XSS is not present.

>  need to completely rewrite the whole data handling of dropdown

I completely agree.

> "double encoding" 

Yes, this is the problem. Not only `&amp;`, but also `&quote;` stripping. This is what this PR addresses and I described it shortly in the PR title/description.

src/definitions/modules/dropdown.js
mvorisek: @lubber-de please review and test this PR:

test string with preserveHTML = false:
```
$makeTestStringFx = static fn ($v) => $v . ' <b>"\' &lt;';
$htmlValues = [
    $makeTestStringFx('d') => $makeTestStringFx('dTitle'),
    $makeTestStringFx('u') => $makeTestStringFx('uTitle'),
];
```

before:
image
image

after:
image
image

There is currently only 1 unrelated issue I am aware with "Lookup multiple allow addition":
image

this happen when the "Lookup multiple allow addition" dropdown is clicked because of https://github.com/fomantic/Fomantic-UI/blob/2.9.4/src/definitions/modules/dropdown.js#L863 as it "the getters" currently return escaped value although it is only internal when preserveHTML = false, this is described in the end of https://github.com/fomantic/Fomantic-UI/issues/3199#issuecomment-2710662588. I do not want to fix all getters in this PR, I will update #3199 with the problem once this PR is landed.

mvorisek:
Test strings with preserveHTML = false:
```
$makeTestStringFx = static fn ($v) => $v . ' <b>"\' &lt;';
$htmlValues = [
    $makeTestStringFx('d') => $makeTestStringFx('dTitle'),
    $makeTestStringFx('u') => $makeTestStringFx('uTitle'),
];
```

before:
image
image

after:
image
image

mvorisek:
For unknown reasons to me, multiple values are escaped by default by `get values` API in https://github.com/fomantic/Fomantic-UI/blob/2.9.4/src/definitions/modules/dropdown.js#L2012. This is addressed in https://github.com/fomantic/Fomantic-UI/pull/3205/commits/de27144e217d6e2a44e14da81259de598e95e978 to fix https://github.com/fomantic/Fomantic-UI/issues/3199#issuecomment-2710662588.

I tested both (t/f) `preserveHTML` and this PR fixes all #3199 issues.

lubber-de (changes requested):
I took the time and went through (possibly) all related old issues and PRs the past years.
Most of them work fine with this PR. :+1:

However, i found 3 of them which are now broken when this PR (de27144e217d6e2a44e14da81259de598e95e978) is applied:

## 1.  Original Issue: #1373, Original PR: #1770

### FUI 2.9.4
https://jsfiddle.net/lubber/23htdnw0/1/
--> works

### Your PR applied
https://jsfiddle.net/lubber/gmh5reoy/5/
--> fails (even though preserveHTML is true, the labels are encoded)
--> fails (when preserveHTML is false, ampersand is double encoded)



## 2. Original Issue: #2782, Original PR: #2790

Mentioned example c) 

### FUI 2.9.4
https://jsfiddle.net/lubber/q8hL21yg/1/
--> works

### Your PR applied
https://jsfiddle.net/lubber/zap03L4k/5/
--> fails (Ampersand always encoded even when preserveHTML is true. 
--> fails (When preserveHTML is false, the ampersand is double encoded)




## 3. Original Issue: #2817, Original PR: #2705

### FUI 2.9.4
https://jsfiddle.net/lubber/97druzp4/1/
--> works

### Your PR applied
https://jsfiddle.net/lubber/rhs4aq7u/3/
--> fails (always encoded)

mvorisek:
Thank you for the extensive review.

I extracted changes I consider safe into #3212 to minimize the LoC changed of this PR to discuss it later.

src/definitions/modules/dropdown.js
```diff
@@ -710,7 +710,9 @@
                             module.remove.message();
                         }
                         if (settings.allowAdditions) {
-                            module.add.userSuggestion(module.escape.htmlEntities(query));
+                            module.add.userSuggestion(settings.preserveHTML
+                                ? module.escape.htmlEntities(query)
+                                : query);
```
lubber-de: This allows for XSS when preserveHTML is false as the query string is completely unsanitized.
Try https://jsfiddle.net/lubber/dt43gsqp/8/ (your PR is applied)

click into the dropdown and paste "<script>alert('gotcha!')</script>"
-> Alert happens
xss


Original issue was #1207

lubber-de (changes requested):
I found an issue. on useradditions and preserveHTML: false the query string stay unsanitized. Before the change the query string was always html encoded
Shouldn't preserveHTML: false mean to remove/encode the query string rather than _not_ encoding it?

src/definitions/modules/dropdown.js
mvorisek: Is https://github.com/fomantic/Fomantic-UI/pull/3205#pullrequestreview-2713368360 (review comment) the same issue as https://github.com/fomantic/Fomantic-UI/pull/3205#discussion_r2011888271 (comment above)?

> https://jsfiddle.net/lubber/dt43gsqp/8/

Where is the problem, what do you expect? The XSS is not in Fomantic-UI dropdown, but instead of in the JS code in the jsfiddle - try to remove the `onChange` (or replace `.html(` with `.text(` there).

https://github.com/fomantic/Fomantic-UI/pull/3205/commits/a995f2fce5067ca029f07031ec5ddfe179496088 the point is when `preserveHTML = false`, all APIs should work on text (ie. what is rendered/what user see), not HTML.

lubber-de: Yes, the same review comment.
 I woud expect to have the query being encoded/sanititzed when preserveHTML is false. 
Instead, it is encoded when preserveHTML is true.
FUI 2.9.4 does always encode the query string regardless of the preserveHTML setting, that's why XSS was never happening in that use case before

mvorisek: With `preserveHTML=false` you tell Fomantic-UI "I want to work with text, instead of HTML". Text is already accepted, text is already stored/sent (in value).

In the jsfiddle, `val` argument in `onChange` was passed previously as HTML. Why would you expect HTML here, when using `preserveHTML=false`, especially when the stored/sent value would be text (ie. not HTML encoded) instead?

lubber-de: preserveHTML: true means: "Keep the given string as it is, dont encode, dont remove anything, even if there is HTML in my string"
preserveHTML: false means "make sure there wont be any HTML stored/displayed/used. Instead encode it or remove HTML" (thats why we set it to false by default to make it secure by default.

Look at the "get.text"or the "templates.escape" function

get.text:
```javascript
// when preserveHTML is true, assume there is HTML inside my string and render it (via .html()
// when preserveHTML is false, any HTML will be removed via .text()
text: function () {
                        return settings.preserveHTML
                            ? $text.html()
                            : $text.text();
                    },
```

templates.escape
```javascript
// when preserveHTML is true it just returns the given string as it is
// when preserveHTML is false the string will be encoded
if (settings !== undefined && settings.preserveHTML) {
                return string;
            }
```

lubber-de: > In the jsfiddle, `val` argument in `onChange` was passed previously as HTML. Why would you expect HTML here, when using `preserveHTML=false`, especially when the stored/sent value would be text (ie. not HTML encoded) instead?

FUI does not know what a user could possibly supply (as custom addition in this case), so preserverHTML: false should make sure any supplied string does not contain HTML and if, it should be encoded / removed

lubber-de: I still believe, the add.usersuggestion function should always encode the given query, regardless of preserveHTML setting as it was before.

lubber-de:
Here is a jsfiddle where XSS takes place inside FUI when preserveHTML is false and the dropdown is NOT multiple:
https://jsfiddle.net/8rabcw71/1/

Paste "<script>alert('gotcha!')</script>" into the second dropdown and it gets executed immedialty.
This way an attacker could add someting less obvious like `<b>hi there</b>`
Also the dropdown kinda gets stuck afterwards....

src/definitions/modules/dropdown.js
mvorisek: As values and other API accept/return raw values, this would be inconsistent then.

I introduced https://github.com/fomantic/Fomantic-UI/pull/3205/commits/a995f2fce5067ca029f07031ec5ddfe179496088 to mainly to prevent double encoding. If you think this is wrong, feel free to commit directly to this/#3224. I am quite sure you will find out this is how the data should move and the code be.

lubber-de: > As values and other API accept/return raw values, this would be inconsistent then.

These are the original SUI issues related to that use case
https://github.com/Semantic-Org/Semantic-UI/issues/4498
https://github.com/Semantic-Org/Semantic-UI/issues/6571
https://github.com/Semantic-Org/Semantic-UI/issues/3318

which are introduced again, if we distinguish for preserveHTML at "add.usersuggestion" (while once fixed by #298)

Anyway: preserveHTML:true should not encode (while storing the data), but preserveHTML: false always should (as only text is expected/accepted), not vice versa as it is since a995f2fce5067ca029f07031ec5ddfe179496088 .
Every other module which uses preserveHTML work this way.

This would make the new default for preserveHTML kept secure as intended.

While i would agree to not mutate the value itself, we should at least prevent execution when rendered inside the dropdown (perhaps this is only needed on non-multiple dropdowns now. 
So each item having the class "addition" (added in "module.create.userChoice()" might always be encoded when the to be rendered domnode is created?)

mvorisek:
> Here is a jsfiddle where XSS takes place inside FUI when preserveHTML is false and the dropdown is NOT multiple: https://jsfiddle.net/8rabcw71/1/
> 
> Paste "<script>alert('gotcha!')</script>" into the second dropdown and it gets executed immedialty. This way an attacker could add someting less obvious like `<b>hi there</b>` Also the dropdown kinda gets stuck afterwards....

Good catch, fixed.

src/definitions/modules/dropdown.js
mvorisek: https://github.com/fomantic/Fomantic-UI/pull/3205#issuecomment-2751965864 was a legit XSS/bug and I fixed. I am not aware of any other issue.

The fix commit is https://github.com/fomantic/Fomantic-UI/pull/3205/commits/fddfe7aa29e4c73496ad552a1d2764b5f8f72564. See l238. If we would escape the input value earlier (with https://github.com/fomantic/Fomantic-UI/pull/3205/commits/12eda65b3df6e6f42fb1a9e0819f8c88f1e824ba reverted), the value in the attribute would be set wrongly otherwise - I really belive we need to work with the raw value until we are building renderable HTML.

lubber-de:
Two things left for me atm:
## 1st
The fix from https://github.com/fomantic/Fomantic-UI/pull/3205/commits/fddfe7aa29e4c73496ad552a1d2764b5f8f72564 should also be applied to https://github.com/fomantic/Fomantic-UI/blob/fddfe7aa29e4c73496ad552a1d2764b5f8f72564/src/definitions/modules/dropdown.js#L2790

so it also becomes

```javascript
html = settings.templates.addition(module.add.variables(message.addResult, settings.templates.escape(value, settings)));
```

## 2nd
As i explained the meaning of preserveHTML in https://github.com/fomantic/Fomantic-UI/pull/3205#discussion_r2012217440
The check at https://github.com/fomantic/Fomantic-UI/blob/fddfe7aa29e4c73496ad552a1d2764b5f8f72564/src/definitions/modules/dropdown.js#L713
should be vice versa.

It should be either

```javascript
module.add.userSuggestion(!settings.preserveHTML
	? module.escape.htmlEntities(query)
	: query);
```

or

```javascript
module.add.userSuggestion(settings.preserveHTML
	? query
	: module.escape.htmlEntities(query));
```

But the related change was implemented in #3224, so it might be adjusted there, if you want the other PR to be merged first

mvorisek:
> 1st

Fixed by me, my mistake. I now checked all `.html(` inputs.

> 2nd

You are worried about https://github.com/fomantic/Fomantic-UI/commit/a995f2fce5067ca029f07031ec5ddfe179496088. No, it makes sense - with `preserveHTML = true` we want to preserve HTML, but not from user input :), thus we encode the input first and then preserve it everywhere else as other values. With `preserveHTML = false` the input is text already, no need to encode it, we encode it only for HTML render.

lubber-de:
> No, it makes sense - with `preserveHTML = true` we want to preserve HTML, but not from user input :),

Sure thing 👍🏼 That's why it was basically always encoded before just in case. 

> thus we encode the input first and then preserve it everywhere else as other values. With `preserveHTML = false` the input is text already, no need to encode it, we encode it only for HTML render.

Alright. I debugged it down now more deeply and i can confirm, after your latest fixes, that the usersuggestion is now safely encoded when finally being added to the DOM 🥳 

Just to clarify/summarize (for myself, if i ever need to rethink/reconfirm this in some later situation/refactoring)
1. The query variable, when preserveHTML is false, is still unsanitized when provided to the `module.filter()` function because it just gets the .val() from the input, no matter what there might be HTML inside
2. It is even still unsanitized, when preserveHTML is false, when provided to the `module.add.userSuggestion` function

Those 2 situations, in addition to the found XSS use case yesterday, made me not believing the code was correct... until today 😉 

3. **BUT**: inside `module.add.userSuggestion` the value GETS encoded when preserveHTML is false: The two LOC where `settings.templates.addition()` is called (that was fixed in the latest commits)

### @mvorisek Thanks for being patient with me (again) and explaining all my confusion

I will now look into the other modules usages for the escape method which were also adjusted in this PR to finally merge your both PRs

src/definitions/modules/modal.js
```diff
@@ -874,7 +874,7 @@
                             return string;
                         }
 
-                        const badChars = /["'<>]|&(?![\d#A-Za-z]{1,12};)/g;
+                        const badChars = /["&'<>]/g;
```
lubber-de: I would like to question/discuss this change again (outside of dropdown) ;)

In all modules, beside dropdown, the escape function is always used inside output rendering functions, the escaped value itself is never stored/mutated internally for later usages again.

When forcing the ampersand to always being encoded it is not possible to supply preencoded htmlentities by the user anymore, so something like 
`title: ' The sun is &gt; earth and "worth" a thousand &dollar;'` wont work properly
because this will be encoded to  `The sun is &amp;gt; earth and &quot;worth&quot; a thousand &amp;dollar;` (without this change it will output the desired  `The sun is &gt; earth and &quot;worth&quot; a thousand &dollar;`

The current regex supports already htmlentity encoded values and only encodes the ampersand if it does not follow entity encoded spec characters and a semicolon.

There are a [ton of possible htmlentity encodes characters](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references) which are possibly provided and wont work anymore. The user is now forced to provide unencoded string

If people would use preserveHTML: true, there isnt any encoding at all. In the above example this would keep the encodings, but breaks the direct quotes.

We can of course question if this is a valid use case (providing mixed(encoded and not encoded characters) to the modules settings).

So, if we keep this change, we force the user to provide *unescaped* strings __or__ use preserveHTML: true and provide already encoded strings.

@mvorisek Your thoughts?

mvorisek: Every encodings is for a decoder. The decoder can be also HTML render/user. And in ideal case, the result should be the same as the original input.

Consider `<b>&amp;` text from database with `preserveHTML=false`, I expect the text to be really rendered as stored, not with exceptions. This is almost always path to problems as we both noticed hardly with dropdown with the current state and past patches.

I am inclined to encode everything, but then we have the `>` -> `&amp;` behaviour of browser...

As the dropdown changes are quite complex, let's focus solely on dropdown in these 2 PRs and then let's discuss the remaining modules. I reverted the changes on other modules.

mvorisek: I think there is a general solution.

Why do we want to allow `&amp;`, `&gt;` unencoded (not double encoded)? Because `&` and `<` cannot be written in innerHTML.

It can be written in HTML attribute and JS/FUI API safely however.

Thus the solution is to encode everything, but when reading the "initial text from innerHTML", decode the text before. This is a logic that can be clearly explaned to developers and best be can squeze of the HTML parser behaviour.

This is a comment only, let's not bother with the impl. in these PRs.

lubber-de (approved):
LGTM, Thanks for your patience :)

PR #3212, opened 2025-03-13
Never "dequote" attributes, always escape

mvorisek:
extracted from #3205

We use here a trick that HTML attributes can be escaped the same way as HTML values - https://jsfiddle.net/45q9zL7j/, thus HTML attributes does not need a separate escape function.

Under normal usage this PR should imply no function change.


lubber-de (approved):
LGTM 👍🏼

PR #3213, opened 2025-03-13
Improve CS: declare variables individually

mvorisek:
This is default and prefered style beause:
- the indentation is minimal
- minimal diff on variable remove/add
- less total LoC
- `var` is used historically, my plan is to land this PR, and then detect not-redeclared variables and declare then using `let`/`const`


lubber-de:
According to https://eslint.org/docs/latest/rules/one-var the default for "one-var" is 'always' (you removed the one-var declaration in eslintrc) and "always" means to _not_ declare each variable by a separate var/let/const ....
did i misunderstood the docs or did i miss something in your PR?

image
image

mvorisek:
You are right, I corrected the 1st commit title to mention "never". AirBnB/Unicorn default is newly used and the reasons why explained in the PR description.

mvorisek:
PR is done. I fixes the semicolon placements using regex manually, as `semi-style` rule removes wrongly new lines after them when autofixing.

lubber-de:
@mvorisek  Please fix merge conflicts, as i just merged #3212 :wink:

mvorisek:
Rebased.

lubber-de:
Again, please rebase / resolve conflicts, as i just merged your other PR #3215 🙂

mvorisek:
PR redone/rebased.

scripts/nightly-version.js
```diff
@@ -20,16 +20,14 @@ const currentRev = childProcess // get the current rev from the repo
     .execSync('git rev-parse HEAD')
     .toString()
     .trim()
-    .slice(0, 7)
-;
+    .slice(0, 7);
```
lubber-de, line 23: @mvorisek This rule change is inconsistent (at several places)
You changed the one-var rule so each line could be easier deleted/added without touching another line to change the end "," to ";" or vice versa.

The same argument could be applied here too: Forcing the semicolon to the last method call here would need a change to that line if i ever want to add another method to the string chain.

So, when the argument for the one-var rule is "i can change the single line without touching others", i think this should be applied here too, so the last semicolon should be kept in a separate line as before (So i would be able to delete/add it without touching any other line)

I won't continue reviewing until this is adjusted/discussed

mvorisek: This is true and I am aware of it.

The only negative is extra line in diff when the last chained call is removed/added.

IMO long-term-stable CS outperforms this negative.

Given this negative is only one of many reasons of this PR and given there is no `semi-style` rule allowing us placing semicolon on a separate line, we should do a compromise and use `semi-style: last`.

This config is also default in default esling config, AirBnB and Unicorn stylesets, so let's land it and have the CS normalized and asserted by CI.

tasks/config/admin/templates/less-package.js
```diff
@@ -1,6 +1,5 @@
 let
-    where = 'client' // Adds files only to the client
-;
+    where = 'client'; // Adds files only to the client
```
lubber-de: This should be merged with the previous line, so it's only

```javascript
let where = 'client';
```

In css-package.js, this was already done

mvorisek: fixed

mvorisek:
Let's please merge this PR before anything else to prevent conflicts.

I did a verification by replacing `\s+` with `\n` after the 1st autofixed commit and after all commits. All non-whitespace changes are wanted, thus there is no functional change.

lubber-de (approved):
LGTM

PR #3215, opened 2025-03-14
Refactor HTML escape functions

mvorisek:
extracted from #3205

The new code is simpler, faster and easy to fix ampersand escape later.

No functional change.


lubber-de (approved):
LGTM 👍🏼

PR #3216, opened 2025-03-14
feat(dropdown): maintain a lower opacity of the clear icon unless being hovered

### LINKED ISSUE #3211: [Dropdown] Unfocused - focused - hovered actionable icons in input fields
# Feature Request

Currently, when an input/dropdown is focused, all icons are displayed with 100% opacity.

Image

I find this not nice especially when the dropdown is clearable, the clearable icon is visually the strongest, although it is used rarely.

This is a feature request to make the icons opacity tri-state. Display the icons with reduced opacity even if the input/dropdown field is focused and with opacity=1 only when the actionable icon itself is hovered.

This will improve UX:
- not important controls not displayed too strongly
- clickable controls will imply clickability by interacting with the cursor when hovered

here is repro for a general input - https://dev.atk4.org/demos/form-control/input2.php (the eye icon at password fields)

ko2in:
## Description
When the dropdown allows for clearing, the clear icon should maintain a lower opacity unless it's being hovered. This improves user experience by indicating the option to clear the selection only when the user interacts with it.

## Testcase
**Current Version:** https://fomantic-ui.com/modules/dropdown.html#clearable-selection

Click the dropdown menu and check the clear icon. It displays with full opacity.

**New Version:** https://jsfiddle.net/ko2in/6gerqn2c/

Click the dropdown menu and check the clear icon. It will remain visible with low opacity until you hover the mouse over the icon.

## Screenshot
**Current Version:**
Dropdown-Old-Feature

**New Version:**
Dropdown-New-Feature

## Closes
#3211

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -616,13 +616,17 @@ select.ui.dropdown {
         .ui.ui.selection.dropdown@{notInverted}:has(> input:focus) {
             border-color: @selectionFocusBorderColor;
             box-shadow: none;
-            & > i.icon {
+            & > i.icon:not(.remove) {
                 opacity: @selectionIconFocusOpacity;
             }
         }
     }
-    .ui.ui.selection.dropdown:focus > i.icon {
-        opacity: @selectionIconFocusOpacity;
+    .ui.ui.selection.dropdown {
+        &:focus, &:has(> input:focus) {
+            & > i.icon:not(.remove) {
+                opacity: @selectionIconFocusOpacity;
+            }
+        }
     }
 
     /* Visible */
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -622,7 +622,8 @@ select.ui.dropdown {
         }
 
         .ui.ui.selection.dropdown {
-            &:focus, &:has(> input:focus) {
+            &:focus,
+            &:has(> input:focus) {
                 & > i.icon:not(.remove) {
                     opacity: @selectionIconFocusOpacity;
                 }
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -620,11 +620,12 @@ select.ui.dropdown {
                 opacity: @selectionIconFocusOpacity;
             }
         }
-    }
-    .ui.ui.selection.dropdown {
-        &:focus, &:has(> input:focus) {
-            & > i.icon:not(.remove) {
-                opacity: @selectionIconFocusOpacity;
+
+        .ui.ui.selection.dropdown {
+            &:focus, &:has(> input:focus) {
+                & > i.icon:not(.remove) {
+                    opacity: @selectionIconFocusOpacity;
+                }
             }
         }
     }
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -628,7 +628,7 @@ select.ui.dropdown {
                     opacity: @selectionIconFocusOpacity;
                 }
                 & > i.icon.remove:not(:hover) {
-                    opacity: calc(@actionIconFocusOpacity);
+                    opacity: @actionIconFocusOpacity;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/definitions/modules/dropdown.less
```diff
@@ -616,13 +616,17 @@ select.ui.dropdown {
...
-            & > i.icon {
+            & > i.icon:not(.remove) {
                 opacity: @selectionIconFocusOpacity;
             }
         }
     }
-    .ui.ui.selection.dropdown:focus > i.icon {
-        opacity: @selectionIconFocusOpacity;
+    .ui.ui.selection.dropdown {
+        &:focus, &:has(> input:focus) {
```
lubber-de: The `:has` selector is [too new](https://caniuse.com/?search=%3Ahas) (unless we decide to drop every browser older than May 2022 (iOS 15.4) or even December 2023(!) when Firefox activated this by default).... (Even the decision to drop older than 2021 wasn't clear #2854)
So the selector should be wrapped as we always do whenever we need to use :has to make things work:

```less
@supports selector(:has(.f)) {
....
}
```

The previous selector is already wrapped into that, so you might just move your additional selector into the support block as well

ko2in: @lubber-de Updated. Thanks.

mvorisek:
#3211 is about clickable icons in general. It would be great if there can be some "middle opacity" (opacity after field is focused, but icon not hovered), as there can be more clickable icons which UX will be inconsistent with the "remove icon" after this PR.

ko2in:
@mvorisek I don't see there will be more clickable icons except the current arrow and clear icons as this component is not typically designed to include multiple actionable elements in accordance with its intended user experience.

Even so, it's not possible to know what icons should have lower opacity than other icon unless we explicitly specify. So, we still need to extend the current CSS in this case.

The default opacity for the icon in it's non-interacting state is **0.5**. Any adjustment between **0.5** and **1** for such a tiny icon, result in a change that is likely to be barely noticeable. So, I don't see any significant improvement here.

mvorisek:
Here https://jsfiddle.net/sq5b1ewL/ is a demo, I am neutral on it as it look slighly more usable on my monitor, but not sure how it renders on other displays.

ko2in:
I've added an additional state for remove icon with the opacity level between inactive and active state, which is 68% of active state as in @mvorisek example.

I chose to use the variable name `@actionIconFocusOpacity` instead of `@removeIconFocusOpacity` to prevent any potential confusion, as it may represent something other than removal in the context of user customization.

src/definitions/modules/dropdown.less
```diff
@@ -616,13 +616,22 @@ select.ui.dropdown {
...
-        opacity: @selectionIconFocusOpacity;
+
+        .ui.ui.selection.dropdown {
+            &:focus,
+            &:has(> input:focus) {
+                & > i.icon:not(.remove) {
+                    opacity: @selectionIconFocusOpacity;
+                }
+                & > i.icon.remove:not(:hover) {
+                    opacity: calc(@actionIconFocusOpacity);
```
lubber-de: The calc method should be either moved to the variable declaration itself or escaped here if your intention is to generate "calc()" as real CSS property.
calc() itself is a LESS function, so in some situations the value is calculated by LESS itself. Don't know your intention, but to allow the user to define anything as opacity we should move the calc() inside the variable.

Here is a short overview about the differences
image

ko2in: Already fixed.

ko2in: Sorry. I just remove `calc()` as I think it's unnecessary. We can safely use the evaluated value from LESS.

ko2in:
@lubber-de I've rebased and updated again. Can you please merge it before it's too outdated from base branch?

src/definitions/modules/dropdown.less
```diff
@@ -616,13 +616,22 @@ select.ui.dropdown {
         .ui.ui.selection.dropdown@{notInverted}:has(> input:focus) {
             border-color: @selectionFocusBorderColor;
             box-shadow: none;
-            & > i.icon {
+            & > i.icon:not(.remove) {
                 opacity: @selectionIconFocusOpacity;
             }
```
lubber-de: This block can now be removed as it is doubled by your new selector block below. And the check for notInverted doesnt seem to be relevant (for the icons) (your new selector overrides it anyway and the jsfiddle works fine)

ko2in: Thanks. Fixed.

lubber-de (approved):
LGTM

PR #3217, opened 2025-03-14
fix(dropdown): API query does not run when the dropdown is opened by pressing the down arrow key

### LINKED ISSUE #3204: [Dropdown] values set in constructor should not prevent implicit remove data fetch
# Bug Report

## Steps to reproduce

1. open https://jsfiddle.net/5cw84dmn/
2. click the page somewhere (needed for the next step)
3. press "tab" key (to focus the dropdown input)
4. press "down" key (to show the dropdown menu)
5. notice only "iniTitle" menu item is displayed

Now click somewhere else and open the dropdown using click, and notice, the menu is implicitly refreshed when opened differently.

## Expected result

Dropdown implicitly refreshes before the dropdown is shown no matter how the dropdown is opened (via click, "down" key, via `.dropdown('show')` API...).

## Actual result

Dropdown does not always refresh.

Dropdown does not refresh even on the 1st load when the `values` are set in the constructor (as in the repro). When `values` are not set in the constructor, dropdown refreshes at the 1st load. This is because of https://github.com/fomantic/Fomantic-UI/blob/2.9.4/src/definitions/modules/dropdown.js#L538. I would expect the save dropdown menu the opened with "down" key as when the input is clicked. When clicked, the dropdown does refresh.

ko2in:
## Description
When the dropdown uses API settings, it retrieves data from a remote source only if there are no items present in the dropdown.

If we press the down arrow key to open the dropdown, it doesn't execute the query method to fetch resources from remote API, if the dropdown is initialized with default values.

Naturally, the dropdown clears any existing items if it is opened by clicking the icon or focusing on the search field. But, it doesn't happen when pressing the down key. So, the dropdown remains to display the initial values without refreshing and doesn't receive the new data.

This issue is resolved by specifying the dropdown as focused when it's opened by pressing the down arrow key.

## Testcase
**Bug:** https://jsfiddle.net/5cw84dmn/

**Fix:** https://jsfiddle.net/lubber/qcnb5jg8/

## Closes
#3204

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1757,6 +1757,7 @@
                                 }
                                 // down arrow (open menu)
                                 if (pressedKey === keys.downArrow && !module.is.visible()) {
+                                    focused = true;
                                     module.verbose('Down key pressed, showing dropdown');
                                     module.show();
                                     event.preventDefault();


lubber-de (approved):
LGTM (i adjusted your PR description with a working jsfiddle for the fixed version)

mvorisek:
Thank you, works great!

PR #3218, opened 2025-03-15
Use "CSS.escape()" for building dynamic CSS selectors

mvorisek:
extracted from #3205

Values in CSS selectors should never be dequoted (this is lossy operation), but instead of escaped using native CSS.escape(). This is described in https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static#in_context_uses.

Under normal usage this PR should imply no function change.

### DIFF
--- a/src/definitions/modules/tab.js
+++ b/src/definitions/modules/tab.js
@@ -254,14 +254,6 @@
                     },
                 },
 
-                escape: {
-                    string: function (text) {
-                        text = String(text);
-
-                        return text.replace(regExp.escape, '\\$&');
-                    },
-                },
-
                 set: {
                     auto: function () {
                         let url = typeof settings.path === 'string'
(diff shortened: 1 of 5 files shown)


lubber-de:
Please resolve conflicts, as i just merged #2584

mvorisek:
rebased

src/definitions/modules/tab.js
```diff
@@ -364,8 +356,7 @@
                             settings.onLoad.call($tab[0], currentPath, parameterArray, historyEvent);
                         } else if (tabPath.search('/') === -1 && tabPath !== '') {
                             // look for in page anchor
-                            tabPath = module.escape.string(tabPath);
-                            $anchor = $('#' + tabPath + ', a[name="' + tabPath + '"]');
+                            $anchor = $('#' + tabPath + ', a[name="' + CSS.escape(tabPath) + '"]');
```
lubber-de: The first part of the selector is not escaped
```suggestion
                            $anchor = $('#' + CSS.escape(tabPath) + ', a[name="' + CSS.escape(tabPath) + '"]');
```

mvorisek: I fixed it and checked all other `#' \+` cases.

lubber-de (approved):
LGTM

PR #3219, opened 2025-03-16
Fix dropdown "get values" to always return raw value

mvorisek:
Unify API, unencoded values are returned/accepted by other APIs.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1801,7 +1801,7 @@
                         return Math.ceil($sizer.width() + 1);
                     },
                     selectionCount: function () {
-                        let values = module.get.values();
+                        let values = module.get.values(true);
                         let count;
                         count = module.is.multiple()
                             ? (Array.isArray(values) ? values.length : 0)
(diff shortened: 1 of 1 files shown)


mvorisek:
The https://github.com/fomantic/Fomantic-UI/pull/3219/commits/d80f0ba2377c40d9152ab9c938feb46a6e0b995e commit will probably break something as analysed in https://github.com/fomantic/Fomantic-UI/pull/3205#pullrequestreview-2679705613.

However, I belive "get values" API should return always the raw result no matter the `!module.has.selectInput() && module.is.multiple()` condition.

Help with adjusting the code where the result is used welcomed. I have not enough knowledge what/when the values should be escaped nor I am aware of old bugs/fixes.

lubber-de:
Your PR basically reverts #1770
However, as you refactored/improved/simplified the whole encoding in #3205 the original PR changes really dont seem to be necessary anymore.

I quickly tested your latest commit with the original testcase and it seems to be fine 👍🏼 
https://jsfiddle.net/lubber/2x31e5o8/1/

mvorisek:
Fixes were landed in #3233, #3234, #3235. PR is rebased and unchanged. Please review. There is no function change except the external API result fixed.

lubber-de (approved):
LGTM

PR #3222, opened 2025-03-17
Fix non-string type HTML escape

mvorisek:
fix #3215


mvorisek:
issue example:

image

lubber-de (approved):
LGTM

PR #3223, opened 2025-03-17
HTML escape should accept only string and null

mvorisek:
#3222 fixes missing `number.replace` function well, but `null` type is converted to `"null"`. `""` is mostly wanted to be displayed instead.

`undefined` is converted to `"undefined"`. But most `undefined` cases are handled explicitly and values passed using `setting.someValue` should always be defined. Thus do not have to handle this case in general.

Casting `object` or `array` types using `String()` is bad and should never be done.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3544,6 +3544,9 @@
                         return text.replace(regExp.escape, '\\$&');
                     },
                     htmlEntities: function (string) {
+                        if (string === null) {
+                            string = '';
+                        }
                         console.assert(typeof string === 'string');
 
                         const escapeMap = {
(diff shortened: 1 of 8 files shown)


mvorisek:
This PR is draft and I want to finish it after #3205.

I belive the https://github.com/fomantic/Fomantic-UI/pull/3222#issuecomment-2727766628 can be fixed with explicitly on the caller side on a few lines where `number` is passed. After reviewing another dropdown code, it seems this case is handled on other places explicitly already.

@lubber-de what do you think?

mvorisek:
This PR addresses  https://github.com/fomantic/Fomantic-UI/pull/3222#issuecomment-2727766628.

Refs:
- https://stackoverflow.com/questions/12594683/how-to-force-a-string-value-in-html-5-data-attributes
- https://github.com/fomantic/Fomantic-UI/blob/2.9.4/src/definitions/modules/dropdown.js#L2905

This behaviour is sadly caused by https://github.com/jquery/jquery/blob/3.7.1/src/data.js#L24-L47. This is impossible to be fully fixed in Fomantic-UI JS without patching jQuery on the page load (prior any jQuery operations).

See related https://github.com/atk4/ui/pull/2274 / https://github.com/atk4/ui/commit/45d47a9bebebb2f610ca3121c632cfefb65022e8.

PR #3224, opened 2025-03-18
Always HTML escape ampersand in dropdown

mvorisek:
extracted from #3205

And fix many double encoding issues.

All escape functions must be lossless and must return the same value when called again on a reverse function.

When using preserveHTML=false and the text is read from innerHTML input (`<option>xxx</option>`), `&amp;` must be encoded in HTML code as `&amp;amp;` to be preserved 1:1, This is due `&` -> `&amp;` browser HTML normalization. Explained in [1], [2], [3].

[1] original issue: https://github.com/fomantic/Fomantic-UI/issues/2782
[2] browser behaviour: https://jsfiddle.net/wdyjfvz0/
[3] how to encode innerHTML for `<option>` tag: https://3v4l.org/E2BQY

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -232,7 +232,7 @@
                             : [values];
                         $.each(values, function (index, value) {
                             if (module.get.item(value) === false) {
-                                html = settings.templates.addition(module.add.variables(message.addResult, value));
+                                html = settings.templates.addition(module.add.variables(message.addResult, settings.templates.escape(value, settings)));
                                 $userChoice = $('<div />')
                                     .html(html)
                                     .attr('data-' + metadata.value, value)
@@ -2787,7 +2787,7 @@
                                 .attr('data-' + metadata.text, value)
                                 .removeClass(className.filtered);
                             if (!settings.hideAdditions) {
-                                html = settings.templates.addition(module.add.variables(message.addResult, value));
+                                html = settings.templates.addition(module.add.variables(message.addResult, settings.templates.escape(value, settings)));
                                 $addition
                                     .html(html);
                             }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3787,7 +3787,7 @@
 
         maxSelections: false, // When set to a number, limits the number of selections to this count
         useLabels: true, // whether multiple select should filter currently active selections from choices
-        delimiter: ',', // when multiselect uses normal <input >, the values will be delimited with this character
+        delimiter: ',', // when multiselect uses normal <input>, the values will be delimited with this character
 
         showOnFocus: false, // show the menu on focus
         allowReselection: false, // whether current value should trigger callbacks when reselected
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1971,7 +1971,9 @@
                                     values.push({
                                         name: name,
                                         value: value,
-                                        text: module.escape.htmlEntities(text),
+                                        text: settings.preserveHTML
+                                            ? text
+                                            : module.escape.htmlEntities(text),
                                         disabled: disabled,
                                     });
                                 }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1973,9 +1973,7 @@
                                     values.push({
                                         name: name,
                                         value: value,
-                                        text: settings.preserveHTML
-                                            ? text
-                                            : module.escape.htmlEntities(text),
+                                        text: text,
                                         disabled: disabled,
                                     });
                                 }
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2052,7 +2052,7 @@
                                         return;
                                     }
                                     if (isMultiple) {
-                                        if ($.inArray(module.escape.htmlEntities(String(optionValue)), value.map(String).map(module.escape.htmlEntities)) !== -1) {
+                                        if ($.inArray(module.escape.htmlEntities(String(optionValue)), value.map(String)) !== -1) {
                                             $selectedItem = $selectedItem
                                                 ? $selectedItem.add($choice)
                                                 : $choice;
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1950,7 +1950,7 @@
                             .find('option')
                             .each(function () {
                                 let $option = $(this);
-                                let name = $option.html();
+                                let name = module.escape.assumeUnescapedAmpersand($option.html());
                                 let disabled = $option.attr('disabled');
                                 let value = $option.attr('value') !== undefined
                                     ? $option.attr('value')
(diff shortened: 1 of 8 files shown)


mvorisek:
@lubber-de please test and let me know if there is any usecase it breaks. This PR itself does not intend to fully fix #3199.

I am quite sure about the 2nd commit. Together with the 1st commit it fixes https://github.com/fomantic/Fomantic-UI/pull/3205#pullrequestreview-2679705613 well.

I am not fully sure about the 3rd commit, as "get values" API does encode only multiple values which is inconsistent design. But it does fix most issues of my local repros. So if such change does not break anything, I propose to land it and fix the API design later in a separate commit.

mvorisek:
@lubber-de might I have your review?

lubber-de (changes requested):
Unfortunately this still breaks in 2 situations when this PR is applied (by latest commit 9c732da6a48d5515e4ad24e3249e64eee0e949b3)

## 1.  Original Issue: #1373, Original PR: #1770

### FUI 2.9.4
https://jsfiddle.net/lubber/23htdnw0/1/
--> works

### FUI 2.10.0 beta (current develop 45ac3fe85a2123a2e060ceba7ce6f210eef3bb54)
https://jsfiddle.net/lubber/tdjmzscr/1/
--> works

### Your PR applied
https://jsfiddle.net/lubber/gmh5reoy/8/
--> fails (I cannot remove the 3rd entry (item 3" inches) inside the 2nd dropdown (made of select) anymore once selected ... seems the internal comparison between hidden input and data attribute does not match anymore)


## 2. Original Issue: #2782, Original PR: #2790

Mentioned example c) 

### FUI 2.9.4
https://jsfiddle.net/lubber/q8hL21yg/1/
--> works

### FUI 2.10.0 beta (current develop 45ac3fe85a2123a2e060ceba7ce6f210eef3bb54)
https://jsfiddle.net/lubber/f86g9L3v/2/
--> works

### Your PR applied
https://jsfiddle.net/lubber/zap03L4k/5/
--> fails (When preserveHTML is false, the ampersand is double encoded)

I havent checked the other modules changes yet (they are the same), but i feel, always encoding the ampersand, instead of checking for already encoded htmlentities (the regex you changed from `["'<>]|&(?![\d#A-Za-z]{1,12};)` to `["&'<>]` ) should not be changed. At least this will prevent double encoding

mvorisek:
> 1. Original Issue: #1373, Original PR: #1770

This is (will be) fixed in #3199 already.

> 2. Original Issue: #2782, Original PR: #2790

image

This issue is given by browser behaviour - https://jsfiddle.net/0m21broh/. Without no code, browser parses/normalizes `&` as if it was `&amp;`. This is legal, as both `&` and `&amp;` render as `&`.

I propose to fix this issue later once this PR and #3199 is landed. The solution would be to to revert `&amp;` to `&` where obtained as early as possible.

> I havent checked the other modules changes yet (they are the same), but i feel, always encoding the ampersand, instead of checking for already encoded htmlentities (the regex you changed from ["'<>]|&(?![\d#A-Za-z]{1,12};) to ["&'<>] ) should not be changed. At least this will prevent double encoding

It must be changed, otherwise values containing `&amp;` are saved as `&` which is a bug. Dropdown must never mutate values and all encoding functions must be strictly bijective.

With this said, I propose to merge this PR as is, then let me finish #3199 and then let's look into the `&` with `preserveHTML: false`. This is present only when the values are initialized thru interpretable HTML (`<option>xxx</option>`), not when using attributes or JS.

lubber-de:
Don't get me wrong: I don't want to merge PRs which **knowingly** break previously fixed/working functionality (and isn't a intended breaking design change like #3200). Even if i trust you, to fix this in a later PR.

Everything that worked before should work afterwards. The develop branch is supposed to be the always working/approved one.
So, if the issues are fixed in another PR already, merge them into this one and i am fine with that 🙂 And your original #3199 would also be fixed then (?)

> Without no code, browser parses/normalizes & as if it was &amp;. This is legal, as both & and &amp; render as &.

Well, then we could also argue why the & is encoded at all if surrounded by space  when a single & is still valid. (2.9.4 and 2.10 beta dont do this, your change does). Now people still have single & in their data which would break now.

> It must be changed, otherwise values containing &amp; are saved as & which is a bug. Dropdown must never mutate values and all encoding functions must be strictly bijective.

Ok, agreed, still needs testing (espcially for the other modules and their usages)

mvorisek:
> Don't get me wrong: I don't want to merge PRs which **knowingly** break previously fixed/working functionality (and isn't a intended breaking design change like #3200). Even if i trust you, to fix this in a later PR.
> 
> Everything that worked before should work afterwards. The develop branch is supposed to be the always working/approved one. So, if the issues are fixed in another PR already, merge them into this one and i am fine with that 🙂 And your original #3199 would also be fixed then (?)

100% agree and yes, #3199 would be fixed with this and #3205.

> > Without no code, browser parses/normalizes & as if it was &. This is legal, as both & and & render as &.

It turned out the fix is easy. I checked all `.html()` calls (and there are not innerHtml/outerHtml).

All expect 1 are irrelevant as they are either for preserveHTML=true or https://github.com/fomantic/Fomantic-UI/blob/2.9.4/src/definitions/modules/dropdown.js#L3129.

So only one `.html()` result was needed to be fixed/"assume ampersand unescaped".

I also checked all other files/modules affected by https://github.com/fomantic/Fomantic-UI/pull/3224/commits/987a21b5f6d0442b09cf9c177a76d7fe8c4ec51b and all `.html()` usages in them do not have to be fixed.

Works well on all repros I tested.

To integrate this, please:

1. review this PR for correctness
2. review #3205 for correctness and test all edge usecases
3. if everything is fine, merge this PR and then immediatelly #3205 (there are no git conflicts expected)

mvorisek:
@lubber-de might you look into this?

After this PR and https://github.com/fomantic/Fomantic-UI/pull/3205, the develop branch should be fixed /wo any BC break.

lubber-de (changes requested):
Unfortunately this still breaks in 1 situation when this PR is applied (by latest commit a995f2fce5067ca029f07031ec5ddfe179496088)

## 1.  Original Issue: #1373, Original PR: #1770

### FUI 2.9.4
https://jsfiddle.net/lubber/23htdnw0/1/
--> works

### FUI 2.10.0 beta (current develop 45ac3fe85a2123a2e060ceba7ce6f210eef3bb54)
https://jsfiddle.net/lubber/tdjmzscr/1/
--> works

### Your PR applied
https://jsfiddle.net/lubber/gmh5reoy/10/
--> FAIL: 1st dropdown: When (item 3" inches) is selected and removed again, it is not listed in the dropdown anymore
--> FAIL: I (still) cannot remove the entry (item 3" inches) inside the 2nd (select) dropdown anymore
--> fails, regardless, if preserveHTML is true or false

However, i tried the same using your latest commit of #3205 : And there it works(!) https://jsfiddle.net/lubber/e5ru0239/1/

So, why don't we just merge the other PR where everything seems to be fixed (i would need to check all the other use-cases there as well) instead of this "in-between" PR here? (which still breaks the above mentioned situation)

mvorisek:
> However, i tried the same using your latest commit of #3205 : And there it works(!) https://jsfiddle.net/lubber/e5ru0239/1/
> 
> So, why don't we just merge the other PR where everything seems to be fixed (i would need to check all the other use-cases there as well) instead of this "in-between" PR here? (which still breaks the above mentioned situation)

That is great and expected.

This PR is solely about "ampersand and double encoding fixing".

I prefer to merge this first and then the 2nd PR separately, as both PRs fix something different, so this way squashed history will be much more readable. Develop branch will be "stable" afterwards.

lubber-de:
Please see my comment for the "always encode ampersand for non dropdown functions which dont expect a reverse function" at https://github.com/fomantic/Fomantic-UI/pull/3205#discussion_r2014713814 (i should have commented that in this PR as you want this one to be merged first

lubber-de (approved):
LGTM. Thanks for your patience!

PR #3225, opened 2025-03-19
feat(chore): remove check for mutationoberserver support

lubber-de:
## Description

Followup of #3183 
This PR removes the check for MutationOberserver support as [all browser since IE11 are supporting it](https://caniuse.com/?search=mutationobserver)


ko2in (approved):
LGTM.

PR #3231, opened 2025-03-26
Fix lossy HTML escape in all modules

mvorisek:
as discussed in https://github.com/fomantic/Fomantic-UI/pull/3205#discussion_r2014886407

The first 2 commits fix #3224.

The remaining commits are relanding postponed https://github.com/fomantic/Fomantic-UI/pull/3224/commits/8590c2101875d0158960ccb9135b87469dc40008. I checked all `escape(` usages in each module. Text like `&amp;` should be always encoded into `&amp;amp;` as the meaning is to encode the input and the input is never from innerHTML.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1955,7 +1955,7 @@
                             .find('option')
                             .each(function () {
                                 let $option = $(this);
-                                let name = module.escape.assumeUnescapedAmpersand($option.html());
+                                let name = module.escape.assumeUnescapedAmpLtGt($option.html());
                                 let disabled = $option.attr('disabled');
                                 let value = $option.attr('value') !== undefined
                                     ? $option.attr('value')
@@ -3557,13 +3557,13 @@
                     },
 
                     // https://github.com/fomantic/Fomantic-UI/issues/2782
-                    // https://jsfiddle.net/wdyjfvz0/
-                    assumeUnescapedAmpersand: function (string) {
+                    // https://jsfiddle.net/3efL7jnt/
+                    assumeUnescapedAmpLtGt: function (string) {
                         if (settings.preserveHTML) {
                             return string;
                         }
 
-                        return string.replaceAll('&amp;', '&');
+                        return string.replaceAll('&amp;', '&').replaceAll('&lt;', '<').replaceAll('&gt;', '>');
                     },
                 },
 
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3562,7 +3562,13 @@
                             return string;
                         }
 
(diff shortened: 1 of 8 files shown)


src/definitions/modules/embed.js
```diff
@@ -612,7 +612,7 @@
 
         templates: {
             escape: function (string) {
-                const badChars = /["'<>]|&(?![\d#A-Za-z]{1,12};)/g;
+                const badChars = /["&'<>]/g;
```
lubber-de: What about making this regEx a setting parameter like "regEx.escape" like we have in search, form and dropdown? (default /["&'<>]/g)
This way, the previous approach (allowing to not encode existing htmlentities), would be at least possible to adjust by the user via

```javascript
.dropdown({
regExp: {
    escape: /["'<>]|&(?![\d#A-Za-z]{1,12};)/g
}
})
```

mvorisek: I am strongly againts. The reason is such function/regex should never be changed by the developer. And as discussed in https://github.com/fomantic/Fomantic-UI/pull/3205#discussion_r2014886407 and each commit message, the original regexes are only helpful for input data from innerHTML. This is the case only in 1 single usage in dropdown.

If user want to call FUI API with data from HTML, he must deescape the input like https://github.com/fomantic/Fomantic-UI/pull/3231/commits/17dd1fd017a521994a570b213eed0302e3000b40.

src/definitions/modules/dropdown.js
```diff
@@ -3557,13 +3557,13 @@
...
-                    // https://jsfiddle.net/wdyjfvz0/
-                    assumeUnescapedAmpersand: function (string) {
+                    // https://jsfiddle.net/3efL7jnt/
+                    assumeUnescapedAmpLtGt: function (string) {
                         if (settings.preserveHTML) {
                             return string;
                         }
 
-                        return string.replace('&amp;', '&');
+                        return string.replaceAll('&amp;', '&').replaceAll('&lt;', '<').replaceAll('&gt;', '>');
```
lubber-de: Using replaceAll will force us to limit the next Version to iOS13.1+
I tend to do that, as i also think about other newer syntax elements which will make the codebase simpler (we officially support the last 4 iOS Version, so that would be fine anyway) 

Otherwise the current codebase will work on iOS 10+

Just saying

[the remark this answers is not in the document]
mvorisek: fixed in https://github.com/fomantic/Fomantic-UI/pull/3231/commits/d57df08c4bb624a6a39abadbc85c8c2244edb4e3

mvorisek: fixed by coincidence, see https://github.com/fomantic/Fomantic-UI/pull/3231#discussion_r2015020379

lubber-de (approved):
LGTM

PR #3233, opened 2025-03-27
fix dropdown "get remoteValues" - use unencoded value

mvorisek:
extracted from #3219


mvorisek:
I have verified the corectness by testing what is passed into `module.read.remoteData`/`module.save.remoteData` with preserveHTML = false/true.

mvorisek:
@lubber-de can you merge this and the other 2 PRs?

lubber-de (approved):
LGTM

PR #3234, opened 2025-03-27
fix dropdown cmp in "get item" - use unencoded value

mvorisek:
extracted from #3219

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2036,8 +2036,8 @@
                         let isMultiple;
                         value = value !== undefined
                             ? value
-                            : (module.get.values() !== undefined
-                                ? module.get.values()
+                            : (module.get.values(true) !== undefined
+                                ? module.get.values(true)
                                 : module.get.text());
                         isMultiple = module.is.multiple() && Array.isArray(value);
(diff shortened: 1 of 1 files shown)


mvorisek:
I am 95% sure this fix is right, please review carefully.

lubber-de:
I am a bit confused about the intention of this (and other ) PR  :
in https://github.com/fomantic/Fomantic-UI/pull/3219#issuecomment-2755732872 you want to remove the whole "raw" setting, which i so far (once) tested and it seems to be the right way .

With the other PRs you are now making use of the raw parameter.... is the final intention to keep the raw logic , intially invented in #1770, or not?

mvorisek:
Here we change it from non-raw to raw. #3219 is refactoring PR only.

> is the final intention to keep the raw logic , intially invented in https://github.com/fomantic/Fomantic-UI/pull/1770, or not?

Yes, raw only, non-raw does not make sense anywhere.

mvorisek:
Does this makes sense?

I maybe splitted the PR a little too much, but all 3 cases were quite different with very different repro steps, so I was a little to scared to solve them in a single PR...

lubber-de (approved):
LGTM

PR #3235, opened 2025-03-27
fix dropdown cmp in "set value" - use unencoded value

mvorisek:
extracted from #3219

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2540,7 +2540,7 @@
                             $input.addClass(className.noselection);
                         }
                         let hasInput = $input.length > 0;
-                        let currentValue = module.get.values();
+                        let currentValue = module.get.values(true);
                         let stringValue = value !== undefined
                             ? String(value)
                             : value;


src/definitions/modules/dropdown.js
```diff
@@ -2540,7 +2540,7 @@
                             $input.addClass(className.noselection);
                         }
                         let hasInput = $input.length > 0;
-                        let currentValue = module.get.values();
+                        let currentValue = module.get.values(true);
```
mvorisek, line 2543: I am quite sure here, but for dropdown multiple, this seems wrong as the `currentValue` of type array is compared to string later.

But this PR does not intend to fix that, it fixes comparing unencoded value to unencoded value.

lubber-de: Yes, this strange codeblock logic is still original SUI. However, it "works" as the comparison will return false when it's a multiple dropdown thus this block is always skipped.

mvorisek: I was unable to find any usecase when this implied issues.

However, we need to compare apples to apples, so the this PR fixes the 1st step to compare raw to raw values.

lubber-de (approved):
LGTM

Issue #3232, opened 2025-03-26
[Dropdown] Allow to edit "multiple item"


mvorisek:
# Feature Request

repro: https://jsfiddle.net/evor619s/

Dropdown multiple should allow to fix typos/change in/of already added item.

Steps to reproduce:

1. open the repro
2. click the "foo" item
3. I would expect the item to be changed to regular input (like when searching/adding a new item) and be able to change it to "f2oo" with "mouse click after letter "f" and typing 2 on the keyboard"

Depending on the `allowAdditions` config the item can or can't be edited to unlisted value.

This feature request is especially useful for longer items like email addresses. (currently the existing item has to be deleted and retyped)

BrianJavier99:
Hello @mvorisek, I have attempted to implement this feature request in this [pull request](https://github.com/fomantic/Fomantic-UI/pull/3247).

My implementation is a little bit different: it allows you to edit existing selected items, instead of using the search input to edit the values, as I have found it to be possibly confusing in a few situations.

mvorisek:
Might I know what exactly you found confusing?

I belive the editing should be done using the regular editing input/API to maintain the same look&feel. User should be definitely allowed to change some value to a value listed from all existing values (or remote API).

lubber-de:
What about using `contenteditable="plaintext-only"` instead of switching to an input field? This way the whole label stays visible as it is?

BrianJavier99:
> Might I know what exactly you found confusing?
> 
> I belive the editing should be done using the regular editing input/API to maintain the same look&feel. User should be definitely allowed to change some value to a value listed from all existing values (or remote API).

- When trying to edit an existing item while there's an existing value in the search input, do we discard the current value in the search input and replace it with the value of what we're trying to edit?
   ![Image](https://github.com/user-attachments/assets/06426b49-c458-41e9-9621-b13dea889a5f)

- What happens if we try to edit another input while we're in the middle editing one? This can lead to multiple selected items being accidentally deleted. 
   ![Image](https://github.com/user-attachments/assets/27310fa6-efde-41fe-ae73-0c78d8e92384)
   ![Image](https://github.com/user-attachments/assets/aa4ed097-cc54-48ef-af80-8ecebd0d8b21)

BrianJavier99:
> What about using `contenteditable="plaintext-only"` instead of switching to an input field? This way the whole label stays visible as it is?

[...]

PR #3247, opened 2025-04-17
feat(dropdown): allow editing of added items

BrianJavier99:
## Description

I have added a functionality that allows direct editing of the labels by clicking them when `allowAdditions` is set to **true**.
The expected behavior is as follows:
- All behaviors detailed below will only trigger if `allowAdditions` is set to **true**.
- Clicking anywhere inside the item (except for the close icon) will trigger the edit functionality.
- When edit functionality is triggered, the close icon will disappear, and the keypress events for the `Del` and `Backspace` keys are also ignored to prevent accidental deletion of the item.
- To save changes, one has to click outside of the element to trigger the blur, or press `Enter` key.
- If editing a selected item that is not added by the user, what goes behind the scenes is that **the non-addition item will get unselected, and a new item will be added to the selection**, which is now considered a user addition. This means that the non-addition item will appear back to the dropdown menu.
- When editing an existing item while having a value at the search input at the same time, pressing `Enter` key would only save the changes made to the existing item.
- When editing an existing user-added item, if the new value matches one of the unselected menu items, that menu item will be have the class "filtered" added to it, making it disappear in the selection.
- The changes made when editing item labels reflect in the dropdown's hidden input value.
## Testcase

https://jsfiddle.net/ReiVanBjair/a37v91gb/15/

## Screenshot (if possible)

image
image
image
image
image

## Closes

#3232


ko2in:
@BrianJavier99 Looks great :+1:

I think it's more effective to bind the escape key for canceling edits. And, the text box design could use a minor tweak; perhaps we could remove the background and padding from the container. This way, leaving the text box by itself would improve both spacing and the editing experience. :thinking: 

FUI-Editing-Dropdown

PR #3257, opened 2025-04-30
fix(table): inverted basic table background

### LINKED ISSUE #3251: [Table] `very basic` + `inverted` table style seems wrong
# Bug Report
When using the `very basic` + `inverted` class style for tables, the padding is removed on the sides which makes the table look off with it's background-color.

## Steps to reproduce
```html
<table class="ui very basic inverted table">[...]
```
## Expected result
The background-color is transparent for the `inverted` table (just like it is for non-inverted).

## Actual result
The background-color is dark/non-transparent (`rgba(0, 0, 0, 0.15)`)

## Testcase
https://jsfiddle.net/pgkb8zd0/1/

## Screenshot (if possible)
| Inverted Table                                                                                     | Regular Table                                                                                     |
| -------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
| Inverted Table | Regular Table |

## Version
2.9.4

lubber-de:
## Description
`inverted basic table` table cells needed more specificity to override the background color in its header/footer elements

## Testcase
Remove CSS to see issue
https://jsfiddle.net/yjgb6e3v/

## Closes
#3251

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -1667,19 +1667,19 @@
     .ui.basic.table > tfoot {
         box-shadow: none;
     }
-    .ui.basic.table > thead > tr > th,
-    .ui.basic.table > tbody > tr > th,
-    .ui.basic.table > tfoot > tr > th,
-    .ui.basic.table > tr > th {
+    .ui.ui.basic.table > thead > tr > th,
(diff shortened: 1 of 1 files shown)


ko2in (approved):
LGTM.

PR #3258, opened 2025-05-02
fix(dropdown): keep object values as string

### LINKED ISSUE #3005: [Dropdown] Using serialized JSON as value prevent the select value from being updated.
# Bug Report

## Steps to reproduce

1. Selection Option A
2. Submit 

## Expected result
Select value is should be updated according to the option value.

## Actual result
Select value is not updated according to the option value.

## Testcase
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <script src="https://cdn.jsdelivr.net/npm/jquery@3.7.1/dist/jquery.min.js"></script>
    <script src="https://cdn.jsdelivr.net/npm/fomantic-ui@2.9.3/dist/semantic.min.js"></script>
    <link href="https://cdn.jsdelivr.net/npm/fomantic-ui@2.9.3/dist/semantic.min.css" rel="stylesheet">
    <title>Demo</title>
</head>
<body>
    <form method="get" accept-charset="utf-8">
        <select name="select" id="select">
            <option value='{"foo": "bar"}'>Option A</option>
            <option value='{"foo2": "bar2"}'>Option B</option>
            <option value='1'>Option C</option>
        </select>
        <button>Submit</button>
    </form>
    <script>$("#select").dropdown({
        debug: true,
        verbose: true
    })</script>
    </body>
</html>
```

## Version

2.9.3 on Firefox 122.0.1

lubber-de:
## Description
If a JSON value was used as a dropdown item value, it was internally broken as the value was casted to String resulting into `[object Object]`
Now, having multiple JSON values for more than 1 dropdown item the value couldn't be compared / set correctly as each values had the same `[object Object]`.
This PR fixes it by replacing the simple castings to JSON stringify which keeps the value structure so the dropdown items keep their unique value

## Testcase
- Open console
- Select "Option A"
- check the debug log for "Updating input value"

### Broken
https://jsfiddle.net/ReiVanBjair/ry4123z0/11/

### Fixed
https://jsfiddle.net/lubber/L0ysfoqj/

### Broken
image

### Fixed
image

## Closes
#3005

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1852,6 +1852,11 @@
                             return range.text.length - rangeLength;
                         }
                     },
+                    stringValue: function (value) {
+                        return typeof value === 'object'
+                            ? JSON.stringify(value)
+                            : String(value);
+                    },
                     value: function () {
                         let value = $input.length > 0
                             ? $input.val()
@@ -1923,7 +1928,7 @@
                         let choiceValue = $choice.data(metadata.value);
 
                         return choiceValue !== undefined
-                            ? JSON.stringify(choiceValue)
+                            ? module.get.stringValue(choiceValue)
                             : (typeof choiceText === 'string'
                                 ? String(
                                     settings.ignoreSearchCase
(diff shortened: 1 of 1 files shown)


ko2in (commented):
It doesn't work well with encoded JSON string. https://jsfiddle.net/ko2in/ftpLscv6/

For security, the JSON object usually inserts as encoded format especially from server side like PHP htmlentities function.

ko2in (approved):
LGTM.

mvorisek:
Kind of https://github.com/atk4/ui/pull/2274 and related to #3223.

The issue with this PR is "string formatted like valid JSON" will not be formatted as before - ex. input `[0 ]`.

PR #3269, opened 2025-05-12
change(modal): allows image description flexbox to take available space

### LINKED ISSUE #3268: [modal] form in modal with image doesn’t utilise full width
# Bug Report
If a form is included in a modal with an image to the side, the form fields do not use all the available space

By changing flex to `flex: 1 0 auto` on description class, it seems to then use the full width, but not sure if better solution exists or if this breaks something else

## Testcase
https://jsfiddle.net/f9uwm0ov/

## Version
2.9.4

ko2in:
## Description
The flex item for the image description in the modal does not utilize the available space, which restricts its child elements from being displayed at full width when required.

It results in considerable adverse layout impacts when an element, such as a form is displayed inside the description container. The form components, including input fields, fail to utilize the available space, causing the form to appear constricted within the modal.

This PR modifies this behavior of the flexbox item for the description, enabling it to be flexible and allowing it to grow and shrink.

So, the description container will grow in the absence of sibling elements following it, allowing the child elements within to take available space and display in a full-width format, particularly in the context of the form layout.

## Testcase
**Before:** https://jsfiddle.net/f9uwm0ov/

**After:** https://jsfiddle.net/ko2in/d7o8fjak/

### Before
Before

### After
After

## Closes
#3268


lubber-de (approved):
LGTM

PR #3273, opened 2025-05-13
fix(modal): Change default min-width value of description flexbox to 'auto' 

ko2in:
## Description
The default min-width value for the image description flexbox container within the modal is unset, resulting in a warning for an invalid value appearing in the browser's developer console.

The PR modifies the default value to 'auto'.

## Testcase
Go to the following fiddles. Open the browser developer's tool, and inspect the description element inside the modal, and looks at it's CSS properties.

**Before:** https://jsfiddle.net/ko2in/d7o8fjak/

**After:** https://jsfiddle.net/ko2in/zkb43x51/

### Before
invalid-min-width-value

### After
valid-min-width-value


lubber-de (approved):
LGTM

PR #3286, opened 2025-06-10
fix(label): resolve the overlap issue during the image transition

### LINKED ISSUE #3279: [Visibility, Label] Ribbon label showing behind the image on transition
# Bug Report
When using visibility transition to lazy load card image, ribbon label showing behind the image during transition

## Steps to reproduce
1. Create a card with image
2. Add ribbon label to the image
3. Add visibility for lazy loading the image

## Expected result
Ribbon label should be appeared on top on the image during transition

## Actual result
Ribbon label showing behind the image

## Testcase
https://jsfiddle.net/spjeqh47/3/

## Version
2.9.4

ko2in:
## Description
When using the ribbon label for the image card that features a visibility transition effect for lazy loading, it was observed that the ribbon label was partially overlapped by the image during the transition. This occurred because the ribbon is positioned absolutely without z-index value, which results the sibling image element to overlap.

This PR sets z-index value to the ribbon label within the image card which resolve the overlap issue.

## Testcase
**Before**: https://jsfiddle.net/spjeqh47/3/
**After**: https://jsfiddle.net/ko2in/9sma0c3z/

## Screenshot
| Before | After |
| ----------- | ----------- |
| fui-label-bug | fui-label-fixed |

## Closes
#3279

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -471,6 +471,7 @@ a.ui.label {
     .ui.card .image > .ribbon.label {
         position: absolute;
         top: @ribbonImageTopDistance;
+        z-index: 1;
     }
     .ui.card .image > .ui.ribbon.label,
     .ui.image > .ui.ribbon.label {


lubber-de (approved):
LGTM

PR #3294, opened 2025-06-19
fix(reveal): Add background and width to content

### LINKED ISSUE #3292: [Reveal, Card] Visible content and hidden content displayed together
# Bug Report
When using `reveal` elements to show or hide content on hover, the visibile content and hidden content displayed together inside `card` content.

## Expected result
The hidden content should be hidden by default

## Actual result
The visibile content and hidden content displayed together. Its because visible content doesn't have initial background.

## Testcase
https://jsfiddle.net/fm4pb18L/

## Screenshot (if possible)
Image

## Version
2.9.4

ghabriel25:
## Description
This PR fix transparent background on visible content that makes hidden content displayed as well and add default width to the content so it will fully hidden on hover.

Add `background` to `.ui.reveal>.visible.content` and `width` to `.ui.move.reveal>.content`

## Testcase
https://jsfiddle.net/w80avgu9/9/

## Closes
#3292

### DIFF
--- a/src/definitions/elements/reveal.less
+++ b/src/definitions/elements/reveal.less
@@ -155,6 +155,7 @@
 
     .ui.move.reveal > .content {
         display: block;
+        width: 100%;
         float: left;
         white-space: normal;
         margin: 0;
--- a/src/definitions/elements/reveal.less
+++ b/src/definitions/elements/reveal.less
@@ -28,6 +28,7 @@
 }
 
 .ui.reveal > .visible.content {
+    background: #fff;
     position: absolute !important;
     top: 0 !important;
     left: 0 !important;


lubber-de (changes requested):
Your fix is too generic, it only works for a reveal inside a non inverted card.

## background: #fff;
- When used inside an `inverted card`, the fixed white background is misplaced.
- As reveal can be used anywhere, you cannot tell if its running inside a white container or some other colored card / toast / table / body / etc..

## width: 100%
- breaks on images, hover reveal looks messed up having a big wide white block moving over the image

See https://jsfiddle.net/lubber/r73ubz6v/

ghabriel25:
@lubber-de Thanks for your review!

 I realized that this fix was too generic after looking at your points. I'm going to close this PR as not sure how to proceed with all variation especially because it can be used to all elements. I'll review past PR to get more information

PR #3304, opened 2025-07-06
fix(modal, flyout, sidebar): remove style, if initial bodymargin was 0

### LINKED ISSUE #3303: bodyMargin after closing the modal window
# Bug Report

## Steps to reproduce
1. The initial state of the body tag on the page
Image

2.
A dialog box has been opened and css style has been added to the body tag.
Image

3.
After closing the modal window, css style remains in the body tag, they interfere with the work of other plugins.
Image

This problem prevents plugins that open modal windows by adding a class.

## Expected result

 style="margin-right: 0px

## Actual result

 style="margin-right: 0px  or delete the style attribute.

## Version
Fomantic-UI 2.9.3

lubber-de:
## Description

When a modal/flyout/sidebar is opened, it adjusts the body (or context) margin to avoid moving the modal by the width of a possible scrollbar.
When the modal closes, it reverts the value back to the initial one.
However, the default of margin/padding if 0, so there is no need to keep the inline style to 0. Instead the style attribute can be simply removed. This solves some issues in cases people use custom classes which adjusts the body margin as well

## Closes
#3303

### DIFF
--- a/src/definitions/modules/flyout.js
+++ b/src/definitions/modules/flyout.js
@@ -81,7 +81,8 @@
             let ignoreRepeatedEvents = false;
             let isBody = $context[0] === $body[0];
             let initialBodyMargin = '';
-            let tempBodyMargin = '';
+            let initialBodyMarginInt = 0;
+            let tempBodyMargin = 0;
             let hadScrollbar = false;
             let windowRefocused = false;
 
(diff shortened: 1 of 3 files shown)


ko2in (approved):
LGTM.

PR #3326, opened 2025-08-17
feat(lint): modern color function notation

lubber-de:
## Description
Switching to modern RGB CSS color function notation (default in stylelint-config-standard)

More optimizations like switching to percentages will be covered by additional PRs


ko2in (approved):
LGTM.

PR #3360, opened 2025-09-18
fix(chore): replace deprecated event triggering

lubber-de:
## Description

Replace deprecated event creation by modern Event() constructor. Only IE11 was not supporting it, so it's time to replace this 😉


ko2in (dismissed):
LGTM. Not sure why do we use plural here since it's a single event.

lubber-de:
> LGTM. Not sure why do we use plural here since it's a single event.

I changed it to singular. Perhaps it was plural because of the previously used "HTMlEvents" method

PR #3363, opened 2025-09-23
fix(api): remove legacy code support and checks

lubber-de:
## Description

Remove support and checks for SUI 1.0 legacy settings (not even documented since years anymore)


ko2in (approved):
LGTM.

PR #3400, opened 2025-11-11
feat(chore): adjust minimum browser support to 2021+ features

lubber-de:
## Description

Because of the recent (#3396) and upcoming changes, we need to clarify the browser support:

- Chrome >= 88
- Edge >= 88
- Safari >= 14.1
- Firefox >= 82
- Opera >= 75
- Samsung Internet >= 15
- iOS >= 14.1
- Android >= 8

See https://browsersl.ist/#q=chrome+%3E%3D+88%2C+edge+%3E%3D+88%2C+safari+%3E%3D+14.1%2C+firefox+%3E%3D+82%2C+ios+%3E%3D+14.1%2C+android+%3E%3D+8%2C+opera+%3E+75%2C+samsung+%3E%3D+15%2C+%3E0.3%25

According to the above seen statistics we could even raise this even more, but lets see how far the next version really goes :wink:

THis PR adjusts the readme and build tasks when using browsersList accordingly

## Related
#2854

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,10 +55,12 @@ Fomantic includes an interactive installer to help setup your project.
 
 ### 💻 Browser Support
 
-* Last 2 Versions of Firefox, Chrome, Safari Mac, Edge
-* Last 4 Versions of Android, Chrome for Android, iOS Safari[^1]
-
-[^1]: Fomantic-UI should basically still work in iOS Safari 11+, Android 7+, but, starting from v2.9.0, we won't support them anymore if anything works different than in recent versions.
+* [![Chrome](https://cdnjs.cloudflare.com/ajax/libs/browser-logos/75.0.1/chrome/chrome_16x16.png)]() Chrome >= 88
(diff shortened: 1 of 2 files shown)


ko2in (dismissed):
LGTM.

ko2in:
> * Edge >= 88

Do we really need to have for Edge since it's using the same engine as Chrome?

lubber-de:
Edge also exists as ie engine variant until v18, so to clarify this (and also the browser list rule to not create old vendor specific properties) I would keep it. As edge is the default for win 10/11 I would keep this info. Yes, other browsers like brave/opera/... nowadays also use the chromium engine, but market share is much lower.

Also caniuse.com aswell as MDN specifies them in that order
I would add opera and samsung instead to be consistent with those ?

ko2in:
> I would add opera and samsung instead to be consistent with those ?

Yes. Maybe, this would be a good idea to add those since those can be used by the end-users. It's OK to forget about Brave as only people in tech field prefer to use it in my opinion.

PR #3450, opened 2026-02-20
build(deps-dev): move actions and octokit to devDependencies

hkrutzer:
## Description
This moves octokit and Github actions dependencies to devDependencies,
so that consumers of the package won't have to install them when installing the package.

## Testcase
N/a

## Screenshot (if possible)
N/a

## Closes
-


lubber-de (approved):
Thanks. Makes sense. I just had to make eslint happy, so i slightly adjusted your PR.
LGTM!

hkrutzer:
Thank you!

PR #3528, opened 2026-07-05
fix(flyout): support dynamic heights on mobile devices

lubber-de:
## Description

Modern followup of #2644 
This is now completely solvable via pure CSS, by using the "dvh" unit supporting dynamic view heights on mobile devices (which removes the address bar on scrolling which in turn changes the height)
The now unnecessary JS code, which would otherwise calculate wrong values for the min height, can be completely removed.
Same applies for fullscreen modal, so this PR adjusts this as well.

This a **breaking change** as it needs a modern browser (12/2022+, iOS15.4+) supporting the dvh unit
https://caniuse.com/?search=dynamic+viewport+units

## Testcase
 Try the fomantic docs site for flyout and run the first example on a mobile device. Then scroll up/down and the actions area gets too high.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved flyout and modal scrolling height behavior in modern browsers using dynamic viewport sizing for more reliable layout.
  - Prevented flyouts from performing resize-time height recalculations in lateral layouts to reduce layout jumps.

- **Documentation**
  - Refreshed the “Browser Support” section with higher minimum supported versions and updated how browser icons are displayed.

- **Compatibility**
  - Updated the default browser targets to align with the newly raised minimum versions.


lubber-de:
@coderabbitai review

PR #3542, opened 2026-07-18
fix(lint): enable @typescript-eslint/triple-slash-reference rule

lubber-de:
## Description

- Enabled the typescript eslint rule `@typescript-eslint/triple-slash-reference`
- rule "ban-types" does not exist in eslint 8 so i simply removed it inside this PR as well

@prudho @KiddoV You seem to be an active user of the types, so i would appreciate if you could test if the change to `import` will break anything (at least eslint is satisfied...)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -106,9 +106,7 @@ module.exports = {
             'no-use-before-define': 'off',
 
             // TODO rules with a lot of errors to be fixed manually, fix in a separate PR
-            '@typescript-eslint/ban-types': 'off', // 16 eslint errors only, help wanted!
             '@typescript-eslint/no-explicit-any': 'off',
-            '@typescript-eslint/triple-slash-reference': 'off',
         },
     }],
 };
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2,32 +2,31 @@
 // Project: https://github.com/fomantic/Fomantic-UI
 // Definitions by: Fomantic Team <https://github.com/fomantic>
 
-/// <reference types="jquery" />
+import type {} from 'jquery';
 
-/// <reference path="fomantic-ui-accordion.d.ts" />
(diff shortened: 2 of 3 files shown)


KiddoV:
Hi @lubber-de,

Sorry, it's been a while since I last checked on this library.

After updating my `package.json` from `"fomantic-ui": "^2.9.4"` to:

```json
{
  "fomantic-ui": "github:lubber-de/Fomantic-UI#triple-slash-reference"
}
```

all of my ESLint checks started reporting the following error:

```text
Property '___' does not exist on type 'JQuery<HTMLElement>'.
```

The application still runs, but I've run into a few issues:

1. ESLint reports the error above for all Fomantic jQuery methods.
2. All of my FUI icons are missing and are rendered as empty squares instead. (Not sure if this related?)

Am I missing an additional setup step after switching to this branch?

lubber-de:
> all of my ESLint checks started reporting the following error:
> 
> ```
> Property '___' does not exist on type 'JQuery<HTMLElement>'.
> ```

Very strange, if this (IMHO unrelated) change causes it for your code. 
> 
> The application still runs, but I've run into a few issues:
> 
> 1. ESLint reports the error above for all Fomantic jQuery methods.

Again, thanks for testing, if eslint breaks for you because of this change, i'll put it this PR on hold.
I just wanted to make the eslinter happy at build time. Maybe prudho ran into a similar issue back then, when he invented the types, and therefore used the triple slash method.

> 2. All of my FUI icons are missing and are rendered as empty squares instead. (Not sure if this related?)
 > Am I missing an additional setup step after switching to this branch?

Not seeing the icons cannot be caused by this PR.
The develop branch aka nightly aka 2.10 beta, still uses the same FA icons as 2.9.4 does.
Only difference: We dropped IE, so the woff (not woff2) font files were removed. This probably causes it for you, if you still rely on them (?)

